PR #74, opened 2021-11-05
Publish to Sonotype Maven Central 

notmandatory:


notmandatory:
I published a new 0.1.2 version with the commits that add the test fixtures back in. Still have some issues getting android tests to work with jvm test fixtures.

notmandatory:
I had to duplicate the test code for the bdk-android and bdk-jvm projects for now, couldn't find a way to share the code via the jvm testFixtures. 

I also bumped the version to `0.1.3-dev` to differentiate current work from published `0.1.2` version and future `0.1.3` version.

PR #83, opened 2021-11-12
Refactor transaction 'id' property to 'txid'

thunderbiscuit:
This is a small refactor of the `TransactionDetails` object, renaming the `id` property to `txid`.

- [x] I have PGP signed my commit.

I ran the Android demo and the tests and everything works well. I couldn't test the iOS side however, but all mentions of `id` have been changed to `txid` in the iOS demo as well.


artfuldev:
We also need to update the ios example app.

artfuldev (approved):

artfuldev:
LGTM! @notmandatory you might wanna take a look.

notmandatory (approved):
ACK b0ef4652aef11a855b02653216c37df8226168d8

:rocket:

PR #89, opened 2021-11-14
[WIP] Add Python language bindings

thunderbiscuit:
This is a draft PR that brings in Python language bindings.

It isn't really functional at this point because UniFFI [does not support callbacks for Python bindings yet](https://github.com/mozilla/uniffi-rs/pull/1086), preventing the `OnlineWallet` from working. This is also the reason why I had to create a `bdk-python.udl` file specific for the Python bindings, removing the `BdkProgress` interface.

- [x] My commits are PGP signed
- [x] The `-p` flag is added to `build.sh`
- [x] The `-p` flag is added to `test.sh`
- [x] The test is passing

I will look at writing more of the tests we can currently do, and prepare some for when callbacks are ready and the full API is available. Currently very much WIP. Comments welcome.


build.sh
```diff
@@ -124,6 +125,25 @@ build_android() {
...
+      ;;
+    "Linux")
+      echo -n "linux" ;;
+  esac
+  echo "library to python sub-project"
+}
+
+build_python() {
+  copy_lib_python
+  uniffi-bindgen generate src/bdk-python.udl --no-format --out-dir bindings/bdk-python/test/ --language python
```
notmandatory, line 144: Shouldn't the output be to the `bindings/bdk-python/src` folder and then imported by your test?

Some info in the ["Packaging Python Projects"](https://packaging.python.org/tutorials/packaging-projects/) docs.

thunderbiscuit: Yes this is a good idea. I'm a bit rusty on my Python and wanted to commit something that was running, but I'll take a look at cleaning up the directory!

thunderbiscuit:
The current file layout is the recommended way to do things (`src/` and `tests/` in separate directories), but it is causing me some slight path issues.

I have been unsuccessful at having the native library live inside the `src/` directory with the `bdk.py` module. If you do that the tests will not recognize the symbols.

My current thinking is that the `src/` directory should have all the required files for the package (which would include the native lib), and the `tests/` directory should be able to simply import the library and run the tests. This is not the case at the moment. Anyone with more Python experience feel free to throw suggestions at me. For now not a dealbreaker, I'll just keep building on that.

Worthy of note is that this PR currently breaks the `bdk.udl` file used for the Kotlin and Swift bindings. So it's really not for merging, but more for testing. At the moment you can't build both Kotlin/Swift _and_ Python with the same `.udl` file, and adding a `bdk-python.udl` file also requires changes to the `build.rs` and the `lib.rs` files, so I don't think it's a good way to do it either. For now this branch can _only_ build the Python bindings.

notmandatory:
Will go in it's own repo. Closing this PR.

PR #100, opened 2022-02-08
Generate Python bindings

afilini:
This PR adds a `generate` binary that can be used to generate bindings without having to install the uniffi cli separately.

It's designed to be modular, even though right now only Python is implemented.

This can be run with:

```
GENERATE_PYTHON_BINDINGS_OUT=./outdirectory cargo run --bin generate --features generate-python
``` 

And it will place the `bdk.py` file in `outdirectory`.

The `GENERATE_PYTHON_BINDINGS_FIXUP_LIB_PATH` variable can be used to patch `bdk.py` and replace the library path, which is very useful when packaging the shared lib into a Python wheel


thunderbiscuit:
Tested ACK. I tested the resulting `bdkpython` package for `3.9` on macOS M1 and it worked well.

For other reviewers, here are some notes on the new workflow. After this PR is merged and the [setuptools-rust](https://github.com/afilini/bdk-python/tree/setuptools-rust) branch has been merged in `bdk-python`, the workflow is different than with the Swift and JVM bindings and goes like so:
1. Run the `generate.sh` script, which produces the bindings file, fixes the `loadIndirect()` function, and brings the file in the python `src` directory
2. Build the package using the typical `python -m build`, where it relies on `setuptools-rust` to build the native lib file and add it to the package
3. Use GitHub Actions to build for Linux, macOS, and Windows operating systems for Python versions 3.7 and up.

thunderbiscuit:
I'm not sure if we should generalize the idea of the binary that allows for not using the cli or if we should create other ones for the other bindings, but it's definitely something that needs addressing (the requirement on the cli and the version match and all).

artfuldev:
Just a question. Does this make more or less sense to move to bdk-python repo?

thunderbiscuit:
I'm not sure I understand what you mean. Where do you have in mind would be best to have it? 

At the moment the bdk-python library is one of my repos ([here](https://github.com/thunderbiscuit/bdk-python)), but soon it will be under the bitcoindevkit org like bdk-swift and bdk-kotlin.

afilini:
> Build the package using the typical python -m build

Just a sidenote, I've always used `python3 setup.py -v bdist_wheel` to build the wheel, but I don't think it makes any difference really

> Does this make more or less sense to move to bdk-python repo?

I initially thought about moving this to `bdk-python`, but since there's no rust code at all in there we would have had to create a new cargo project just for this, which seemed wasteful. In theory we could make it very modular and keep everything in here, for all the languages. It will also avoid us breaking things downstream (i.e. we make a change in bdk-ffi that accidentally breaks bdk-python) because everything is in one repo and it can be tested here

artfuldev:
Initially I had this idea of separate repos for each language target, with rust code and modifications for each, even with slightly different udls, and some custom code (like language specific helpers) but sharing the same core. I was just thinking from that angle. I haven't been able to do anything towards this though. In the meanwhile we started using git submodules instead of cargo dependencies for the different ffi projects. IIRC, we did the bdk-swift due to Swift packaging system using git repos and I guess we now have separate repos for everything but not quite in the way I imagined. So just thought about it again when I saw a similar idea (shared rust core and different commands to build custom versions). But I understand the language specific repos are actually free from rust code.

afilini:
I haven't looked closely at the bindings yet, so if you feel like it's better to split this up then sure, we can do it! I would just try to avoid as much as we can to duplicate code, otherwise it's just gonna be very hard to maintain.

I thought with uniffi we wouldn't need too many language specific stuff, apart from the way we generate them.

artfuldev:
@thunderbiscuit I think there's always going to be rust on the libraries' side of things, and there's no need to abstract that away. Today, the rust side of things is inside a folder `bdk-ffi` in the language-specific repositories - whether it's a sub-module or not is an implementation detail.

I agree with the thin rust core, but I think there are more than 3 well-defined layers. I understand the concern for a consistent API across SDKs in target languages - I didn't mean to violate that, so let me take back my remark about _slightly different_ UDLs: it seems to have caused some unnecessary confusion.

I'm all for creating language-specific extensions which stay in the language-specific repositories. Which is where this PR made me think - if this is specific to the python language, why does it reside here? But it's just a question and not a blocking concern. We can always merge this PR and revisit the topic a little later.

artfuldev:
@afilini I also think only the way we generate and/or package stuff would change.

artfuldev:
What do we think of passing the language and output directory as CLI args? like:
```sh
cargo run --bin generate --lang=python --out=./out
```

notmandatory:
I agree with @artfuldev that ideally we'd be able to have all language specific stuff in their respective repos but it's not a blocking issue for this PR and if we can find a better way to organize stuff we can do it in a future PR.  

As for using arguments to generate (ie. `--out ./out`) vs using env vars right now with `build.sh` scripts in the lang repos who whould run this `generate` bin, using env vars makes sense for parameters. In the future this could change if decide to create unique version of the `generate` bin for each language that replace the existing `build.sh` scripts. But for where our repos are now this makes sense.

afilini:
> What do we think of passing the language and output directory as CLI args

I considered this, but I didn't want to add extra dependencies to parse the command line, so the env variables were just the quickest alternative to implement

thunderbiscuit:
After discussing with @notmandatory and @artfuldev, I do think it makes sense to eventually consider leaving only the generic parts of the language bindings (the `udl` and `lib.rs` files) in `bdk-ffi`, and bring all language specific things in their respective repositories. This would keep the API easy to maintain while not muddling the ffi repo with language-specific issues like deployment and quirks like what is needed on the Python side.

Whether we want to merge this first and later on consider pulling all of that out (including the `build.sh` file) in the specific repos or try to do that right from the get go I'm indifferent towards. I just think it would be nice to have the Python bindings published soonish so that we can test/play with it, and prep for the potential hackathon event (thinking sometime in April/May) which I think would benefit from having it live.

artfuldev:
I think there's no reason to hold off on merging this PR @notmandatory

PR #104, opened 2022-02-25
Add PSBT deserialize and serialize functions, remove details

### LINKED ISSUE #103: Build fails when `--features generate-python` not used
The `generate-python` feature is optional and not default, but if it's not selected the `bin` target does not build because the required (but optional) `uniffi_bindgen` dependency is not included. 

This should be easy to fix, but in the meantime a work around is to build with `--features generate-python`, even for kotlin or swift projects.

notmandatory:
1. Fix bin/generate with no features
2. Add `PartiallySignedBitcoinTransaction::deserialize` function as named constructor to decode from a string per [BIP 0174]
3. Add `PartiallySignedBitcoinTransaction::serialize` function to encode to a string per [BIP 0174]
4. Remove `PartiallySignedBitcoinTransaction.details` struct field

[BIP 0174]:https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#encoding

Fixes #103

### DIFF
--- a/src/bdk.udl
+++ b/src/bdk.udl
@@ -131,6 +131,9 @@ interface Wallet {
 interface PartiallySignedBitcoinTransaction {
   [Throws=BdkError]
   constructor([ByRef] Wallet wallet, string recipient, u64 amount, float? fee_rate);
+  [Name=deserialize,Throws=BdkError]
+  constructor(string psbt_base64);
+  string serialize();
 };
 
 dictionary ExtendedKeyInfo {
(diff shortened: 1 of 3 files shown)


thunderbiscuit:
ACK 7c91413.

I'm not clear on what the serialize and deserialize methods add/provide. Mind expanding on that?

notmandatory:
Oh ya sorry to be clear these methods convert a PSBT to/from a string format that can be passed around, it uses a standard base-64 character set and is specified in the [BIP 0174 encoding](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#encoding) section.

thunderbiscuit:
I see. Yes that's necessary for exporting and importing psbts in string format. Cheers!

PR ACK.

PR #111, opened 2022-03-02
Remove hard coded sync progress value

### LINKED ISSUE #108: BdkProgress returns non 100 value when complete
When syncing a wallet using
`try wallet.sync(progressUpdate: self.syncProgress, maxAddressParam: nil)` where self.syncProgress is a Progress() object:

```
class Progress : BdkProgress {
    var value = Float()
    func update(progress: Float, message: String?) {
        value = progress
        print("progress", progress, message as Any)
    }
}
```

I am receiving a value of `21` but was expecting `100`, indicating a fully synced wallet.

notmandatory:
Fixes #108 

When `bdk-ffi` was only a prototype a value was added to test that the `wallet.sync` `progress_update` callback was working. This PR removes that hard coded value (21.0). Some bdk blockchains do not return a progress update and will never call the `progress_update` callback.  

The blockchain clients that update progress are: `compact_filters` and `rpc`. And the ones that don't are: `electrum` and `esplora`. Since right now only `electrum` and `esplora` are supported for `bdk-ffi` the `progress_update` call back isn't that useful, all calls to `wallet.sync` should sync return right away without reporting any progress.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -259,7 +259,6 @@ impl Wallet {
         progress_update: Box<dyn BdkProgress>,
         max_address_param: Option<u32>,
     ) -> Result<(), BdkError> {
-        progress_update.update(21.0, Some("message".to_string()));
         self.get_wallet()
             .sync(BdkProgressHolder { progress_update }, max_address_param)
     }


notmandatory:
Moving this to a `0.3.1` bug fix tag so I can include it in a new `bdk-swift` release.

PR #115, opened 2022-03-11
Add flush method to wallet

### LINKED ISSUE #114: Database is getting corrupted on Android
I assume this is related to the `flush()` method Ricardo implemented a little while back.

I cannot seem to reproduce the problem consistently which doesn't help, but the gist of it is that at some point the sats sent out are not recognized as having been spent, and no amount of `sync` can fix the issue. The only way to fix is to delete and reload/resync the wallet from scratch.

I think we should try to flush every time we sync, but I haven't been able to find examples of that, and my attempts currently don't compile.

thunderbiscuit:
This PR cannot be merged before the `flush()` method is added to bdk (see upcoming PR)

This fixes #114. Took @notmandatory and I half a day to do it all, but we fully tested the new flush method and it did resolve the bug.

Note that at the moment the `database()` method on the wallet returns an immutable reference, hence why we cannot simply do:
```rust
fn flush(
    &self,
) -> Result<(), Error> {
    self.get_wallet().database().borrow().flush()
}
```
directly in bdk-ffi to trigger the flush.

### DIFF
--- a/src/bdk.udl
+++ b/src/bdk.udl
@@ -126,6 +126,8 @@ interface Wallet {
   void sync(BdkProgress progress_update, u32? max_address_param);
   [Throws=BdkError]
   Transaction broadcast([ByRef] PartiallySignedBitcoinTransaction psbt);
+  [Throws=BdkError]
+  void flush();
 };
 
 interface PartiallySignedBitcoinTransaction {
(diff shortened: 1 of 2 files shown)


notmandatory:
I think this one can be closed in favor of using sqlite db from #116.

thunderbiscuit:
Agreed.

PR #120, opened 2022-03-16
Fix Wallet.broadcast function, now returns a tx id as a hex string

notmandatory:


thunderbiscuit:
I have been using the fact that the `broadcast()` method was returning a `Transaction` data class in my two apps, mostly parsing it for total fees and txid (plus confirmation status but that's always pending upon broadcasting). I guess I could get that data from the `get_transactions()` method, which will return it as part of the list.

I'll think it over and test some things out in my wallets. Maybe decoupling the broadcast and the resulting `Transaction` data structure isn't a bad thing.

thunderbiscuit:
Ok after looking at bdk that's really [what the `broadcast()` function returns](https://docs.rs/bdk/latest/bdk/wallet/struct.Wallet.html#method.broadcast). I guess we were just sugaring it up a little bit and I didn't know haha! I think it's better to stay as close to the BDK API as we can.

Technical ACK. Will test tomorrow.

thunderbiscuit:
Tested ACK f51f612.

Issue #86, opened 2021-11-12
Update transaction building API to enable spend all / sweep


notmandatory:
To send all / sweep all coins from a wallet add to the TxBuilder:

1. **TxBuilder drain_to(string address)** see `bdk` [wallet::tx_builder::TxBuilder::drain_to](https://docs.rs/bdk/latest/bdk/wallet/tx_builder/struct.TxBuilder.html#method.drain_to)

2. **TxBuilder drain_wallet()** see `bdk` [wallet::tx_builder::drain_wallet](https://docs.rs/bdk/latest/bdk/wallet/tx_builder/struct.TxBuilder.html#method.drain_wallet)

PR #123, opened 2022-03-25
Add an initial version of TxBuilder

### LINKED ISSUE #24: Add TxBuilder

artfuldev:
Fixes #24 

* Adds an initial version of TxBuilder
* Removes PSBT constructor (we needed this because that was the only way to create a PSBT
* Removes unnecessary `pub` keywords in PSBT `impl`
* Removes a `.to_owned()` call due to `cargo clippy` complaints
* We can add another commit to add functionality related to #86

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -310,7 +310,7 @@ impl TxBuilder {
         let wallet = wallet.get_wallet();
         let mut tx_builder = wallet.build_tx();
         for (address, amount) in &self.recipients {
-            let address = Address::from_str(address).expect("recipient address parsing failed");
+            let address = Address::from_str(address).map_err(|e| BdkError::Generic(e.to_string()))?;
             tx_builder.add_recipient(address.script_pubkey(), *amount);
         }
         if let Some(sat_per_vb) = self.fee_rate {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -322,7 +322,6 @@ impl TxBuilder {
                 internal: Mutex::new(psbt),
             })
             .map(Arc::new)
-            .map_err(|_| BdkError::Generic("Failed to build transaction".to_string()))
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -112,6 +112,7 @@ impl Progress for BdkProgressHolder {
     }
 }
 
+#[derive(Debug)]
 struct PartiallySignedBitcoinTransaction {
     internal: Mutex<PartiallySignedTransaction>,
 }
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -303,4 +277,53 @@ fn restore_extended_key(
...
+            recipients: self.recipients.to_vec(),
+            fee_rate: Some(sat_per_vb),
+        })
+    }
+
+    fn build(&self, wallet: &Wallet) -> Result<Arc<PartiallySignedBitcoinTransaction>, Error> {
+        let wallet = wallet.get_wallet();
+        let mut tx_builder = wallet.build_tx();
+        for (address, amount) in &self.recipients {
+            let address = Address::from_str(address).expect("recipient address parsing failed");
```
notmandatory, line 313: Can we move this to the `add_recipient` function so validation is done there (with possible error Result)?

src/lib.rs
```diff
@@ -303,4 +277,53 @@ fn restore_extended_key(
...
+        if let Some(sat_per_vb) = self.fee_rate {
+            tx_builder.fee_rate(FeeRate::from_sat_per_vb(sat_per_vb));
+        }
+        tx_builder
+            .finish()
+            .map(|(psbt, _)| PartiallySignedBitcoinTransaction {
+                internal: Mutex::new(psbt),
+            })
+            .map(Arc::new)
+            .map_err(|_| BdkError::Generic("Failed to build transaction".to_string()))
```
notmandatory, line 325: I think all we need to do is add the new possible BdkError enums to the `bdk.udl` file and then they won't have to be mapped to the `BdkError::Generic`.

notmandatory:
Overall looks great! I'll give it a more thorough testing and review tomorrow. For testing I usually point my local `bdk-kotlin` repo to a local checkout of your `bdk-ffi` branch, and then update the tests there, build, and verify with the Kotlin JVM and Android tests.

thunderbiscuit:
I'll test on Monday as well.

src/lib.rs
artfuldev: We can, but let's talk about this today. Having a result chain breaks the builder pattern. We either want to fail at every step or expose a builder. Having failing intermediate steps is not usually how a builder is exposed.

src/lib.rs
artfuldev: I think this may have been an oversight. If the error that `tx_builder.finish()` exposes is only the `BdkError` then we don't need this transform, as the `BdkError` enum is already fully described in our `bdk.udl` file. I can check again if there are other cases and if the `map_err` can be safely ignored.

artfuldev:
Thanks @notmandatory @thunderbiscuit

src/lib.rs
notmandatory: I did a little more reading on the builder pattern and I think you're right that it's more user friendly to chain setters together without having to worry about any input validation errors until the build end. You can disregard my suggestion. :slightly_smiling_face:

notmandatory (approved):
Tested ACK bc43d2eb1a74cf05d110f97ecc8e7c0a472bfec8

Have https://github.com/bitcoindevkit/bdk-kotlin/pull/40 ready to add tests for this.

PR #137, opened 2022-04-08
Add address index to return type of get_address

### LINKED ISSUE #130: Feature request: address index
It would be quite useful to get the index of the derived addresses when using the `get_new_address()` and `get_last_unused_address()` methods.

This index is held in the `AddressInfo` struct:
```rust
/// A derived address and the index it was found at
/// For convenience this automatically derefs to `Address`
#[derive(Debug, PartialEq)]
pub struct AddressInfo {
    /// Child index of this address
    pub index: u32,
    /// Address
    pub address: Address,
    /// Type of keychain
    pub keychain: KeychainKind,
}
```

thunderbiscuit:
This is my first stab at adding an API from bdk into to bdk-ffi. Looking for feedback. In particular, I found I had to create a brand new struct in `lib.rs` (`AddressInformation`) whose purpose was mostly just to mirror the `AddressInfo` struct coming from bdk. I assume there is a better way and I should be able to use the type from bdk directly; please let me know if so.

This PR is marked as draft because if the approach is agreed upon I'd still need to change the `get_new_address()` method to follow the same API. I assume also worthy of discussion is whether we might as well bring the whole of `AddressInfo` into bdk-ffi, which would add the `KeychainKind` field (and associated type) to `AddressInfo`/`AddressInformation`.

This PR fixes #130.

### Notes to reviewers
This PR changes the return type of `get_last_unused_address()` from `String` to `AddressInfo`.

```rust
pub struct AddressInfo {
    pub index: u32,
    pub address: String,
}
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,12 +11,13 @@ use bdk::database::any::{AnyDatabase, SledDbConfiguration, SqliteDbConfiguration
 use bdk::database::{AnyDatabaseConfig, ConfigurableDatabase};
 use bdk::keys::bip39::{Language, Mnemonic, WordCount};
 use bdk::keys::{DerivableKey, ExtendedKey, GeneratableKey, GeneratedKey};
+use bdk::wallet::AddressInfo as BdkAddressInfo;
+use bdk::wallet::AddressIndex as BdkAddressIndex;
 use bdk::miniscript::BareCtx;
+use std::convert::{TryFrom, From};
 use bdk::{
     BlockTime, Error, FeeRate, SignOptions, SyncOptions as BdkSyncOptions, Wallet as BdkWallet,
 };
-use std::convert::TryFrom;
-use std::convert::From;
 use std::fmt;
 use std::ops::Deref;
 use std::str::FromStr;
@@ -26,15 +27,12 @@ uniffi_macros::include_scaffolding!("bdk");
 
 type BdkError = Error;
 
-// we redefine a simpler version of bdk::wallet::AddressInfo
-// it has an `address` field of type String
-// instead of the `address` field of type `Address` defined in bitcoin::util::address::Address
 pub struct AddressInfo {
     pub index: u32,
     pub address: String,
 }
 
-impl From<bdk::wallet::AddressInfo> for AddressInfo {
+impl From<BdkAddressInfo> for AddressInfo {
     fn from(x: bdk::wallet::AddressInfo) -> AddressInfo {
         AddressInfo {
             index: x.index,
@@ -43,18 +41,16 @@ impl From<bdk::wallet::AddressInfo> for AddressInfo {
     }
 }
 
-// we redefine a simpler version of bdk::wallet::AddressIndex
-// only keeping the `New` and `LastUnused` variants of the enum
 pub enum AddressIndex {
     New,
     LastUnused,
 }
 
-impl From<AddressIndex> for bdk::wallet::AddressIndex {
-    fn from(x: AddressIndex) -> bdk::wallet::AddressIndex {
+impl From<AddressIndex> for BdkAddressIndex {
+    fn from(x: AddressIndex) -> BdkAddressIndex {
         match x {
-            AddressIndex::New => bdk::wallet::AddressIndex::New,
-            AddressIndex::LastUnused => bdk::wallet::AddressIndex::LastUnused
+            AddressIndex::New => BdkAddressIndex::New,
+            AddressIndex::LastUnused => BdkAddressIndex::LastUnused
         }
     }
 }
@@ -269,25 +265,10 @@ impl Wallet {
     }
 
     fn get_address(&self, address_index: AddressIndex) -> Result<AddressInfo, BdkError> {
-        match self.get_wallet().get_address(bdk::wallet::AddressIndex::from(address_index)) {
-            Ok(address_info) => Ok(AddressInfo::from(address_info)),
-            Err(bdk_error) => Err(bdk_error),
-        }
+        self.get_wallet()
+            .get_address(address_index.into())
+            .map(AddressInfo::from)
     }
-
-    // fn get_new_address(&self) -> Result<AddressInformation, BdkError> {
-    //     match self.get_wallet().get_address(AddressIndex::New) {
-    //         Ok(address_info) => Ok(AddressInformation::from(address_info)),
-    //         Err(BdkError) => Err(BdkError),
-    //     }
-    // }
-
-    // fn get_last_unused_address(&self) -> Result<AddressInformation, BdkError> {
-    //     match self.get_wallet().get_address(AddressIndex::LastUnused) {
-    //         Ok(address_info) => Ok(AddressInformation::from(address_info)),
-    //         Err(BdkError) => Err(BdkError),
-    //     }
-    // }
 
     fn get_balance(&self) -> Result<u64, Error> {
         self.get_wallet().get_balance()
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13,6 +13,7 @@ use bdk::keys::bip39::{Language, Mnemonic, WordCount};
 use bdk::keys::{DerivableKey, ExtendedKey, GeneratableKey, GeneratedKey};
 use bdk::miniscript::BareCtx;
 use bdk::wallet::AddressIndex;
+use bdk::wallet::AddressInfo;
 use bdk::{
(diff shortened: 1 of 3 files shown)


thunderbiscuit:
Major update to this PR:

This now defines a method `get_address()` similar to the one in bdk, which takes an enum `AddressIndex` which specifies whether to return a "new" address or a "last unused" address.

I don't use the full `AddressIndex` enum from bdk which has the `Peek(u32)` and `Reset(u32)` variants, but instead define a subset of it with only the `New` and `LastUnused` variants.

thunderbiscuit:
@artfuldev @notmandatory namespacing here is unclear for me, and I'm wondering if you have ideas to make sure it's consistent or insights in one way or another. Here is the problem as I see it:

It is sometimes the case that we can't expose a struct fully, because it itself uses another struct which itself uses another struct and so on, which would all need to be defined/exposed in the ffi layer but are not always relevant to the task at hand. In such cases we can define "simplified" versions of those structs which flatten the complexity. But now those new structures need naming, and the bdk names are quite well chosen, so it's a bit of a tricky situation, also because from the user's point of view they now have to know that `AddressInformation` actually maps to `AddressInfo` and the like.

Here is what I've done in this PR. Not sure if it was a clean/good idea or if something else might be better, but let me know what you think. In cases where I needed to simplify a struct, I kept/stole the name from bdk and redefined a simpler struct, and then wrote the conversion methods (thanks for the tip @artfuldev!) which pull from the fully qualified bdk struct. It looks something like this:

```rust
// we redefine a simpler version of bdk::wallet::AddressInfo
// it has an `address` field of type String
// instead of the `address` field of type `Address` defined in bitcoin::util::address::Address
pub struct AddressInfo {
    pub index: u32,
    pub address: String,
}

impl From<bdk::wallet::AddressInfo> for AddressInfo {
    fn from(x: bdk::wallet::AddressInfo) -> AddressInfo {
        AddressInfo {
            index: x.index,
            address: x.address.to_string()
        }
    }
}
```

artfuldev:
This is fine. We can retain the names and use bdk namespace qualifiers to refer to the original types. I don't think it's a deviation. We do the same for transaction, TxBuilder, and for the database and blockchain configuration types.

thunderbiscuit:
One more thing I'm not so sure on is the use or not of a reference in the definition of the `From` trait. The current code for `TransactionDetails` is this:
```rust
 impl From<&bdk::TransactionDetails> for TransactionDetails {
    fn from(x: &bdk::TransactionDetails) -> TransactionDetails {
        TransactionDetails {
            fees: x.fee,
            txid: x.txid.to_string(),
            received: x.received,
            sent: x.sent,
        }
    }
}
```

But mine doesn't take a reference:
```rust
impl From<bdk::wallet::AddressInfo> for AddressInfo {
    fn from(x: bdk::wallet::AddressInfo) -> AddressInfo {
        AddressInfo {
            index: x.index,
            address: x.address.to_string()
        }
    }
}
```

I'm unclear as per the tradeoff here between the two.

src/lib.rs
```diff
@@ -12,16 +12,51 @@ use bdk::database::{AnyDatabaseConfig, ConfigurableDatabase};
 use bdk::keys::bip39::{Language, Mnemonic, WordCount};
 use bdk::keys::{DerivableKey, ExtendedKey, GeneratableKey, GeneratedKey};
 use bdk::miniscript::BareCtx;
-use bdk::wallet::AddressIndex;
+// use bdk::wallet::AddressIndex;
+// use bdk::wallet::AddressInfo;
```
artfuldev, line 16: Can we remove this?

src/lib.rs
```diff
@@ -12,16 +12,51 @@ use bdk::database::{AnyDatabaseConfig, ConfigurableDatabase};
...
 use bdk::{BlockTime, Error, FeeRate, SignOptions, Wallet as BdkWallet};
 use std::convert::TryFrom;
+use std::convert::From;
```
artfuldev, line 19: We can merge these into
```
use std::convert::{TryFrom, From};
```

src/lib.rs
```diff
@@ -193,21 +228,26 @@ impl Wallet {
...
+    //         Err(BdkError) => Err(BdkError),
+    //     }
+    // }
+
+    // fn get_last_unused_address(&self) -> Result<AddressInformation, BdkError> {
+    //     match self.get_wallet().get_address(AddressIndex::LastUnused) {
+    //         Ok(address_info) => Ok(AddressInformation::from(address_info)),
+    //         Err(BdkError) => Err(BdkError),
+    //     }
+    // }
```
artfuldev, line 250: We can remove these

src/lib.rs
```diff
@@ -193,21 +228,26 @@ impl Wallet {
...
-        self.get_wallet()
-            .get_address(AddressIndex::New)
-            .unwrap()
-            .address
-            .to_string()
+    fn get_address(&self, address_index: AddressIndex) -> Result<AddressInfo, BdkError> {
+        match self.get_wallet().get_address(bdk::wallet::AddressIndex::from(address_index)) {
+            Ok(address_info) => Ok(AddressInfo::from(address_info)),
+            Err(bdk_error) => Err(bdk_error),
+        }
```
artfuldev, line 235: we can rewrite this with a [`Result::map`](https://doc.rust-lang.org/std/result/enum.Result.html#method.map) and an [`Into::into`](https://doc.rust-lang.org/std/convert/trait.Into.html):
```rust
self.get_wallet()
    .get_address(address_index.into())
    .map(AddressInfo::from)
```

artfuldev (changes requested):
I've added a few minor improvements

artfuldev:
We probably have to rebase after the bdk 0.18 update.

thunderbiscuit:
Yeah I think we should wait until #150 and #148 are in and then I'll be able to finish the rebase. If I'm quick enough we can even fit it in `0.6.0`.

src/lib.rs
thunderbiscuit: Done.

src/lib.rs
```diff
@@ -12,16 +12,48 @@ use bdk::database::{AnyDatabaseConfig, ConfigurableDatabase};
...
 use std::str::FromStr;
 use std::sync::{Arc, Mutex, MutexGuard};
 
 uniffi_macros::include_scaffolding!("bdk");
 
 type BdkError = Error;
 
+// we redefine a simpler version of bdk::wallet::AddressInfo
+// it has an `address` field of type String instead of the `address` field of type `Address`
+// defined in bitcoin::util::address::Address
```
artfuldev, line 26: Do we need this? I think it's fairly evident that we're shadowing/proxying here. If it's notes for the reviewers, we can add them as comments instead.

src/lib.rs
```diff
@@ -12,16 +12,48 @@ use bdk::database::{AnyDatabaseConfig, ConfigurableDatabase};
...
+    fn from(x: bdk::wallet::AddressInfo) -> AddressInfo {
+        AddressInfo {
+            index: x.index,
+            address: x.address.to_string()
+        }
+    }
+}
+
+// we redefine a simpler version of bdk::wallet::AddressIndex
+// only keeping the `New` and `LastUnused` variants of the enum
```
artfuldev, line 42: Same as above

src/lib.rs
```diff
@@ -12,16 +12,48 @@ use bdk::database::{AnyDatabaseConfig, ConfigurableDatabase};
...
+pub enum AddressIndex {
+    New,
+    LastUnused,
+}
+
+impl From<AddressIndex> for bdk::wallet::AddressIndex {
+    fn from(x: AddressIndex) -> bdk::wallet::AddressIndex {
+        match x {
+            AddressIndex::New => bdk::wallet::AddressIndex::New,
+            AddressIndex::LastUnused => bdk::wallet::AddressIndex::LastUnused
```
artfuldev, line 52: nitpick: we can also alias `bdk::wallet::AddressIndex` as `BdkAddressIndex`

src/lib.rs
thunderbiscuit: Yeah agreed I'll remove them.

src/lib.rs
thunderbiscuit: I like it for consistency with the others!

thunderbiscuit:
Thanks for the comments @artfuldev!

artfuldev (approved):
LGTM! 🎉

PR #139, opened 2022-04-12
Add sign / verify for arbitrary messages

q-src:
This PR adds support for the creation of signatures of arbitrary messages and their verification. A PR for bdk-kotlin will follow.

Signatures can be created using a given `wif`, `prv` or `xprv`. They can be verified using a `pub`, `xpub` or `address`. 

Since these are my very first steps with Rust, please take a closer look at the coding style as well.

### DIFF
--- a/src/bdk.udl
+++ b/src/bdk.udl
@@ -144,6 +144,16 @@ interface EcdsaMessageSigner {
   string sign(string msg);
 };
 
+interface EcdsaMessageSignatureVerifier {
+  [Throws=BdkError, Name=from_pub]
+  constructor(string pubkey);
+  [Throws=BdkError, Name=from_xpub]
+  constructor(string xpub);
+  [Throws=BdkError, Name=from_address]
+  constructor(string address);
+  boolean verify(string sig, string msg);
+};
+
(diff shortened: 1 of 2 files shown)


notmandatory:
This looks cool! I'll take a look and it might be something we want to have in `bdk` eventually.

thunderbiscuit:
@q-src we've discussed this in 2 dev calls and think it might be valuable to have as part of the core Rust bitcoindevkit. Are you interested in opening a PR to add the feature right in bdk?

Also our dev calls are open and bi-weekly, so if ever that's something you're interested in joining we'd love to have you. You can join our [Discord](https://discord.com/invite/d7NkDKm) and you'll find all the info there. Cheers!

thunderbiscuit:
Awesome! You can find the notes and proposed agenda for the upcoming calls both for the core Rust library and the language bindings [here](https://github.com/bitcoindevkit/.github/issues).

I keep the agenda for the language bindings evergreen [here](https://hackmd.io/@thunderbiscuit/devcall).

thunderbiscuit:
Closing in favour of bitcoindevkit/bdk#601.

PR #140, opened 2022-04-15
Add RBF methods to TxBuilder

### LINKED ISSUE #133: Add TxBuilder RBF methods
Add methods to TxBuilder to allow users to control RBF.

See https://docs.rs/bdk/latest/bdk/wallet/tx_builder/struct.TxBuilder.html for docs on what the functions do. 

```udl
interface TxBuilder {
  ... 
  TxBuilder enable_rbf();
  TxBuilder enable_rbf_with_sequence(u32 nsequence);
}

```

artfuldev:
Fix #133 

We need to create another `RbfValue` enum because the actual type is only visible within the crate in `bdk`.

### DIFF
--- a/src/bdk.udl
+++ b/src/bdk.udl
@@ -146,6 +146,8 @@ interface TxBuilder {
   TxBuilder fee_rate(float sat_per_vbyte);
   TxBuilder drain_wallet();
   TxBuilder drain_to(string address);
+  TxBuilder enable_rbf();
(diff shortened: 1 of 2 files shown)


notmandatory (approved):
tACK 220835cffd9288fc6dbccd27551a3c9b25d74a95

Looks good, tested manually via bdk-kotlin and works as expected.

PR #148, opened 2022-04-20
Add Wallet and TxBuilder tests using test funded wallet

### LINKED ISSUE #141: Add tests for TxBuilder API
We need to add tests for the FFI layer of TxBuilder API as we're dealing with some local state. We need a few test wallets which are funded in order to test. cc @notmandatory

notmandatory:
This PR adds an example test using the `bdk` test funded wallet.  The example test makes sure the `bdk-ffi` `TxBuilder` is able to drain a single wallet UTXO to a single address.  More tests can be added as we need them in future PRs.

Required to complete #141

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ crate-type = ["staticlib", "cdylib"]
 name = "bdkffi"
 
 [dependencies]
-bdk = { version = "0.18", features = ["all-keys", "use-esplora-ureq", "sqlite-bundled"] }
+bdk = { version = "0.19", features = ["all-keys", "use-esplora-ureq", "sqlite-bundled"] }
 
 uniffi_macros = { version = "0.16.0", features = ["builtin-bindgen"] }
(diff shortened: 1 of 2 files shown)


artfuldev:
Super cool. If we get a funded wallet with a descriptor I can wrap up #141

notmandatory:
This is on hold until I can get https://github.com/bitcoindevkit/bdk/pull/600 released in `bdk`.

notmandatory:
This should be merged after #162 .

thunderbiscuit (commented):
Concept ACK. Wondering if we can break the tests into a different file just to keep the lib.rs from getting too big.

notmandatory:
> Concept ACK. Wondering if we can break the tests into a different file just to keep the lib.rs from getting too big.

I know it looks strange to have such big files with tests at the end, but this is the canonical place to put [Rust unit tests](https://doc.rust-lang.org/book/ch11-03-test-organization.html#unit-tests).

PR #149, opened 2022-04-22
Update to bdk 0.18.0

notmandatory:
Changes that were needed to match updated bdk APIs:
* new bdk-ffi `Blockchain`  interface with `constructo(BlockchainConfig)` and `broadcast()` functions
* `Blockchain.broadcast()` function is now void (doesn't return txid)
* added `PartiallySignedBitcoinTransaction.txid()` getter function
* `Wallet.sync()` function now takes a `Blockchain` and optional `Progress` callback interface

Tests performed:
* from `bdk-kotlin/bdk-ffi` added local remote to this branch, ran build.sh script
* fixed broken `bdk-kotlin` jvm and android tests and confirmed tests all pass

### DIFF
--- a/src/bdk.udl
+++ b/src/bdk.udl
@@ -116,7 +116,7 @@ interface Blockchain {
   [Throws=BdkError]
   constructor(BlockchainConfig config);
   [Throws=BdkError]
-  string broadcast([ByRef] PartiallySignedBitcoinTransaction psbt);
+  void broadcast([ByRef] PartiallySignedBitcoinTransaction psbt);
 };
 
 callback interface Progress {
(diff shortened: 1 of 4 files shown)


notmandatory:
Per review with @thunderbiscuit removed unneeded `BdkError` type alias.

src/bdk.udl
```diff
@@ -129,15 +136,14 @@ interface Wallet {
...
-  [Throws=BdkError]
-  string broadcast([ByRef] PartiallySignedBitcoinTransaction psbt);
+  void sync([ByRef] Blockchain blockchain, Progress? progress);
 };
 
 interface PartiallySignedBitcoinTransaction {
   [Throws=BdkError]
   constructor(string psbt_base64);
   string serialize();
+  string txid();
```
thunderbiscuit, line 146: Just a question: doesn't the txid change over the course of the lifetime of the psbt if the inputs are not using SegWit? Could that be a problem here?

thunderbiscuit (approved):
Tested ACK. Just one small question regarding getting txids from unfinished PSBTs.

PR #150, opened 2022-05-03
Add new BumpFeeTxBuilder interface

notmandatory:
Add BumpFeeTxBuilder to bump the fee on an unconfirmed tx created by the Wallet. The structure of the new interface is:

```udl
interface BumpFeeTxBuilder {
  constructor(string txid, float new_fee_rate);
  BumpFeeTxBuilder allow_shrinking(string address);
  BumpFeeTxBuilder enable_rbf();
  BumpFeeTxBuilder enable_rbf_with_sequence(u32 nsequence);
  [Throws=BdkError]
  PartiallySignedBitcoinTransaction build([ByRef] Wallet wallet);
};
```
Fixes #150

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -406,7 +406,7 @@ impl TxBuilder {
         })
     }
 
-    fn build(&self, wallet: &Wallet) -> Result<Arc<PartiallySignedBitcoinTransaction>, Error> {
+    fn finish(&self, wallet: &Wallet) -> Result<Arc<PartiallySignedBitcoinTransaction>, Error> {
         let wallet = wallet.get_wallet();
         let mut tx_builder = wallet.build_tx();
         for (address, amount) in &self.recipients {
@@ -484,7 +484,7 @@ impl BumpFeeTxBuilder {
         })
     }
 
-    fn build(&self, wallet: &Wallet) -> Result<Arc<PartiallySignedBitcoinTransaction>, Error> {
+    fn finish(&self, wallet: &Wallet) -> Result<Arc<PartiallySignedBitcoinTransaction>, Error> {
         let wallet = wallet.get_wallet();
         let txid = Txid::from_str(self.txid.as_str())?;
         let mut tx_builder = wallet.build_fee_bump(txid)?;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -443,8 +443,8 @@ impl TxBuilder {
 struct BumpFeeTxBuilder {
     txid: String,
     fee_rate: f32,
-    rbf: Option<RbfValue>,
     allow_shrinking: Option<String>,
+    rbf: Option<RbfValue>,
 }
 
 impl BumpFeeTxBuilder {
@@ -470,8 +470,8 @@ impl BumpFeeTxBuilder {
         Arc::new(Self {
             txid: self.txid.clone(),
             fee_rate: self.fee_rate,
-            rbf: Some(RbfValue::Default),
             allow_shrinking: self.allow_shrinking.clone(),
+            rbf: Some(RbfValue::Default),
         })
     }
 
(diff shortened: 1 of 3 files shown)


notmandatory:
I also need to rebase this PR after #149 is merged.

notmandatory:
@SeverinAlexB this one is ready to review, let me know if the API makes sense for you.

src/lib.rs
```diff
@@ -440,4 +440,78 @@ impl TxBuilder {
...
+    fn enable_rbf_with_sequence(&self, nsequence: u32) -> Arc<Self> {
+        Arc::new(Self {
+            txid: self.txid.clone(),
+            fee_rate: self.fee_rate,
+            rbf: Some(RbfValue::Value(nsequence)),
+            allow_shrinking: self.allow_shrinking.clone(),
+        })
+    }
+
+    fn build(&self, wallet: &Wallet) -> Result<Arc<PartiallySignedBitcoinTransaction>, Error> {
```
thunderbiscuit, line 487: The FFI layer uses `build` whereas the bdk library uses `finish`. Is there a reason for the difference in name? Wondering if we should fix that up before releasing 0.6.0.

src/lib.rs
```diff
@@ -440,4 +440,78 @@ impl TxBuilder {
     }
 }
 
+struct BumpFeeTxBuilder {
+    txid: String,
+    fee_rate: f32,
+    rbf: Option<RbfValue>,
+    allow_shrinking: Option<String>,
```
thunderbiscuit, line 447: Small nit: the order of the `rbf` and `allow_shrinking` parameters are switched between the struct declaration and the different methods (sometimes rbf is first sometimes second).

src/lib.rs
```diff
@@ -440,4 +440,78 @@ impl TxBuilder {
...
+}
+
+impl BumpFeeTxBuilder {
+    fn new(txid: String, fee_rate: f32) -> Self {
```
thunderbiscuit, line 451: The `new` associated function returns `Self` whereas the other 3 methods return `Arc<Self>`. Could that create a problem somewhere? Or can the `build` method consume both `self` and `Arc<Self>`?

thunderbiscuit (approved):
Tested ACK. Looks good to me, I only have a few minor questions, mostly to clear up my understanding.

src/lib.rs
notmandatory: Yes good point, I think we should keep function names in sync where ever we can. Fixed.

src/lib.rs
notmandatory: Good catch, for readability/maintainability it does matter. Fixed.

src/lib.rs
notmandatory: My understanding is that since `Arc<T>`  implements `AsRef<T>` that is can be used as a &T, or &self in this case.  The reason for returning an `Arc<T>` on a function that returns Self is specific to how uniffi works, https://mozilla.github.io/uniffi-rs/udl/interfaces.html#managing-shared-references. And not needing to return Arc<Self> for the constructor we also don't do because uniffi doesn't require it there. There could be a more efficient way to do this were we don't allocate a whole new builder on each call, but given the short life of these structs anyway I don't think it's a problem.

SeverinAlexB:
Looks good to me. I am not 100% sure what `allow_shrinking` does but I guess it's about manually selecting outputs that can be changed.

thunderbiscuit:
@SeverinAlexB that's correct.

The default behaviour of bdk with RBF is to increase the fee by reducing one of the change outputs (if there are any). In the case where there aren't any change outputs or the change output would be below the dust limit, it will look into the wallet to add an extra input so it can create more change, and from that change pull more satoshis for fees. This is done in order to preserve the exact amount "paid". `allow_shrinking` enables you to specify the exact UTXO you want to reduce in order to leave more for fees (one example of that would be if you want the _recipient_ of the payment to pay for the extra fees).

PR #153, opened 2022-05-12
Simplify using struct update syntax and clone

artfuldev:


notmandatory (approved):
ACK e5cd7cb3a200ba6e907b4f0ec9fdfd2f621ffd6f

Much better way to do it, yet another great rust features I didn't know about.

Issue #87, opened 2021-11-12
Add api to generate child keys from master key


notmandatory:
**Description**

Add an API to `bdk-ffi` to derive a child key pair from a master extended key and a derivation path string (eg. "m/84'/1'/0'/0" or "m/84h/1h/0h/0").  Also test the new `bdk-ffi` functions and structs via the `bdk-kotlin` and/or `bdk-swift` projects. See the `bdk-cli` key -> [derive](https://github.com/bitcoindevkit/bdk-cli/blob/b0c78e0aef181b2156c29549c3e27712166a4e49/src/lib.rs#L1122) subcommand as an example.

**Expected Outcomes**  
* Update `bdk.udl` and `lib.rs` with functions and structs as needed
* Update `bdk-kotlin` or `bdk-swift` and add tests to cover new key generation functionality 

**Resources**
* [BDK Website](https://bitcoindevkit.org)
* [BDK github repo](https://github.com/bitcoindevkit/bdk)
* [BDK Discord](https://discord.gg/dstn4dQ)
* [uniffi-rs](https://mozilla.github.io/uniffi-rs/)

**Skills Required**  
* Experience with git. [Guide](https://git-scm.com/book/en/v2)
* Basic familiarity with rust. [First seven chapters of the book](https://doc.rust-lang.org/book/)
* Basic familiarity with Kotlin or Swift
* Familiarity with bitcoin descriptors and address generation

**Mentor(s)**  

@artfuldev @notmandatory 

**Difficulty**  

Medium

**Competency Test (optional)**  
* Install rust, compile `bdk-ffi` and build and test `bdk-swift` or `bdk-kotlin`. 
* Read through the BDK docs.
* Install and run example `bdk-kotlin` or `bdk-swift` example wallet.
* Familiarity with basic rust, should be able to write basic structs and functions.

notmandatory:
@dhruv-1001 will be working on this issue this year's summer of bitcoin program.

kirillzh:
[...]

PR #154, opened 2022-05-15
Add child key pair generation api

dhruv-1001:
Would like to know if anything can be improved. Completes #87

### DIFF
--- a/src/bdk.udl
+++ b/src/bdk.udl
@@ -245,3 +245,8 @@ interface BumpFeeTxBuilder {
   [Throws=BdkError]
   PartiallySignedBitcoinTransaction finish([ByRef] Wallet wallet);
 };
+
+interface DerivationPath {
+  [Throws=BdkError]
+  constructor(string path);
+};
--- a/src/bdk.udl
+++ b/src/bdk.udl
@@ -1,10 +1,4 @@
-namespace bdk {
-  [Throws=BdkError]
-  ExtendedKeyInfo generate_extended_key(Network network, WordCount word_count, string? password);
-
-  [Throws=BdkError]
-  ExtendedKeyInfo restore_extended_key(Network network, string mnemonic, string? password);
-};
+namespace bdk {};
 
 [Error]
 enum BdkError {
@@ -99,12 +93,6 @@ dictionary BlockTime {
 interface Transaction {
   Unconfirmed(TransactionDetails details);
   Confirmed(TransactionDetails details, BlockTime confirmation);
-};
-
-dictionary ExtendedKeyInfo {
-  string mnemonic;
-  string xprv;
-  string fingerprint;
 };
 
 enum WordCount {
--- a/src/bdk.udl
+++ b/src/bdk.udl
@@ -250,3 +250,26 @@ interface DerivationPath {
   [Throws=BdkError]
   constructor(string path);
 };
+
+interface DescriptorSecretKey {
+  [Throws=BdkError]
+  constructor(Network network, string mnemonic, string? password);
+
+  [Throws=BdkError]
+  DescriptorSecretKey derive(DerivationPath path);
+
+  DescriptorSecretKey extend(DerivationPath path);
+
+  DescriptorPublicKey as_public();
+
+  string as_string();
+};
+
+interface DescriptorPublicKey {
+  [Throws=BdkError]
+  DescriptorPublicKey derive(DerivationPath path);
+
+  DescriptorPublicKey extend(DerivationPath path);
+
+  string as_string();
+};
--- a/src/bdk.udl
+++ b/src/bdk.udl
@@ -1,4 +1,7 @@
-namespace bdk {};
+namespace bdk {
+  [Throws=BdkError]
+  string generate_mnemonic(WordCount word_count);
+};
 
 [Error]
 enum BdkError {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -351,46 +351,6 @@ impl Wallet {
     }
 }
 
-pub struct ExtendedKeyInfo {
-    mnemonic: String,
-    xprv: String,
-    fingerprint: String,
-}
-
-fn generate_extended_key(
-    network: Network,
-    word_count: WordCount,
-    password: Option<String>,
-) -> Result<ExtendedKeyInfo, Error> {
-    let mnemonic: GeneratedKey<_, BareCtx> =
-        Mnemonic::generate((word_count, Language::English)).unwrap();
-    let mnemonic = mnemonic.into_key();
-    let xkey: ExtendedKey = (mnemonic.clone(), password).into_extended_key()?;
-    let xprv = xkey.into_xprv(network).unwrap();
-    let fingerprint = xprv.fingerprint(&Secp256k1::new());
-    Ok(ExtendedKeyInfo {
-        mnemonic: mnemonic.to_string(),
-        xprv: xprv.to_string(),
-        fingerprint: fingerprint.to_string(),
-    })
-}
-
-fn restore_extended_key(
-    network: Network,
-    mnemonic: String,
-    password: Option<String>,
-) -> Result<ExtendedKeyInfo, Error> {
-    let mnemonic = Mnemonic::parse_in(Language::English, mnemonic).unwrap();
-    let xkey: ExtendedKey = (mnemonic.clone(), password).into_extended_key()?;
-    let xprv = xkey.into_xprv(network).unwrap();
-    let fingerprint = xprv.fingerprint(&Secp256k1::new());
-    Ok(ExtendedKeyInfo {
-        mnemonic: mnemonic.to_string(),
-        xprv: xprv.to_string(),
-        fingerprint: fingerprint.to_string(),
-    })
-}
-
 fn to_script_pubkey(address: &str) -> Result<Script, BdkError> {
     Address::from_str(address)
         .map(|x| x.script_pubkey())
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -627,6 +627,12 @@ impl BumpFeeTxBuilder {
     }
 }
 
+fn generate_mnemonic(word_count: WordCount) -> Result<String, BdkError> {
+    let mnemonic: GeneratedKey<_, BareCtx> =
+        Mnemonic::generate((word_count, Language::English)).unwrap();
+    Ok(mnemonic.to_string())
+}
+
 uniffi::deps::static_assertions::assert_impl_all!(Wallet: Sync, Send);
 
 // The goal of these tests to to ensure `bdk-ffi` intermediate code correctly calls `bdk` APIs.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,7 @@
 use bdk::bitcoin::hashes::hex::ToHex;
 use bdk::bitcoin::secp256k1::Secp256k1;
 use bdk::bitcoin::util::psbt::PartiallySignedTransaction;
+use bdk::bitcoin::util::bip32::DerivationPath as BdkDerivationPath;
 use bdk::bitcoin::{Address, Network, OutPoint as BdkOutPoint, Script, Txid};
 use bdk::blockchain::any::{AnyBlockchain, AnyBlockchainConfig};
 use bdk::blockchain::{
@@ -631,6 +632,20 @@ fn generate_mnemonic(word_count: WordCount) -> Result<String, BdkError> {
     let mnemonic: GeneratedKey<_, BareCtx> =
         Mnemonic::generate((word_count, Language::English)).unwrap();
     Ok(mnemonic.to_string())
+}
+
+struct DerivationPath {
+    derivation_path_mutex: Mutex<BdkDerivationPath>,
+}
+
+impl DerivationPath {
+    fn new(path: String) -> Result<Self, BdkError> {
+        BdkDerivationPath::from_str(&path)
+            .map(|x| DerivationPath {
+                derivation_path_mutex: Mutex::new(x),
+            })
+            .map_err(|e| BdkError::Generic(e.to_string()))
+    }
 }
 
 uniffi::deps::static_assertions::assert_impl_all!(Wallet: Sync, Send);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -676,45 +676,49 @@ impl DescriptorSecretKey {
         let secp = Secp256k1::new();
         let descriptor_secret_key = self.descriptor_secret_key_mutex.lock().unwrap();
         let path = path.derivation_path_mutex.lock().unwrap().deref().clone();
-        let descriptor_x_key = match descriptor_secret_key.deref() {
-            BdkDescriptorSecretKey::XPrv(descriptor_x_key) => Some(descriptor_x_key),
-            _ => None,
+        match descriptor_secret_key.deref() {
+            BdkDescriptorSecretKey::XPrv(descriptor_x_key) => {
+                let derived_xprv = descriptor_x_key.xkey.derive_priv(&secp, &path)?;
+                let key_source = match descriptor_x_key.origin.clone() {
+                    Some((fingerprint, origin_path)) => (fingerprint, origin_path.extend(path)),
+                    None => (descriptor_x_key.xkey.fingerprint(&secp), path),
+                };
+                let derived_descriptor_secret_key = BdkDescriptorSecretKey::XPrv(DescriptorXKey {
+                    origin: Some(key_source),
+                    xkey: derived_xprv,
+                    derivation_path: BdkDerivationPath::default(),
+                    wildcard: descriptor_x_key.wildcard,
+                });
+                Ok(Arc::new(Self {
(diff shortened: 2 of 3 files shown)


dhruv-1001:
@notmandatory @artfuldev please review this draft.

src/lib.rs
```diff
@@ -313,6 +321,32 @@ fn restore_extended_key(
     })
 }
 
+fn derive_child_key_pair(
+    str_xprv: String,
+    str_path: String
+) -> Result<ChildKeyPair, Error> {
```
artfuldev, line 327: nitpick: `Error` can be `BdkError`

artfuldev:
hey I just had a few seconds, didn't do a complete review

src/bdk.udl
```diff
@@ -173,6 +175,11 @@ dictionary ExtendedKeyInfo {
   string fingerprint;
 };
 
+dictionary ChildKeyPair {
+  string xprv;
+  string xpub;
```
artfuldev, line 180: Do we not need to expose `ExtendedPrivKey` and `DerivationPath`? @notmandatory

src/bdk.udl
```diff
@@ -3,6 +3,8 @@ namespace bdk {
   ExtendedKeyInfo generate_extended_key(Network network, WordCount word_count, string? password);
   [Throws=BdkError]
   ExtendedKeyInfo restore_extended_key(Network network, string mnemonic, string? password);
+  [Throws=BdkError]
+  ChildKeyPair derive_child_key_pair(string str_xprv, string str_path);
```
artfuldev, line 7: Do we not need to expose `ExtendedPrivKey` and `DerivationPath`? @notmandatory

src/lib.rs
```diff
@@ -313,6 +321,32 @@ fn restore_extended_key(
...
+    let xprv = ExtendedPrivKey::from_str(&str_xprv)?;
+    let path = DerivationPath::from_str(&str_path)?;
+    let derived_xprv = &xprv.derive_priv(&secp, &path)?;
+    let origin: KeySource = (xprv.fingerprint(&secp), path);
+    let derived_xprv_desc_key: DescriptorKey<Segwitv0> = 
+        derived_xprv.into_descriptor_key(Some(origin), DerivationPath::default())?;
+    if let Secret(desc_seckey, _, _) = derived_xprv_desc_key {
+        let desc_pubkey = desc_seckey
+            .as_public(&secp)
+            .map_err(|e| Error::Generic(e.to_string()))?;
```
artfuldev, line 338: The `bdk-cli` code here uses `.unwrap()` - @notmandatory can you add some context?

src/bdk.udl
artfuldev: @notmandatory if this is the only way to create a child key pair, can it be a constructor?

artfuldev (commented):
Just have some questions, mostly for @notmandatory

src/bdk.udl
dhruv-1001: I modified the bdk-ff commit that the bdk-kotlin uses, and made same changes. After that, I added test for the deriveChildKeyPair() function 
```
@Test
    fun validChildKeys() {
        val str_xprv = "tprv8ZgxMBicQKsPdMTRB73kkTfC665bK3JL33eAFZ3MUiGZj8jniKQroG2yuamoQcRb24GG8dKRmVgaisjcZrDc9L3fyMf2NW19oajw8iWSwMV"
        val str_path = "m/84'/1'/0'/0"
        val childKeyPair = deriveChildKeyPair(str_xprv, str_path)
        assertEquals(
            "[59a3697d/84'/1'/0'/0]tprv8hZWYaWfKprzVnSVaaxwYun2yVp9TgpvcosjrGYwgL9hKMU1Jsrp3xKsNCpxSNBSEgUPp7c8A1MYLQcMcq7Ab9eafzNZUHQaBtGBvfCQiMg/*",
            childKeyPair.xprv
        )
        assertEquals(
            "[59a3697d/84'/1'/0'/0]tpubDEFYgzYuUCYfPFUHUEdXxKS9YXL5d21qC7UX8nbF6bx69qimwGgQESwjYN52awKwDd5PWGKXWAw27EBUk6QzSK3eEwsK8odY8FxVn4poRDp/*",
            childKeyPair.xpub
        )
    }
```

And It worked 

(I used the CLI to derive this result first, and used the same key for testing)

artfuldev: Not discounting that it works, just checking if the types are okay or if we need something more elaborate.  and if child key pair could have a constructor instead of this function `deriveChildKeyPair`. also wondering if `ChildKeyPair` is what we want to expose - it could be any key pair, not necessarily only a "child" key pair. just thinking out loud.

src/lib.rs
dhruv-1001: Got it, will change this.

src/bdk.udl
notmandatory: I think it's fine to keep the types and design for this simple to start and not add additional types for `ExtendedPrivKey` and `DerivationPath`. They can always be added later if we find a need, I think most apps will 1. `generate_extended_key` and get the user to write down the mnemonic, 2) from the `ExtendedKeyInfo.xprv` or possibly from importing a `xprv` from somewhere else,  derive a descriptor key, and 3) create a descriptor for the wallet, currently needs to be done putting strings together but in future we can make it simpler with some standard/common templates. That all said, I have two naming change suggestions:

1. rename `ChildKeyPair` to  `DescriptorKeyInfo` which is more in line with what `bdk` calls it
2. rename arguments `str_xprv` and `str_path` to simply `xprv` and `path` since type is already enforced thus redundant in the variable name.

As for creating the `ChildKeyPair` or `DescriptorKeyInfo` with a constructor, if we do that we need to change it from `dictionary` (pass by value) to an `interface` (pass by reference), so I'd prefer to keep it a `dictionary` and not use a constructor.

src/lib.rs
notmandatory: > The `bdk-cli` code here uses `.unwrap()` - @notmandatory can you add some context?

I think maybe you're looking at an older version of `bdk-cli` :-) this is the cli code works also without an `.unwrap()`.

src/bdk.udl
dhruv-1001: Noted. Will force push the changes soon.

dhruv-1001:
@artfuldev @notmandatory any suggestions here?

src/lib.rs
```diff
@@ -292,44 +295,89 @@ impl Wallet {
...
-    word_count: WordCount,
-    password: Option<String>,
-) -> Result<ExtendedKeyInfo, Error> {
-    let mnemonic: GeneratedKey<_, BareCtx> =
+impl ExtendedKeyInfo {
+    fn new(
+        network: Network,
+        word_count: WordCount,
+        password: Option<String>,
+    ) -> Self {
```
artfuldev, line 314: Can we retain `Result<T, E>` semantics here? Did we remove it intentionally? I'd prefer not having `.unwrap()` and `throw`ing exceptions in Kotlin etc.

dhruv-1001: Got it, will add back the `Result<T, E>` and force push the changes.

src/lib.rs
```diff
@@ -292,44 +295,89 @@ impl Wallet {
     }
 }
 
-pub struct ExtendedKeyInfo {
+pub struct DescriptorKeyInfo {
+    xprv: Option<String>,
+    xpub: String,
+}
```
artfuldev, line 301: welcome this type! @notmandatory do you have any take on the name?

notmandatory: Now that I'm looking at it all together and the naming I'm thinking my original suggestion is not quite right. The use cases I'm trying to solve for are:

    Generate new mnemonic seed words, or create from existing seed words
    Create new extended private key for a mnemonic and network, with optional password
    Derive new public or private extended key from a private extended key and derivation path
    (future) Create a wallet descriptor from a standard template given a public or private extended key

Based on this I think the types we need are Mnemonic, ExtendedPrivKey, ExtendedPubKey.

src/lib.rs
```diff
@@ -2,6 +2,11 @@ use bdk::bitcoin::hashes::hex::ToHex;
 use bdk::bitcoin::secp256k1::Secp256k1;
 use bdk::bitcoin::util::psbt::PartiallySignedTransaction;
 use bdk::bitcoin::{Address, Network, Script, Txid};
+use bdk::bitcoin::util::bip32::{
+    DerivationPath as BdkDerivationPath,
+    ExtendedPrivKey as BdkExtendedPrivKey,
+    ExtendedPubKey as BdkExtendedPrivKey,
```
notmandatory, line 8: Looks like a typo here:
```suggestion
    ExtendedPubKey as BdkExtendedPubKey,
```

src/bdk.udl
```diff
@@ -1,9 +1,23 @@
...
   [Throws=BdkError]
-  ExtendedKeyInfo restore_extended_key(Network network, string mnemonic, string? password);
+  constructor(Network network, string mnemonic, string? password);
+  ExtendedPrivKey derive_xprv(DerivationPath? derivation_path);
+  ExtendedPubKey derive_xpub(DerivationPath? derivation_path);
+  string to_string();
+};
+
+interface ExtendedPubKey {
+  string to_string();
```
kirillzh, line 20: Looks like this won't compile in Kotlin since `toString()` is an open, member function of `Any` that needs to be overriden and uniffi doesn't seem to have a way to declare an overriding API, as far as I can tell. A simple workaround here would be to rename `to_string()` to something else (maybe `ExtendedPubKey#value` or `#raw` read-only property?)

kirillzh: Filed an issue with uniffi to see if it's feasible to allow adding `override` modifier in Kotlin: https://github.com/mozilla/uniffi-rs/issues/1286.

src/lib.rs
```diff
@@ -285,44 +289,85 @@ impl Wallet {
...
-}
-
-fn generate_extended_key(
-    network: Network,
-    word_count: WordCount,
-    password: Option<String>,
-) -> Result<ExtendedKeyInfo, Error> {
+fn generate_mnemonic(word_count: WordCount) -> Result<String, BdkError> {
     let mnemonic: GeneratedKey<_, BareCtx> =
         Mnemonic::generate((word_count, Language::English)).unwrap();
```
notmandatory, line 294: Rather than doing an `unwrap()` you should map the `Error` to a `BdkError`.

src/lib.rs
```diff
@@ -285,44 +289,85 @@ impl Wallet {
...
+    Ok(mnemonic.to_string())
+}
+
+struct DerivationPath {
+    derivation_path: Mutex<BdkDerivationPath>,
+}
+
+impl DerivationPath {
+    fn new(path: String) -> Result<Self, BdkError> {
+        let path = BdkDerivationPath::from_str(&path).unwrap();
```
notmandatory, line 304: Same here, should map the `Error` to a `BdkError`.

notmandatory:
Overall looks good, I commented on a few places you can remove the `.unwrap()` and use `map_error` and `map` to generate your `Result`. There are a few others I didn't comment on, but in general try to avoid `.unwrap()` if you can, except in tests where it's OK.

thunderbiscuit (commented):
It's great to have tests in there because it allowed me to visualize (at least partially) what the API would look like.

Overall this is a big/complex API to add, so I really think we need to make sure it's done right. I'm still unclear on a few things; see comments below.

src/lib.rs
```diff
@@ -285,44 +290,119 @@ impl Wallet {
...
+            .into_descriptor_key(None, BdkDerivationPath::master())?;
+        Ok(Self {
+            descriptor_key_mutex: Mutex::new(descriptor_key),
+        })
+    }
+
+    fn derive(
+        &self,
+        derivation_path: Arc<DerivationPath>,
+        descriptor_path: Option<Arc<DerivationPath>>,
```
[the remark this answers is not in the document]
thunderbiscuit: How about the BDK names `origin` and `derivation_path`? https://docs.rs/bdk/latest/bdk/keys/enum.ExtendedKey.html#method.into_descriptor_key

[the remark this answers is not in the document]
thunderbiscuit: Here's a potential function signature:
```rust
fn derive(
    &self,
    origin_path: Option<Arc<DerivationPath>>,
    descriptor_path: Option<Arc<DerivationPath>>,
) -> Result<Arc<DescriptorKey>, BdkError>

File: src/lib.rs [hunk 12]
```diff
@@ -584,4 +664,64 @@ mod test {
...
+    ) -> Result<Arc<DescriptorKey>, BdkError> {
+        let origin_path = Arc::new(DerivationPath::new(derivation_path).unwrap());
+        let derivation_path = descriptor_path.map(|dp| Arc::new(DerivationPath::new(dp).unwrap()));
+        descriptor_key.derive(origin_path, derivation_path)
+    }
+
+    #[test]
+    fn test_generate_descriptor_key() {
+        let descriptor_key = get_descriptor_key();
+        assert_eq!(descriptor_key.as_string(), "tprv8ZgxMBicQKsPdWuqM1t1CDRvQtQuBPyfL6GbhQwtxDKgUAVPbxmj71pRA8raTqLrec5LyTs5TqCxdABcZr77bt2KyWA5bizJHnC4g4ysm4h/*");
```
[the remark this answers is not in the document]
thunderbiscuit: I was confusing the `ExtendedKey` type with the `DescriptorKey` type. Nevermind!

thunderbiscuit:
@dhruv-1001 this PR will need a small rebase of the changelog.

notmandatory (approved):
ACK 590ab377ad162097b7927ba85bff774f8217f61a

This one turned out to be a real journey, thanks for sticking with it!

thunderbiscuit (approved):

notmandatory:
I had a little chat with @thunderbiscuit and @dhruv-1001 and it looks like the logic in this PR probably does belong in `bdk`. I'm going to take a stab at adding `DescriptorKey::derive()` and `DescriptorKey::extend()` functions in `bdk`. Then we can review and discuss there and expose what ever API we come up with here to `bdk-ffi`. How does that sound @quad ?

notmandatory:
@quad I've created a draft`bdk` PR to explore what I had in mind for an API. The main diff I think with what you have is that I'm doing everything at the `DescriptorKey` level. Let me know what you and your team think and/or we can jump on a call to discuss.

https://github.com/bitcoindevkit/bdk/pull/670

src/lib.rs
```diff
@@ -667,14 +632,172 @@ impl BumpFeeTxBuilder {
...
+struct DescriptorPublicKey {
+    descriptor_public_key_mutex: Mutex<BdkDescriptorPublicKey>,
+}
+
+impl DescriptorPublicKey {
+    fn derive(&self, path: Arc<DerivationPath>) -> Result<Arc<Self>, BdkError> {
+        let secp = Secp256k1::new();
+        let descriptor_public_key = self.descriptor_public_key_mutex.lock().unwrap();
+        let path = path.derivation_path_mutex.lock().unwrap().deref().clone();
+        let descriptor_x_key = match descriptor_public_key.deref() {
```
notmandatory, line 747: I think this match code and other's like it in your code would be clearer like this:
```rust
match descriptor_secret_key.deref() {
      BdkDescriptorSecretKey::XPrv(descriptor_x_key) => {
          let derived_xprv = descriptor_x_key.xkey.derive_priv(&secp, &path)?;
          let key_source = match descriptor_x_key.origin.clone() {
              Some((fingerprint, origin_path)) => (fingerprint, origin_path.extend(path)),
              None => (descriptor_x_key.xkey.fingerprint(&secp), path),
          };
          let derived_descriptor_secret_key = BdkDescriptorSecretKey::XPrv(DescriptorXKey {
              origin: Some(key_source),
              xkey: derived_xprv,
              derivation_path: BdkDerivationPath::default(),
              wildcard: descriptor_x_key.wildcard,
          });
          Ok(Arc::new(Self {
              descriptor_secret_key_mutex: Mutex::new(derived_descriptor_secret_key),
          }))
      },
      BdkDescriptorSecretKey::SinglePriv(_) => {
          unreachable!()
      }
  }
```
This way it's clear that since we're not supporting single keys the alternative branch is unreachable (which will still cause a panic if for some reason it is reached). 

At some point we should support single keys.. and then in the derive/extend functions we'll need to throw a run-time Err, which is less than ideal. I'd rather avoid that, but with no support for generics in the APIs we can build with uniffi it may be the least ugly approach. But not a concern for this PR.

notmandatory (changes requested):
Overall looks good! I'd just like the match statements cleaned up to avoid returning None and then unwrap(), which I don't think is as clear.

notmandatory (approved):
ACK 5bc381111bb78d7a3efd9c0a3158385d8cd03441

Good work! this one was a challenging API to simplify for FFI and I think this way strikes the right balance.

notmandatory:
Note to anyone following this issue, as discussed in https://github.com/bitcoindevkit/bdk/pull/670 we decided not to put this logic in the `bdk` lib, but once this PR is in I'll add some code example in the `bdk` repo, it should be only a couple lines to generate, derive or extend a `DescriptorPublicKey` or `DescriptorSecretKey`.

PR #158, opened 2022-06-09
Add Wallet list_unspent method

zoedberg:
Here is my attempt at implementing the `list_unspent` method.

In order to test this I've updated the `bdk-ffi` submodule in the `bdk-kotlin` project, ran commands to build and publish the android library to a local maven repository and finally imported the library on an Android project written in Kotlin.

The list seems to work fine, except for a strange behaviour I've noticed when calling the method multiple times.

So, in order to exclude bugs from my side, I've wrote a minimal Rust project that directly uses bdk 0.18 to list the unspents of an existing wallet.

From this test I've understood that creating a wallet using a `bdk::database::SqliteDatabase` database and calling the wallet `sync` method leads to each UTXO getting re-inserted in the database even if already present.

Here's a minimal piece of code that shows how to reproduce the bug:

```rust
let blockchain = ElectrumBlockchain::from(Client::new("ssl://electrum.blockstream.info:60002")?);
let database = bdk::database::SqliteDatabase::new("./testdb.sqlite".to_string());
let wallet = Wallet::new(
    "wpkh(tprv8ZgxMBicQKsPdqrqwQmJsdw7Sbt36hJPJyoCQ5BhFwfLcfVHQEwiwPykQPLT9AxUYS26BPQ4yrqTYYGG4HPg4UwJzLJCnK5L8YC6VAUb8nZ/84'/1'/0'/0/*)",
    Some("wpkh(tprv8ZgxMBicQKsPdqrqwQmJsdw7Sbt36hJPJyoCQ5BhFwfLcfVHQEwiwPykQPLT9AxUYS26BPQ4yrqTYYGG4HPg4UwJzLJCnK5L8YC6VAUb8nZ/84'/1'/0'/1/*)"),
    Network::Testnet,
    database,
)?;

wallet.sync(&blockchain, SyncOptions::default())?;

println!("Num unspents: {}", wallet.list_unspent()?.len());

>> Num unspents: 1

wallet.sync(&blockchain, SyncOptions::default())?;

println!("Num unspents: {}", wallet.list_unspent()?.len());

>> Num unspents: 2

println!("Num unspents: {}", wallet.list_unspent()?.len());

>> Num unspents: 2

wallet.sync(&blockchain, SyncOptions::default())?;

println!("Num unspents: {}", wallet.list_unspent()?.len());

>> Num unspents: 3

println!("Num unspents: {}", wallet.list_unspent()?.len());

>> Num unspents: 3
```

Inspecting the sqlite database shows that the UTXO actually gets re-inserted at each sync.

The error does not arise when the database is defined as `MemoryDatabase::default()`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -190,7 +190,8 @@ pub struct LocalUtxo {
     is_spent: bool,
 }
 
-/// Trait used to convert a script to an address using the wallet network
+// This trait is used to convert the bdk TxOut type with field `script_pubkey: Script`
+// into the bdk-ffi TxOut type which has a field `address: String` instead
 trait NetworkLocalUtxo {
     fn from_utxo(x: &bdk::LocalUtxo, network: Network) -> LocalUtxo;
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,7 +15,8 @@ use bdk::miniscript::BareCtx;
 use bdk::wallet::AddressIndex as BdkAddressIndex;
 use bdk::wallet::AddressInfo as BdkAddressInfo;
 use bdk::{
-    BlockTime, Error, FeeRate, SignOptions, SyncOptions as BdkSyncOptions, Wallet as BdkWallet,
+    BlockTime, Error, FeeRate, KeychainKind, SignOptions, SyncOptions as BdkSyncOptions,
+    Wallet as BdkWallet,
 };
 use std::convert::{From, TryFrom};
 use std::fmt;
@@ -182,27 +183,15 @@ pub struct TxOut {
     address: String,
 }
 
-pub enum KeychainKind {
-    External = 0,
-    Internal = 1,
-}
-
-impl From<bdk::KeychainKind> for KeychainKind {
-    fn from(x: bdk::KeychainKind) -> KeychainKind {
-        match x {
-            bdk::KeychainKind::External => KeychainKind::External,
-            bdk::KeychainKind::Internal => KeychainKind::Internal,
-        }
-    }
-}
-
 pub struct LocalUtxo {
     outpoint: OutPoint,
     txout: TxOut,
     keychain: KeychainKind,
+    is_spent: bool,
 }
 
-trait NetworkLocalUtxo: {
+/// Trait used to convert a script to an address using the wallet network
+trait NetworkLocalUtxo {
     fn from_utxo(x: &bdk::LocalUtxo, network: Network) -> LocalUtxo;
 }
 
@@ -216,9 +205,14 @@ impl NetworkLocalUtxo for LocalUtxo {
             txout: TxOut {
                 value: x.txout.value,
                 address: bdk::bitcoin::util::address::Address::from_script(
-                    &x.txout.script_pubkey, network).unwrap().to_string(),
+                    &x.txout.script_pubkey,
+                    network,
+                )
+                .unwrap()
+                .to_string(),
             },
-            keychain: KeychainKind::from(x.keychain),
+            keychain: x.keychain,
+            is_spent: x.is_spent,
         }
     }
 }
@@ -337,7 +331,10 @@ impl Wallet {
 
     fn list_unspent(&self) -> Result<Vec<LocalUtxo>, Error> {
         let unspents = self.get_wallet().list_unspent()?;
-        Ok(unspents.iter().map(|u| LocalUtxo::from_utxo(u, self.get_network())).collect())
+        Ok(unspents
+            .iter()
+            .map(|u| LocalUtxo::from_utxo(u, self.get_network()))
+            .collect())
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -172,6 +172,57 @@ struct Wallet {
     wallet_mutex: Mutex<BdkWallet<AnyDatabase>>,
 }
 
+pub struct OutPoint {
+    txid: String,
+    vout: u32,
+}
+
+pub struct TxOut {
+    value: u64,
+    address: String,
+}
+
+pub enum KeychainKind {
+    External = 0,
+    Internal = 1,
+}
+
+impl From<bdk::KeychainKind> for KeychainKind {
+    fn from(x: bdk::KeychainKind) -> KeychainKind {
+        match x {
+            bdk::KeychainKind::External => KeychainKind::External,
+            bdk::KeychainKind::Internal => KeychainKind::Internal,
+        }
+    }
+}
+
+pub struct LocalUtxo {
+    outpoint: OutPoint,
+    txout: TxOut,
+    keychain: KeychainKind,
+}
+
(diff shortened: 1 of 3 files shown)


notmandatory:
This looks like great stuff. As for the next `bdk-ffi` release the focus will be on upgrading to the `bdk` `0.19.0` release. But once that's out we can get back to adding new APIs like this one.

src/lib.rs
```diff
@@ -172,6 +172,57 @@ struct Wallet {
...
+impl From<bdk::KeychainKind> for KeychainKind {
+    fn from(x: bdk::KeychainKind) -> KeychainKind {
+        match x {
+            bdk::KeychainKind::External => KeychainKind::External,
+            bdk::KeychainKind::Internal => KeychainKind::Internal,
+        }
+    }
+}
+
+pub struct LocalUtxo {
```
thunderbiscuit, line 199: The BDK version of this struct has a boolean `is_spent`. I assume you didn't include it here because `LocalUtxo` will only be returned from the `list_unspent()` method, which means that boolean will only ever been set to `false`. In this case, however, I think the naming could lead to confusion if some developers are familiar with the Rust version of BDK.

I propose we either:
1. Rename the struct `UnspentLocalUtxo` to make explicit the fact that it is unspent, or
2. Add the `is_spent` boolean to it, which for now will only ever be false

I suggest we go with (2) because to me that feels more forward-compatible. if ever we implement other features that use the `LocalUtxo` struct we'll have it in its entirety already. I also feel like it might allow you to simplify some of the mapping between the `LocalUtxo` struct you have here and the one defined in BDK.

src/lib.rs
```diff
@@ -172,6 +172,57 @@ struct Wallet {
...
+    txid: String,
+    vout: u32,
+}
+
+pub struct TxOut {
+    value: u64,
+    address: String,
+}
+
+pub enum KeychainKind {
```
thunderbiscuit, line 185: I'm not sure we need the 0 and 1 here but maybe there is something I'm missing. Can you tell me why you added them and not just use the enum defined in BDK directly? (i.e. no need to define it here in Rust and you could simply have it in the UDL file I think)

thunderbiscuit: Follow-up: my [tests pass](https://github.com/thunderbiscuit/bdk-kotlin-test-suite/blob/23527ef3797eff48afc81b11af0fd7c776c84734/app/src/test/kotlin/bdk/kotlin/testsuite/ListUnspent.kt#L35-L46) when refactoring the KeychainKind out of the lib.rs file like so:
```rust
// pub enum KeychainKind {
//     External = 0,
//     Internal = 1,
// }

// impl From<bdk::KeychainKind> for KeychainKind {
//     fn from(x: bdk::KeychainKind) -> KeychainKind {
//         match x {
//             bdk::KeychainKind::External => KeychainKind::External,
//             bdk::KeychainKind::Internal => KeychainKind::Internal,
//         }
//     }
// }

impl NetworkLocalUtxo for LocalUtxo {
    fn from_utxo(x: &bdk::LocalUtxo, network: Network) -> LocalUtxo {
        LocalUtxo {
            outpoint: OutPoint {
                txid: x.outpoint.txid.to_string(),
                vout: x.outpoint.vout,
            },
            txout: TxOut {
                value: x.txout.value,
                address: bdk::bitcoin::util::address::Address::from_script(
                    &x.txout.script_pubkey, network).unwrap().to_string(),
            },
            keychain: x.keychain,
        }
    }
}
```

src/lib.rs
```diff
@@ -172,6 +172,57 @@ struct Wallet {
...
 }
 
+pub struct OutPoint {
+    txid: String,
+    vout: u32,
+}
+
+pub struct TxOut {
+    value: u64,
+    address: String,
```
thunderbiscuit, line 182: At first glance it looks good to me, but it's still something I want to bring to the attention of other reviewers (@notmandatory, @artfuldev). The `TxOut` in BDK contains a `Script` instead of an address of type `String`. I don't think we need the `Script` at the moment, but it does change the shape of the API, so something to note.

src/lib.rs
```diff
@@ -172,6 +172,57 @@ struct Wallet {
...
+    txout: TxOut,
+    keychain: KeychainKind,
+}
+
+trait NetworkLocalUtxo: {
+    fn from_utxo(x: &bdk::LocalUtxo, network: Network) -> LocalUtxo;
+}
+
+impl NetworkLocalUtxo for LocalUtxo {
+    fn from_utxo(x: &bdk::LocalUtxo, network: Network) -> LocalUtxo {
```
thunderbiscuit, line 210: I'd love it if we started commenting our codebase more (the FFI layer can be quite opaque and I want to make sure we make it easy on new contributors), and I think this is a perfect place to start. 

It took me a few minutes to understand what this `NetworkLocalUtxo` trait was for, and if I understand correctly, you added this trait because the `TxOut` struct now contains an `address` (`String`), and for that you need the [Address::from_script()](https://docs.rs/bitcoin/latest/bitcoin/util/address/struct.Address.html#method.from_script) which requires a network. That's great stuff and let me know if I didn't understand correctly. Long-winded way to say... I think we should just add a small comment explaining what this `NetworkLocalUtxo` is used for 🤣.

thunderbiscuit (commented):
Thanks a bunch for tackling this @zoedberg! Coin selection is an important feature and this is the first step.

I left a few comments there for you; let me know if they all make sense.

zoedberg:
@thunderbiscuit A big thanks to you and everybody who participated to this project, it's a pleasure to contribute to such a well designed and written project. I agree with all the comments, working on in it right now, will push the changes soon :)

src/lib.rs
zoedberg: Sorry, my mistake. I initially assumed that was a requirement of uniFFI but then forgot to try without. Thanks for the suggestion!

src/lib.rs
zoedberg: I also feel that (2) is a better choice here so just added the field

src/lib.rs
zoedberg: I had some hard times finding a fitting and self-explanatory name for this trait, so I agree a comment could be useful here. For next contributions I will keep in mind that comments are welcome :smile:

zoedberg:
Rebased the PR branch and added the suggested changes. Tested them on Android where they seem to work fine :)

thunderbiscuit:
This is ready for merge IMO. I'll wait for one more reviewer to chime in and will squash your commits into one + merge.

thunderbiscuit:
@notmandatory would love your review on this. I'm not sure if we should merge before #154, as this is an easier and ready to go change, 154 might be still WIP a bit (but maybe not? happy to do it the other way too). In any case merging this will allow us to start reviewing #164, which I'd love to start working on this week.

notmandatory (approved):
ACK fad5e3cefdfec2e396928b8621e96407570b3fc4

Issue #160, opened 2022-06-10
PSBT signing is returning an error when transaction is not finalized


thunderbiscuit:
This creates problems for anything that requires more than one signature, as the ffi layer is returning an error when in fact the psbt _is_ properly signed, it simply isn't finalized.

The function signature in bdk is
```rust
pub fn sign(
    &self,
    psbt: &mut Psbt,
    sign_options: SignOptions
) -> Result<bool, Error>
```
The Boolean is `true` if the transaction is finalized and `false` if it isn't. Note that false doesn't imply the signature didn't work, simply that there are more inputs left to be signed. Our current bdk-ffi `sign` function, however, returns an error if the result is `false`:
```rust
fn sign(&self, psbt: &PartiallySignedBitcoinTransaction) -> Result<(), Error> {
    let mut psbt = psbt.internal.lock().unwrap();
    let finalized = self.get_wallet().sign(&mut psbt, SignOptions::default())?;
    match finalized {
        true => Ok(()),
        false => Err(BdkError::Generic(format!(
            "transaction signing not finalized {:?}",
            psbt
        ))),
    }
}
```

artfuldev:
We don't have multisig support yet. I'd rewrite this issue to just request multisig support.

thunderbiscuit:
No it does support it (in a weird way maybe but still) I think. Because the psbt gets modified, even after the `sign()` method returns an error (say it isn't finalized but it was still signable), your psbt is further along (has at least 1 more signature on it) and you could then pass it to a different signer that would be able to apply their signature to it.

Unless you mean in the sense of "because we're not really returning the proper result, it's not really fit for multisig at the moment and is more or less a hack", in which case yes I agree multisig is not really supported (but that's just because we're not returning the correct `Result<bool, Error>` which BDK returns.

notmandatory:
This isn't really about multisig support but about being able to make a partial signature for a multisig descriptor, but anyway I think it's the right way to go to follow the rust BDK api here.

artfuldev:
I meant, when we wrote the PSBT API we were just trying to get a transaction to be broadcast as a proof of concept and multisig really want on our radar. Hence the shape of the present APIs.

notmandatory:
fixed by #161

PR #161, opened 2022-06-10
Match bdk API and return a boolean when signing a PSBT

thunderbiscuit:
This is a fix for #160.

I was looking at the `get_transactions()` method just below and I'm not sure which syntax is best (let me know if you have opinions on this) between

What I have:
```rust
fn sign(&self, psbt: &PartiallySignedBitcoinTransaction) -> Result<bool, Error> {
    let mut psbt = psbt.internal.lock().unwrap();
    self.get_wallet().sign(&mut psbt, SignOptions::default())
}
```

If I mirrored `get_transactions()`:
```rust
fn sign(&self, psbt: &PartiallySignedBitcoinTransaction) -> Result<bool, Error> {
    let mut psbt = psbt.internal.lock().unwrap();
    let finalized = self.get_wallet().sign(&mut psbt, SignOptions::default())?;
    Ok(finalized)
}
```

### DIFF
--- a/src/bdk.udl
+++ b/src/bdk.udl
@@ -160,7 +160,7 @@ interface Wallet {
   u64 get_balance();
 
   [Throws=BdkError]
-  void sign([ByRef] PartiallySignedBitcoinTransaction psbt);
+  boolean sign([ByRef] PartiallySignedBitcoinTransaction psbt);
 
   [Throws=BdkError]
   sequence<Transaction> get_transactions();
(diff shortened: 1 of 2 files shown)


thunderbiscuit:
I wrote tests for this PR on [this branch of my bdk-kotlin repo](https://github.com/thunderbiscuit/bdk-kotlin/blob/tests/fix-bdkffi-160/jvm/src/test/kotlin/org/bitcoindevkit/MultisigTest.kt). Everything seems to be working as expected; the `sign()` method returns true when Alice signs a 1-of-2 multisig and `false` when she's the first signer of a 2-of-2 multisig.

notmandatory (approved):
ACK d52f1e3f5786580b9372c4bed5f28a5acd894342

notmandatory (approved):
reACK 9a3d6098261a73e8005e78a7ea2791734321af93

PR #163, opened 2022-06-10
Add add_data function for TxBuilder op_return

### LINKED ISSUE #132: Add TxBuilder addData method
Add method to TxBuilder to allow users to add data via an OpReturn.

See https://docs.rs/bdk/latest/bdk/wallet/tx_builder/struct.TxBuilder.html for docs on what the addData function does. 

```udl
interface TxBuilder {
  ... 
  TxBuilder add_data(sequence<u8>); 
}

```

waterst0ne:
fixes #132

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -344,6 +344,7 @@ struct TxBuilder {
     drain_wallet: bool,
     drain_to: Option<String>,
     rbf: Option<RbfValue>,
+    data: Vec<u8>,
 }
 
 impl TxBuilder {
@@ -354,6 +355,7 @@ impl TxBuilder {
             drain_wallet: false,
             drain_to: None,
             rbf: None,
+            data: Vec::new(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


thunderbiscuit:
I tested this and [broadcast a transaction out with a message](https://mempool.space/testnet/tx/4f3b4711de738bfc76bfdc8d4eb7702e91709c08af55c1060cdd74d2668688c9) 🔥 🔥 🔥.

src/lib.rs
```diff
@@ -354,9 +355,15 @@ impl TxBuilder {
...
+            data: Vec::new(),
         }
     }
-
+    fn add_data(&self, data: Vec<u8>) -> Arc<Self> {
+        Arc::new(TxBuilder {
+            data,
+            ..self.clone()
+        })
+    }
```
thunderbiscuit, line 366: All other methods on the `TxBuilder` are in the order the struct is defined. That would put `add_data()` at the end, just before `finish()`. Unless we have an explicit other ordering system (I don't think we do in this case if I'm not mistaken), let's keep them in that order (note that it will also mirror the order in the UDL file).

thunderbiscuit (changes requested):
Very small nit, otherwise I think this is ready for merge.

Tested ACK [410eb82](https://github.com/bitcoindevkit/bdk-ffi/pull/163/commits/410eb82587d34e6a431283b4c9c502841d3b35ce).

src/lib.rs
waterst0ne: Acknowledged, I have made the requested change. For next time, I will make a note of not changing the sequence of the code less it requires explicit ordering system. Thank you for reviewing and testing. 🚀

notmandatory (approved):
Looks good! ACK 548217e1b062391c6052ffe521a4425f6254ba0c

thunderbiscuit (approved):

PR #171, opened 2022-07-15
Add documentation

thunderbiscuit:
I think it's high time we document the Rust code.

Because we don't have official API docs at the moment (and we don't really know how long it will be before we can have them), I expect users will be referring to source code for documentation. Because we can't build those in the bindings files themselves at the moment, documenting the Rust code in bdk-ffi is a low hanging fruit that is probably worthy of our time.

This is currently in draft mode, but more or less I propose to try and pull from the official BDK Rust docs whenever the types are the same or the documentation applies, and write our own when needed. Here are examples for `AddressInfo` and `AddressIndex`:

```rust
/// A derived address and the index it was found at.
pub struct AddressInfo {
    /// Child index of this address
    pub index: u32,
    /// Address
    pub address: String,
}

/// The address index selection strategy to use to derived an address from the wallet's external
/// descriptor.
pub enum AddressIndex {
    /// Return a new address after incrementing the current descriptor index.
    New,
    /// Return the address for the current descriptor index if it has not been used in a received
    /// transaction. Otherwise return a new address as with AddressIndex::New.
    ///
    /// Use with caution, if the wallet has not yet detected an address has been used it could
    /// return an already used address. This function is primarily meant for situations where the
    /// caller is untrusted; for example when deriving donation addresses on-demand for a public
    /// web page.
    LastUnused,
}
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -139,19 +139,19 @@ pub enum BlockchainConfig {
 /// A wallet transaction
 #[derive(Debug, Clone, PartialEq, Eq, Default)]
 pub struct TransactionDetails {
-    /// Fee value (sats) if available.
-    /// The availability of the fee depends on the backend. It's never None with an Electrum
-    /// Server backend, but it could be None with a Bitcoin RPC node without txindex that receive
-    /// funds while offline.
-    pub fee: Option<u64>,
+    /// Transaction id.
+    pub txid: String,
     /// Received value (sats)
     /// Sum of owned outputs of this transaction.
     pub received: u64,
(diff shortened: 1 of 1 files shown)


notmandatory:
Looks good ! I think you need to rebase to pickup recently merged #178 .

thunderbiscuit:
Yeah I rebased earlier today. I think it's ready to go.

notmandatory (approved):
ACK cd38001952bbc486ce2ac496a88038cac188ebe1

PR #184, opened 2022-08-17
Add `get_height` and `get_block_hash` methods on blockchain

### LINKED ISSUE #151: Add functions to get current block height and hash
Once getting a block hash is supported in `bdk` then should add ability to get current block height and block hash at a given height to the `Blockchain` interface. These functions are needed to help support wallets that also use LDK, See https://github.com/bitcoindevkit/bdk-swift/issues/10.

Depends on:
  - [x] https://github.com/bitcoindevkit/bdk/issues/603

thunderbiscuit:
Fixes #151.

### Description
This PR adds the `get_height()` and `get_block_hash()` methods to the `blockchain`.

#### All Submissions:

#### New Features:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Add `get_height()` and `get_block_hash()` methods on blockchain [#184]
+
+[#184]: https://github.com/bitcoindevkit/bdk-ffi/pull/184
 
 ## [v0.8.0]
 - Update BDK to version 0.20.0 [#169]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `derive(DerivationPath)` derives and returns child DescriptorPublicKey
     - `extend(DerivationPath)` extends and returns DescriptorPublicKey
     - `as_string()` returns DescriptorPublicKey as String
-    - Add `get_height()` and `get_block_hash()` methods on blockchain [#184]
+  - Add to `interface Blockchain` the `get_height()` and `get_block_hash()` methods [#184]
 - Interfaces Added [#154]
   - `DescriptorSecretKey`
   - `DescriptorPublicKey`
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,6 +3,8 @@ use bdk::bitcoin::secp256k1::Secp256k1;
 use bdk::bitcoin::util::psbt::PartiallySignedTransaction;
 use bdk::bitcoin::{Address, Network, OutPoint as BdkOutPoint, Script, Txid};
 use bdk::blockchain::any::{AnyBlockchain, AnyBlockchainConfig};
+use bdk::blockchain::GetBlockHash;
+use bdk::blockchain::GetHeight;
 use bdk::blockchain::{
     electrum::ElectrumBlockchainConfig, esplora::EsploraBlockchainConfig, ConfigurableBlockchain,
 };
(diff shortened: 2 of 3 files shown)


thunderbiscuit:
I tested this against by [bdk-kotlin-test-suite](https://github.com/thunderbiscuit/bdk-kotlin-test-suite) and everything worked well on Electrum.
block-info

src/lib.rs
```diff
@@ -169,6 +171,18 @@ impl Blockchain {
         let tx = psbt.internal.lock().unwrap().clone().extract_tx();
         self.get_blockchain().broadcast(&tx)
     }
+
+    fn get_height(&self) -> Result<u32, Error> {
+        return self.get_blockchain().get_height();
```
notmandatory, line 176: To return the result I think you need to remove the trailing `;`, otherwise the returned type will be the empty type `()`. You should create a little test to confirm a new Blockchain returns `Ok(0)` for this function to confirm.
```suggestion
        return self.get_blockchain().get_height()
```

src/lib.rs
```diff
@@ -169,6 +171,18 @@ impl Blockchain {
...
+    }
+
+    fn get_block_hash(&self, height: u32) -> Result<String, Error> {
+        let block_hash = self
```
notmandatory, line 180: This is a good place to use `.map()` and also you won't need the `.unwrap()`. It should look something like this:
```suggestion
             self.get_blockchain()
            .get_block_hash(u64::from(height))
            .map(|hash| hash.to_string())
```

notmandatory (changes requested):
Looks good! I just have a couple rust stylistic suggestions.

src/lib.rs
notmandatory: I may be wrong about needing to remove the `;` since this seems to compile fine, if it wasn't returning a Result the CI should be throwing an error.

notmandatory:
Also looks like you'll need to rebase your branch to the latest `master`.

src/lib.rs
thunderbiscuit: My understanding is that the last line is an expression if it doesn't have the semicolon (and is automatically the return value of a function). Moreover I think that's the more idiomatic way to write Rust, but for some reason I had problems with it when I wrote it earlier, yet now it compiles fine. So I'll go with that!

```rust
    fn get_height(&self) -> Result<u32, Error> {
        self.get_blockchain().get_height()
    }
````

File: src/lib.rs (hunk 2 above)
Comment (thunderbiscuit) [reply src/lib.rs]: Yep sounds good!

**Comment by thunderbiscuit** [comment]
Clippy fails the CI build if with a "dead code" error...

```
 error: associated function is never used: `get_height`
   --> src/lib.rs:175:8
    |
175 |     fn get_height(&self) -> Result<u32, Error> {
    |        ^^^^^^^^^^
    |
    = note: `-D dead-code` implied by `-D warnings`

error: associated function is never used: `get_block_hash`
   --> src/lib.rs:179:8
    |
179 |     fn get_block_hash(&self, height: u32) -> Result<String, Error> {
    |        ^^^^^^^^^^^^^^

error: could not compile `bdk-ffi` due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
error: build failed
Error: Process completed with exit code 101.
```

**Comment by notmandatory** [comment]
Your clippy error is only an unfriendly reminder that you didn't add your new functions to the `bdk.udl` file! 🤣 

```udl
interface Blockchain {
...

  [Throws=BdkError]
  u32 get_height();

  [Throws=BdkError]
  string get_block_hash(u32 height);
};
```

thunderbiscuit:
Oh shoot! I had to rebase and then messed up my branch and restarted, and then forgot to commit that file. Fixed!

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `derive(DerivationPath)` derives and returns child DescriptorPublicKey
     - `extend(DerivationPath)` extends and returns DescriptorPublicKey
     - `as_string()` returns DescriptorPublicKey as String
+    - Add `get_height()` and `get_block_hash()` methods on blockchain [#184]
```
notmandatory, line 20: nit on formatting so this doesn't look like it's part of the `interface DescriptorPublicKey`:
```suggestion
  - Add to `interface Blockchain` the `get_height()` and `get_block_hash()` methods [#184]
```

notmandatory (approved):

notmandatory (approved):
ACK 88427e4a0558bcc90b4f3cdda7e6c9df1cbd0dfc

PR #185, opened 2022-08-17
Fix method names to mirror Rust bdk API

### LINKED ISSUE #181: Naming differences between bindings and Rust API
I believe there are two naming differences between the bindings API and the Rust API on the TxBuilder:
1. Our `TxBuilder` uses `get_transactions()` instead of the bdk `list_transactions()` method
2. Our `TxBuilder` uses `get_network()` instead of the bdk `network()` method

Fixing these would create breaking changes but I think we should continue to strive for name parity with the upstream library if we can.

thunderbiscuit:
Fixes #181.

1. `get_network()` is now `network()`
2.  `get_transactions()` is now `list_transactions()`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Rename `get_transactions()` method on `Wallet` interface to `list_transactions()` [#185]
   - Remove `generate_extended_key`, returned ExtendedKeyInfo [#154]
   - Remove `restore_extended_key`, returned ExtendedKeyInfo [#154]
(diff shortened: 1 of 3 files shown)


notmandatory (approved):
ACK 765662e8f8065b63b998a1b36e4b9c147d610d9c

notmandatory (changes requested):
Oops I ACK'd this one too soon. Please also update the CHANGELOG with summary of changed function names.

notmandatory (approved):
ACK 7228fa06a2cc5f60cffccd40600ea1b4339a537a

notmandatory (approved):
ReACK 7da1aaf54141df46def8c54a118d5e071020ebac

Issue #168, opened 2022-06-29
Expose the set_recipients() API from the TxBuilder


thunderbiscuit:
This will allow us to provide a list of recipients instead of calling the `.add_recipient()` method multiple times on the `TxBuilder`.

notmandatory:
@waterst0ne has started looking into this one

notmandatory:
Also need to change the `finish()` method to use this:

https://docs.rs/bdk/latest/bdk/wallet/tx_builder/struct.TxBuilder.html#method.set_recipients

thunderbiscuit:
Awesome work!

PR #186, opened 2022-08-19
Add tx builder set recipients

waterst0ne:
### Description
I just started working on #168 which enables the `tx_builder` to add a list of recipients instead of iterating over the `add_recipient` method multiple times.
- The `set_recipients()` method replaces the precedent list and overwrites it with the new one. 
- You can always use `add_recipient()` to append to the list.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
 - Breaking Changes
   - Rename `get_network()` method on `Wallet` interface to `network()` [#185]
   - Rename `get_transactions()` method on `Wallet` interface to `list_transactions()` [#185]
@@ -24,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `extend(DerivationPath)` extends and returns DescriptorPublicKey
     - `as_string()` returns DescriptorPublicKey as String
   - Add to `interface Blockchain` the `get_height()` and `get_block_hash()` methods [#184]
+  - Add to `interface TxBuilder`  the `set_recipients(recipient: Vec<AddressAmount>)` method [#186] 
 - Interfaces Added [#154]
   - `DescriptorSecretKey`
   - `DescriptorPublicKey`
--- a/src/bdk.udl
+++ b/src/bdk.udl
@@ -168,6 +168,11 @@ dictionary LocalUtxo {
   boolean is_spent;
 };
 
+dictionary AddressAmount {
+  string address;
+  u64 amount;
+};
+
 interface Wallet {
   [Throws=BdkError]
   constructor(string descriptor, string? change_descriptor, Network network, DatabaseConfig database_config);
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -24,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `extend(DerivationPath)` extends and returns DescriptorPublicKey
     - `as_string()` returns DescriptorPublicKey as String
   - Add to `interface Blockchain` the `get_height()` and `get_block_hash()` methods [#184]
+  - Add to `interface TxBuilder`  the `set_recepients(recipients: Vec<AddressAmount>)` method [#186]
```
thunderbiscuit, line 28: Typo here on the word "recipient" in set_recipients:
`set_recipients(recipients: Vec<AddressAmount>)` method

src/bdk.udl
```diff
@@ -168,6 +168,11 @@ dictionary LocalUtxo {
   boolean is_spent;
 };
 
+dictionary AddressAmount {
+ string address;
+ u64    amount;
```
thunderbiscuit, line 173: I don't mind the idea of aligning the fields/methods after the type declaration, but because we don't do it anywhere else in the file I think this should be set similar to the other dictionaries:
```
dictionary AddressAmount {
  string address;
  u64 amount;
}
```

And the realignment be left for a separate PR if it's really where we want to go (but the UDL files are going away anyway so I don't think it's worth considering too hard).

Also note that you need 2 spaces for correct alignment.

src/lib.rs
```diff
@@ -421,6 +424,17 @@ impl TxBuilder {
         })
     }
 
+    fn set_recipients(&self, recipients: Vec<AddressAmount>) -> Arc<Self> {
```
thunderbiscuit, line 427: The addition of the `AddressAmount` type here is a good thing (I ended up doing the same thing manually on the Kotlin side), and I'm wondering if we should also use that type for the `add_recipient()` method above.

src/lib.rs
```diff
@@ -421,6 +424,17 @@ impl TxBuilder {
...
+        let recipients = recipients
+            .iter()
+            .map(|address_amount| (address_amount.address.clone(), address_amount.amount))
```
thunderbiscuit, line 430: Should the `.clone()` method be called on the `amount` as well here?

thunderbiscuit (changes requested):
Overall just a few small things. But the biggest question is whether we should introduce the breaking change that would have `add_recipient()` use the `AddressAmount` type.

src/lib.rs
notmandatory: We are forced to use `AddressAmount` to work in a vector for `add_recipients`.. but for `add_recipient` I think it's best to leave it as it is since it's trivial to still use a `AddressAmount` struct by using it's fields.

src/lib.rs
notmandatory: For `address_amount` it is a native u64 and native values implement the rust `Copy` trait which means the value will be implicitly copied and therefore no `clone()` required.

src/bdk.udl
notmandatory: Good catch, yes I agree the formatting/spacing for the `AddressAmount` dictionary should be formatted and line up with the rest of the UDL file. Unfortunately `cargo fmt` won't do it so it needs to be done manually.

CHANGELOG.md
waterst0ne: thank you thunder.

src/bdk.udl
waterst0ne: I don't remember doing this intentionally but you are right lets keep things uniform. Proposed changes have been updated.

notmandatory:
@waterst0ne oops, it looks like you force pushed an old version somehow. If you need help fixing it back up let me know.

notmandatory (approved):
ACK 8703803bd24dced4d75716a1ff251ce379b60347

notmandatory:
@thunderbiscuit have your changes been addressed? if so you'll need to add an approve before we can merge.

PR #192, opened 2022-09-01
Make `add_recipient()` take any ScriptPubKey 

### LINKED ISSUE #159: Enable building transaction from any given ScriptPubKey
This feature is required to make the language bindings interop with LDK.

The LDK workflow is the following:
1. LDK gives you an output script in raw bytes which you must use to build a ready-to-be-broadcast transaction (the funding transaction)
2. You must give this tx in its raw form back to LDK, which will then broadcast it (potentially using BDK as well, see #157)

I have recreated what is required in Rust [in this small cli tool](https://github.com/thunderbiscuit/txcreator/blob/master/src/create_tx.rs). Of this workflow, I believe only a few lines would be "new" APIs for the bindings, namely:
```rust
// 1. transform the LDK output script from raw bytes into a Script type using either one of:
// let script: Script = Script::from_hex(&output_script_hex).unwrap();
let script: Script = Script::from(output_script_raw);

// 2. extract the tx from the psbt
let funding_tx: Transaction = psbt.extract_tx();

// 3. serialize the tx to give it back to LDK
let funding_tx_encoded = funding_tx.serialize();
```

thunderbiscuit:
Fixes #159.

### Changelog notice
```txt
Breaking Changes:
    - `TxBuilder.add_recipient()` now takes a `Script` instead of an `Address` [#192]
    - `AddressAmount` is now `ScriptAmount` [#192]

New Structs:
    - `Address` and `Script` structs have been added

New features:
  - Add `PartiallySignedBitcoinTransaction.extract_tx()` function.

[#192](https://github.com/bitcoindevkit/bdk-ffi/pull/192)
```

To do:
- [x] Inline documentation
- [x] Clean up commits

### DIFF
--- a/src/bdk.udl
+++ b/src/bdk.udl
@@ -173,8 +173,8 @@ dictionary LocalUtxo {
   boolean is_spent;
 };
 
-dictionary AddressAmount {
-  string address;
+dictionary ScriptAmount {
+  Script script;
   u64 amount;
 };
 
@@ -211,6 +211,8 @@ interface PartiallySignedBitcoinTransaction {
 
   string txid();
 
+  sequence<u8> extract_tx();
+
   [Throws=Error]
   PartiallySignedBitcoinTransaction combine(PartiallySignedBitcoinTransaction other);
 };
@@ -218,7 +220,7 @@ interface PartiallySignedBitcoinTransaction {
 interface TxBuilder {
   constructor();
 
-  TxBuilder add_recipient(string address, u64 amount);
+  TxBuilder add_recipient(Script script, u64 amount);
 
   TxBuilder add_unspendable(OutPoint unspendable);
 
@@ -248,7 +250,7 @@ interface TxBuilder {
 
   TxBuilder add_data(sequence<u8> data);
 
-  TxBuilder set_recipients(sequence<AddressAmount> recipients);
+  TxBuilder set_recipients(sequence<ScriptAmount> recipients);
 
   [Throws=Error]
   PartiallySignedBitcoinTransaction finish([ByRef] Wallet wallet);
@@ -295,4 +297,15 @@ interface DescriptorPublicKey {
   DescriptorPublicKey extend(DerivationPath path);
 
   string as_string();
+};
+
+interface Address {
+  [Throws=Error]
+  constructor(string address);
+
+  Script script_pubkey();
+};
+
+interface Script {
+  constructor(sequence<u8> raw_output_script);
 };
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,8 +1,10 @@
+use bdk::bitcoin::blockdata::script::Script as BdkScript;
 use bdk::bitcoin::hashes::hex::ToHex;
 use bdk::bitcoin::secp256k1::Secp256k1;
 use bdk::bitcoin::util::bip32::DerivationPath as BdkDerivationPath;
+use bdk::bitcoin::util::psbt::serialize::Serialize;
 use bdk::bitcoin::util::psbt::PartiallySignedTransaction;
-use bdk::bitcoin::{Address, Network, OutPoint as BdkOutPoint, Script, Txid};
+use bdk::bitcoin::{Address as BdkAddress, Network, OutPoint as BdkOutPoint, Txid};
 use bdk::blockchain::any::{AnyBlockchain, AnyBlockchainConfig};
 use bdk::blockchain::GetBlockHash;
 use bdk::blockchain::GetHeight;
@@ -35,8 +37,9 @@ use std::sync::{Arc, Mutex, MutexGuard};
 
 uniffi_macros::include_scaffolding!("bdk");
 
-pub struct AddressAmount {
-    pub address: String,
+/// A output script and an amount of satoshis.
+pub struct ScriptAmount {
+    pub script: Arc<Script>,
     pub amount: u64,
 }
 
@@ -302,7 +305,7 @@ impl NetworkLocalUtxo for LocalUtxo {
             },
             txout: TxOut {
                 value: x.txout.value,
-                address: Address::from_script(&x.txout.script_pubkey, network)
+                address: BdkAddress::from_script(&x.txout.script_pubkey, network)
                     .unwrap()
                     .to_string(),
             },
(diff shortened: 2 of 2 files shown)


thunderbiscuit:
The new error is coming from the ffi layer:
```sh
error[E0308]: mismatched types
    --> /Users/user/bdk-ffi/target/debug/build/bdk-ffi-e89d1caffbe2e3c9/out/bdk.uniffi.rs:1491:24
     |
1491 |             Ok(val) => val,
     |                        ^^^ expected struct `Script`, found struct `std::string::String`
```

I can see how it relates to the upgrade: the TxBuilder used to take a String for the address and now takes a script. But at the same time I _think_ I've fixed all Rust code to migrate... not sure where this string is getting delivered from.

thunderbiscuit:
Still required are docs for the new structs.

- [ ] Inline documentation

Cargo.toml
```diff
@@ -15,6 +15,7 @@ name = "bdkffi"
 
 [dependencies]
 bdk = { version = "0.22", features = ["all-keys", "use-esplora-ureq", "sqlite-bundled"] }
+bitcoin = { version = "0.28.1", features = ["use-serde", "base64", "rand"] }
```
tnull, line 18: Not sure we need this import? At least on my side I was able to re-use `bdk::bitcoin` without an issue.

tnull (commented):
From a first look the approach seems reasonable to me. In fact I hand no issue building? I assume above comment regarding a build error has already been addressed?

Cargo.toml
thunderbiscuit: Ah yes good catch. My IDE was showing an error for some reason on the `Script` type but it's totally an IDE bug. In the process of looking into this I had tried to import the type directly from the bitcoin crate to see if it would fix it but then forgot to remove it.

thunderbiscuit:
Oh shoot! Yes Steve and I fixed the problems mentioned above and I forgot to say it here to avoid confusion. Sorry man!

src/lib.rs
```diff
@@ -19,6 +21,7 @@ use bdk::keys::{
     DescriptorSecretKey as BdkDescriptorSecretKey, ExtendedKey, GeneratableKey, GeneratedKey,
 };
 use bdk::miniscript::BareCtx;
+// use bdk::miniscript::Error::Script as MiniscriptScriptError;
```
tnull, line 24: nit: Can be removed.
```suggestion
```

src/lib.rs
```diff
@@ -32,13 +35,15 @@ use std::fmt;
 use std::ops::Deref;
 use std::str::FromStr;
 use std::sync::{Arc, Mutex, MutexGuard};
+// use bitcoin::hashes::serde::Serialize;
```
tnull, line 38: nit: Can be removed.
```suggestion
```

src/lib.rs
```diff
@@ -460,12 +472,44 @@ impl Wallet {
...
+    fn script_pubkey(&self) -> Arc<Script> {
+        Arc::new(Script {
+            script: self.address.script_pubkey(),
+        })
+    }
+}
+
+/// A Bitcoin script.
+#[derive(Clone)]
+pub struct Script {
```
tnull, line 502: Since this is just a wrapper, we maybe could make this less verbose (and avoid the `script.script` parts below) by using a tuple struct?
```suggestion
pub struct Script(BdkScript);
```

tnull (commented):
Generally LGTM, just a few nits.

src/lib.rs
thunderbiscuit: This is interesting, I didn't think of using tuple structs for those (I think we could potentially do the same thing for `Address(BdkAddress)`. Let me take a stab at refactoring and see what it looks like.

thunderbiscuit:
@tnull I agree that the `script.script` is rather verbose. I tried using the tuple struct to refactor and it appears to work, but because we need to access the internals of the tuple struct we end up having to use `script.0.clone()` (unless there is another way to do this I'm not seeing?). I don't have a strong opinion either way, it's mostly replacing the `script.script` by `script.0`.

Here are what the methods look like in the TxBuilder:
```rust
fn add_recipient(&self, script: Arc<Script>, amount: u64) -> Arc<Self> {
    let mut recipients: Vec<(BdkScript, u64)> = self.recipients.clone();
    recipients.append(&mut vec![(script.0.clone(), amount)]);
    Arc::new(TxBuilder {
        recipients,
        ..self.clone()
    })
}

fn set_recipients(&self, recipients: Vec<ScriptAmount>) -> Arc<Self> {
    let recipients = recipients
        .iter()
        .map(|script_amount| (script_amount.script.0.clone(), script_amount.amount))
        .collect();
    Arc::new(TxBuilder {
        recipients,
        ..self.clone()
    })
}
```

notmandatory:
should update changelog above with:

New features:
  - Add `PartiallySignedBitcoinTransaction.extract_tx()` function.

thunderbiscuit:
Done and done. @notmandatory do you have an opinion as to whether to use the tuple struct or the normal struct? I'm ok with both, but we should be consistent if we use the tuple struct and also apply it elsewhere we have a wrapper with only one field. 

I wonder if it makes the code less explicit just because you end up using numbers instead the wrapper/field, but it's not a big deal either way.

tnull:
> @tnull I agree that the `script.script` is rather verbose. I tried using the tuple struct to refactor and it appears to work, but because we need to access the internals of the tuple struct we end up having to use `script.0.clone()` (unless there is another way to do this I'm not seeing?). I don't have a strong opinion either way, it's mostly replacing the `script.script` by `script.0`.

Right, I personally prefer the shorter for of the definition and `script.0` over `script.script`, but don't have strong feelings about it either. So feel free to leave as is.

thunderbiscuit:
As per a discussion in the dev call we'll be merging this PR without using the Tuple Structs and leave the migration to the Tuple Structs to a separate PR (since there are a few different places where these refactorings should be applied).

tnull (approved):

src/bdk.udl
```diff
@@ -296,3 +298,14 @@ interface DescriptorPublicKey {
 
   string as_string();
 };
+
+interface Address {
+  [Throws=BdkError]
```
notmandatory, line 303: This should be changed based on #201 
```suggestion
  [Throws=Error]
```

src/lib.rs
```diff
@@ -1012,7 +1057,7 @@ mod test {
         let tx_builder = TxBuilder::new()
             .drain_wallet()
             .drain_to(drain_to_address.clone());
-        //dbg!(&tx_builder);
+        dbg!(&tx_builder);
```
notmandatory, line 1060: This `dbg!` should be removed

notmandatory (approved):
ACK 7aafc64f701126e670838c02a81a5f3c0a917ad5

But prior to merging please remove the `dbg!`.

PR #194, opened 2022-09-02
Update uniffi-bindgen to 0.19.5

thunderbiscuit:
As per mozilla/uniffi-rs#1322, the new argument introduced in the `generate_bindings()` function should be set to `None` unless we use the new macro-based approach.

This PR unpins the uniffi-bindgen library while also ensuring the CI can build the library properly.

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -10,9 +10,9 @@ jobs:
     strategy:
       matrix:
         rust:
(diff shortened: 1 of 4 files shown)


thunderbiscuit:
Ok so now we're getting an error because of `0.19.6`. Hold on while I figure this out.

notmandatory (approved):
ACK 8166f820b4e4d6e52887de6406ab6b137f8ed43b

PR #197, opened 2022-09-20
Rename KeychainKind variants to ExternalKeys and InternalKeys

notmandatory:
### Description

I ran into an issue with updating `bdk-swift` to the latest `bdk-ffi` version (0.9.0). There's a conflict with the `KeychainKind.Internal` variant, in Swift `uniffi-rs` maps the variant name to `internal` which isn't allowed because it's a Swift keyword.

For now I think the best approach is to created a `bdk-ffi` version of  `KeychainKind` with variants `ExternalKeys` and `InternalKeys` so there won't be any conflict in Swift. 

I also bumped the `bdk-ffi` version to `0.10.0` so we can tag this change as a new API breaking version and make a new `bdk-swift` release. I don't think we need to do new Kotlin and Python releases, but we should warn those users of the breaking change when we do their next releases.

### Notes to the reviewers

This is a breaking change for any apps that use the `LocalUtxo.keychain` field value for filtering or anything else where the original enum variant type is used.  

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,7 +23,7 @@ use bdk::wallet::tx_builder::ChangeSpendPolicy;
 use bdk::wallet::AddressIndex as BdkAddressIndex;
 use bdk::wallet::AddressInfo as BdkAddressInfo;
 use bdk::{
-    Balance as BdkBalance, BlockTime, Error, FeeRate, KeychainKind, SignOptions,
+    Balance as BdkBalance, BlockTime, Error, FeeRate, KeychainKind as BdkKeychainKind, SignOptions,
     SyncOptions as BdkSyncOptions, Wallet as BdkWallet,
 };
 use std::collections::HashSet;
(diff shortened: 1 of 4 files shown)


thunderbiscuit:
I wonder if this would have been fixed by https://github.com/mozilla/uniffi-rs/pull/1170

src/lib.rs
```diff
@@ -282,6 +282,14 @@ pub struct TxOut {
     address: String,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum KeychainKind {
+    /// External keys.
+    ExternalKeys = 0,
```
thunderbiscuit, line 288: Why do these enums require numbers?

notmandatory: In rust it allows you to de-reference the enum to this number, which I think is used when you want to create a new HD path and use 0 for your external path, and 1 for your internal.  I left it here uncase we want to do something similar in the future. 

Also I saw  your note on the `uniffi-rs` project about this issue and I think it's worth waiting on this PR to see what they say about fixing the issue on their side, or at least if they say it's possible.

notmandatory:
Created https://github.com/mozilla/uniffi-rs/pull/1350 to fix this issue the right way.

PR #199, opened 2022-09-26
Add ability to retrieve private keys as bytes

### LINKED ISSUE #188: Add ability to retrieve private master key from a BIP39 seed
This can be used to seed an LDK [KeysManager](https://docs.rs/lightning/0.0.110/lightning/chain/keysinterface/struct.KeysManager.html) with the first 32 bytes from this 64-byte seed.

This enables the entropy we give to LDK to be within the backup system of the on-chain wallet, and allow for 12-to-24-word mnemonics with optional passphrases.

thunderbiscuit:
This feature is needed for compatibility with LDKLite, where the initial entropy given to LDK is the private key of the root of the BIP32 derivation tree.

Closes #188

### Changelog notice
```txt
APIs Added:
    - Add `secret_key_bytes()` method on the `DescriptorSecretKey` [#199]

[#199](https://github.com/bitcoindevkit/bdk-ffi/pull/199)
```

### Notes to the reviewers
I just want to make sure I understand why the `BdkDescriptorSecretKey::SinglePriv(_)` is unreachable (will fix the code to add the `unreachable!()` if my understanding is correct. Is it because we currently only build the DescriptorSecretKey from mnemonics, and always make them extendable?

#### All Submissions:

#### New Features:

### DIFF
--- a/src/bdk.udl
+++ b/src/bdk.udl
@@ -283,7 +283,7 @@ interface DescriptorSecretKey {
 
   DescriptorPublicKey as_public();
 
-  sequence<u8> secret_key_bytes();
+  sequence<u8> secret_bytes();
 
   string as_string();
 };
--- a/src/bdk.udl
+++ b/src/bdk.udl
@@ -283,6 +283,8 @@ interface DescriptorSecretKey {
 
   DescriptorPublicKey as_public();
 
+  sequence<u8> secret_key_bytes();
+
   string as_string();
 };
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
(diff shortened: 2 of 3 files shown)


src/bdk.udl
```diff
@@ -280,6 +280,8 @@ interface DescriptorSecretKey {
 
   DescriptorPublicKey as_public();
 
+  sequence<u8> secret_key_bytes();
```
notmandatory, line 283: How about shortening the name to `secret_bytes()` same as the internal bdk function?

notmandatory (approved):
utACK 1b9bd6d0344c9b193f11881a219bf839eafbdbd9

Up to you if you'd also like to shorten the function name to `secret_bytes()`.

src/bdk.udl
thunderbiscuit: Yeah let's do it. In this case I had made a deliberate choice because I actually don't like the API, (`DescriptorSecretKey.secret_bytes`? it's not explicit to me that it's about to return you the secret key). But I agree we should stick to the names we have in the Rust libs (and if I don't like it I should take it up with the upstream! haha).

Fixed.

Issue #205, opened 2022-10-10
A few concurrency-related issues Block has faced while using bdk-ffi


cgarrett-square:
At Block, We're using the BDK via the bdk-ffi bindings in our iOS and Android Bitcoin Wallet. It's been incredibly helpful so far! We [surfaced a few issues related to synchronicity](https://discord.com/channels/753336465005608961/759431700807286795/1026936461645316106) in the BDK #mobile Discord channel and are creating this issue to track them. As discussed in the thread, I'm starting with a single issue with multiple related topics. But I'm happy to break this into three separate GitHub issues if the team prefers! Also, I'm acknowledging I'm not using the issue template until we've decided if this should be a single issue or three. But I'm happy to provide the additional detail requested in the template when helpful. 

# Issue 1: Concurrent Calls to Wallet in BDK
If we make calls concurrently to the BDK wallet, we sometimes experience the main thread hanging on a mutex lock. The calls involved in one example were `sync` and `get_balance`. Below is a sample stack trace from an app hang when we were calling into the BDK from multiple threads: `get_balance` on Thread 1 (main) and `sync` on Thread 6 (background): 

![block-mutex-wait](https://user-images.githubusercontent.com/72396953/194921183-944d1839-8824-4715-b53b-c5cdab3fe83b.png)

We changed our implementation to ensure all calls made to the BDK wallet are serialized, which fixes the issue. But we worry about performance implications with this solution. We call `sync` often, and have also observed that call hanging and not returning (separate issue below). So, if our calls are forced to be serialized, we could end up in a situation where we have to wait on a call to `sync` (that may never return) to finish before executing some other work.

# Issue 2: Long-running / Non-cancellable Calls
[...]

PR #207, opened 2022-10-13
Add compact block filters blockchain feature using the Nakamoto client

### LINKED ISSUE #118: Add compact block filters blockchain support

thunderbiscuit:
Work in Progress.

Fixes #118.

### Description
This PR adds the compact block filters blockchain option.

### Notes to the reviewers
Not ready for merge yet. Lots to be tested still.

Questions I still want to answer:
- [ ] How much size does it add to the final binary we ship for Android/iOS?
- [ ] Full testing on testnet/mainnet

### Changelog notice
```txt
## Added Features
    - Compact Block Filters client ([Nakamoto]) [#207]

[Nakamoto](https://github.com/cloudhead/nakamoto)
[#207](https://github.com/bitcoindevkit/bdk-ffi/pull/207)
```

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -163,6 +163,7 @@ dictionary CompactFiltersConfig {
   sequence<string> addresses;
   Network network;
   string storage_dir;
+  u32 skip_blocks;
 };
 
 
--- a/bdk-ffi/src/blockchain.rs
+++ b/bdk-ffi/src/blockchain.rs
@@ -55,11 +55,16 @@ impl Blockchain {
                 peers.push(peer);
 
 
+
                 AnyBlockchainConfig::CompactFilters(CompactFiltersBlockchainConfig {
                     peers: peers,
                     network: config.network,
                     storage_dir: config.storage_dir,
-                    skip_blocks : None,
+                    skip_blocks : match usize::try_from(config.skip_blocks) {
+                        Ok(value) => Some(value),
+                        Err(_) => None,
+                    }
+                      ,
                 })
             }
             BlockchainConfig::Rpc { config } => AnyBlockchainConfig::Rpc(BdkRpcConfig {
@@ -237,6 +242,7 @@ pub struct CompactFiltersConfig {
     pub addresses: Vec<String>,
     pub network: Network,
     pub storage_dir: String,
+    pub skip_blocks: u32,
 }
 
 /// Type that can contain any of the blockchain configurations defined by the library.


thunderbiscuit:
I decided to try it on my desktop using bdk-jvm and a small Kotlin cli wallet I made a while back. It seems to be syncing _something_ so far, so that's good. But as a side note the `progress` float returned seem to be going up and down. Sometimes in the 60s sometimes in the 40s, back and forth.

float

thunderbiscuit:
Oh darn it worked! Took about 3-4 minutes and the wallet is displaying the correct balance for the descriptor. 🔥 🔥 🔥

thunderbiscuit:
Ok so so far here is what is happening:

1. On desktop and using bdk-jvm the new blockchain appears to be working
2. On Android not at all, throwing this error: `Cbf(Nakamoto(Handle(Disconnected)))` (both in the emulator and on a physical device, same error)

The odd thing is that these two libraries are pretty much duplicates of each other. So the issue is definitely coming from something on Android.

@cloudhead @rajarshimaitra any ideas? I tried looking through the Nakamoto codebase but couldn't quite pick out the `Handle(Disconnected)` error and what could be triggering it.

rajarshimaitra:
Thanks @thunderbiscuit .. Thats super progress.. Although I am as clueless about the `HandleDisconnect` error.. Just making some wild guess.. The sync process starts by spawning a separate thread that runs the nakamoto client process in background.. The `HandleDisconnect` could mean that for some reason that thread is dropped or terminated and the main bdk thread cannot communicate with it anymore..

Is that something that happens on android? Does it require special consideration for spawning threads?

cloudhead:
> Thanks @thunderbiscuit .. Thats super progress.. Although I am as clueless about the `HandleDisconnect` error.. Just making some wild guess.. The sync process starts by spawning a separate thread that runs the nakamoto client process in background.. The `HandleDisconnect` could mean that for some reason that thread is dropped or terminated and the main bdk thread cannot communicate with it anymore..
> 
> Is that something that happens on android? Does it require special consideration for spawning threads?

This is likely what's going on.

thunderbiscuit:
Ok good stuff guys, thanks for chiming in. I'll need to dig a bit more into this. It means that somehow the connection between the native binary and the glue code is handled differently on Android than it is server-side, and I don't know that we have encountered that before. My next step will be to make sure it works in the unit tests (which run on the dev machine and not in the emulator/device).

Side note, I tested with the Python library and it worked right off the bat.

thunderbiscuit:
Folks it looks like we have liftoff!

<center>
    Nakamoto
</center>

The issue was that I had given a default value of `./nakamoto` for the `datadir` field of the `CBFConfig` in the ffi layer, and in my tests on desktop/server/mobile I left that field `null` so it was using the default path. This worked on the server because I called my test wallet from a directory where I had write permission. 

On Android however, the process is not started from a place where you have write permission (and I haven't quite figured out _where_ this default path leads to the attempted directory creation but it doesn't really matter for now), and so that default path simply doesn't create the `.nakamoto` directory (I'd be nice to get a specific error for that maybe). Subsequent sync calls would then fail. The trick was to provide a valid path for the app to write the directory in. On Android this looks like 

```kotlin
val path: String = applicationContext.filesDir.toString()
val nakamotoConfig = CbfConfig(
    Network.TESTNET,
    "$path/bdk-nakamoto",
)
```

I suspect the easiest way to deal with this in the short run is to make the `datadir` field required on the FFI layer (as per my latest commit), but there might be other ways as well.

A few notes for myself/future reference:
1. On the server a full sync of testnet was about 300-400 MB for the 3 files (`filters.db`, `headers.db`, `peers.json`)
2. The process times out on the UI thread (no surprise there), but that means we end up launching it in a coroutine, and that's where we can get into tricky situations (#205)

cloudhead:
Great work!

In terms of catching the error, is there something nakamoto could have done better? I suspect there would have been an error logged, though I recently reworked all the logging because there were some issues in 0.3.0.

On your notes:
* Mainnet should take about 100 MB total disk space, since there are about 3x less blocks compared to testnet.
* Since <https://github.com/cloudhead/nakamoto/pull/105/commits/1c5afc977fdeb6ab4dd869ec339870b18b03c174>, the block header/filter loading is interruptible, though I'm still working on the API
* I'm curious about the binary size, since the `nakamoto-node` which contains everything that would be imported into bdk is only 5.7MB on my system (release mode).

andreasgriffin:
I tried to get CBF working in python, and was successfull on mainnet (not regtest unfortunately) ( see [diff](https://github.com/bitcoindevkit/bdk-ffi/compare/master...andreasgriffin:bdk-ffi:compact_filters2)  on my branch https://github.com/andreasgriffin/bdk-ffi/commits/compact_filters2   )

image

andreasgriffin:
I really like CBF because it has good privacy, while also offering great speed.

@thunderbiscuit : Does it makes sense to develop https://github.com/andreasgriffin/bdk-ffi/commits/compact_filters2  into a PR right now, even though bdk has a major restructuring ahead?

thunderbiscuit:
Hey thanks for taking a look! 

Instead of making a new PR on this repo directly, I think it might make sense to keep this PR up to date (it's quite old at this point) so that it's easier to review and keep up to date in the future. Would you mind making your PR on my fork instead? Your contributions will still be shown because your commits will actually show up on this PR once I merge your PR on my branch.

I'll rebase this (it's about 500 commits behind) and everything will break, but would you be interested in opening a PR on my [cbf branch](https://github.com/thunderbiscuit/bdk-ffi/tree/cbf) with your working implementation?

To be transparent with you I will not actually consider merging this in the FFI layer until the 1.0 release of BDK, which will impact significantly how sync is done (and so you can expect sizeable changes to the compact block filter crate as well). But for anyone who would like to experiment with it and for testing the advances made by the Rust BDK team on CBF it would be great to have this PR fairly up-to-date and ready to play with.

thunderbiscuit:
Ok I've rebased the changes. If you PR on my CBF branch I'll take the time to review and merge there, so we can get this PR in working order!

thunderbiscuit:
I'm closing this in favour or #379.

PR #212, opened 2022-10-19
Add Mnemonic type and key constructors from string

notmandatory:
### Description

Add the ability to create PublicDescriptorKey or SecretDescriptorKey from a string, including non-extended key types like WIF. An error is thrown if the string isn't a valid key. 

### Notes to the reviewers

The Mnemonic type also needs to be added back since only one constructor type can throw an error per the UDL limitations, and that means we need to type check the mnemonic words in the Mnemonic constructor.   

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/bdk.udl
+++ b/src/bdk.udl
@@ -1,6 +1,5 @@
 namespace bdk {
-  [Throws=Error]
-  string generate_mnemonic(WordCount word_count);
+
 };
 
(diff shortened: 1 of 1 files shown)


notmandatory:
Closing this PR in favor of #219 .

Issue #206, opened 2022-10-12
The Great FFI Bindings Migration


thunderbiscuit:
Just opening this so we have a place to formally discuss the migration of the individual bindings repositories into the bdk-ffi repository. 

This idea has been discussed for a few months now, and I think we have a rough consensus that it would be a change for the better, but there is still lots to work out.

## Update
The migration is going well. A few tasks left to do:
- [x] Clean up licenses
- [x] Re-write the root readme, and maybe add library-level readmes that are appropriate for the new setup
- [x] Bring the PGP key file in the root
- [x] Remove `.editorconfig` file
- [x] Fix Audit workflow

thunderbiscuit:
Some links around how to move a repository into another one while keeping the history of the first one:

1. https://stackoverflow.com/questions/13040958/merge-two-git-repositories-without-breaking-file-history

thunderbiscuit:
A few tasks left to do:
- [ ] Clean up licenses
- [ ] Re-write the root readme, and maybe add library-level readmes that are appropriate for the new setup
- [ ] Bring the PGP key file in the root
- [ ] Remove `.editorconfig` file

PR #217, opened 2022-10-26
Add bdk-python

thunderbiscuit:
This PR brings the bdk-python repo into the bdk-ffi repository.

It's a big change, and I still have one or two unanswered questions (but I thought it would be worthwhile to open a draft PR anyway):
- I'm not sure how I can set the exact version of bdk-ffi to be used in the building of bdk-python. At the moment that's defined as a pure path in the `setup.py` file, but I'd need it to be able to take a git repo and a tag...

This is part of #206.


notmandatory:
I think when we migrate the bindings into the `bdk-ffi` project they will all be versioned together. So when we tag a new `bdk-ffi` version that will also be the version we use to build and release the languages.

thunderbiscuit:
Closing in favour of #218.

PR #218, opened 2022-10-27
Add API for generating master keys directly

thunderbiscuit:
Rough draft of the new API for generating `DescriptorSecretKey` without going through the mnemonics step.

A few questions still left to figure out:
1. Do we want the `generate_from_entropy()` method? I added it as a bonus but we should still discuss if there is a need for it.
2. I'd like to make sure the random number generation is handled properly; this is the very core of the security of the keys. I tried to reproduce what is done for the mnemonic, [see the code here](https://docs.rs/bdk/latest/src/bdk/keys/mod.rs.html#631-637).
3. I'm not sure if I really need to have the `rand` crate imported here (we don't need it on the mnemonic step for example). There might be a way to get this done internally in bdk without generating the entropy at the ffi layer.

### DIFF
--- a/src/bdk.udl
+++ b/src/bdk.udl
@@ -283,6 +283,12 @@ interface DescriptorSecretKey {
   [Throws=BdkError]
   constructor(Network network, string mnemonic, string? password);
 
+  [Name=generate, Throws=BdkError]
+  constructor(Network network);
+
+  [Name=generate_from_entropy, Throws=BdkError]
+  constructor(Network network, sequence<u8> entropy);
+
   [Throws=BdkError]
   DescriptorSecretKey derive(DerivationPath path);
 
--- a/src/bdk.udl
+++ b/src/bdk.udl
@@ -283,10 +283,10 @@ interface DescriptorSecretKey {
   [Throws=BdkError]
   constructor(Network network, string mnemonic, string? password);
 
-  [Name=generate, Throws=BdkError]
+  [Name=generate]
   constructor(Network network);
 
-  [Name=generate_from_entropy, Throws=BdkError]
+  [Name=generate_from_entropy]
   constructor(Network network, sequence<u8> entropy);
 
   [Throws=BdkError]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,7 +1,7 @@
 use bdk::bitcoin::blockdata::script::Script as BdkScript;
 use bdk::bitcoin::hashes::hex::ToHex;
 use bdk::bitcoin::secp256k1::Secp256k1;
-use bdk::bitcoin::util::bip32::DerivationPath as BdkDerivationPath;
+use bdk::bitcoin::util::bip32::{DerivationPath as BdkDerivationPath, ExtendedPrivKey};
 use bdk::bitcoin::util::psbt::serialize::Serialize;
 use bdk::bitcoin::util::psbt::PartiallySignedTransaction;
 use bdk::bitcoin::{Address as BdkAddress, Network, OutPoint as BdkOutPoint, Txid};
@@ -28,6 +28,7 @@ use bdk::{
     Balance as BdkBalance, BlockTime, Error as BdkError, FeeRate, KeychainKind, SignOptions,
     SyncOptions as BdkSyncOptions, Wallet as BdkWallet,
 };
+use rand::{thread_rng, Rng};
 use std::collections::HashSet;
 use std::convert::{From, TryFrom};
 use std::fmt;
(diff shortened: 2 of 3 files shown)


thunderbiscuit:
Would be happy to hear initial feedback from @kirillzh and @jesseposner. Does this address your need? Did you have another API shape in mind?

In Kotlin the call site currently looks like this:
```kotlin
val bip32RootKey: DescriptorSecretKey = DescriptorSecretKey.generate(Network.TESTNET)
val bip84ExtendedKey: DescriptorSecretKey = bip32RootKey.extend(DerivationPath("m/84h/1h/0h/0"))
val descriptor = "wpkh(${bip84ExtendedKey.asString()})"
println(descriptor)
// wpkh(tprv8ZgxMBicQKsPeSUx4F4fC6Vc4rWBxfJV1FVBWWLxqapQpSdZ2huNuhcRUdWPjNFbQmQ285bjXV7jayWU7RyL5XSNMSC5J6wf6gj1RCCh43G/84'/1'/0'/0/*)
```

src/bdk.udl
```diff
@@ -283,6 +283,12 @@ interface DescriptorSecretKey {
   [Throws=BdkError]
   constructor(Network network, string mnemonic, string? password);
 
+  [Name=generate, Throws=BdkError]
```
notmandatory, line 286: You should be able to remove the `Throws=BdkError` for both of these new constructors and update code to just return `Self`.

thunderbiscuit: Done!

notmandatory (approved):
ACK 7b12b6053968caa9c19d2f80069214bb1ef2ef0c

Looks good!

src/lib.rs
```diff
@@ -28,6 +28,7 @@ use bdk::{
     Balance as BdkBalance, BlockTime, Error as BdkError, FeeRate, KeychainKind, SignOptions,
     SyncOptions as BdkSyncOptions, Wallet as BdkWallet,
 };
+use rand::{thread_rng, Rng};
```
dhruv-1001, line 31: Would it be okay to use `thread_rng`, `Rng` this way?
```
use bdk::bitcoin::secp256k1::rand::{thread_rng, Rng};
```
This way we don't have to add `rand` to `Cargo.toml`

thunderbiscuit: Oh good idea. @notmandatory do I understand correctly that if this is coming from re-exports it's good to go? I think in this case `bdk` re-exports `bitcoin` which re-exports `secp256k1` which re-exports `rand`... lol.

notmandatory: The latest version of `bdk` (0.24) updated `rand` version to 0.8, but now that I take a look it's not exported, so I think the best approach is to bring it in as our own dependency as you're doing.

notmandatory:
I just took another look at how entropy and Mnemonics work in `bdk` and now I think the right way to go is do go ahead with a #219 but add `Mnemonic.from_entropy()` function also. Then change the constructor for `DescriptorSecretKey` to take a `Mnemonic` structure instead of a mnemonic String. The advantage to using a full Mnemonic structure created from entropy, is it gives on-chain wallets created from some other entropy the option to still back it up with Mnemonic seed words. Also this means we can use the existing `Mnemonic.from_entropy()` `Mnemonic::generate()` functions and we'll stay closer to how the `bdk` API works.

jesseposner:
@notmandatory Does the `bdk` API recommend a mnemonic to generate an `xprv`? I thought something like [this](https://github.com/bitcoindevkit/bdk/blob/5720e38033778c905e31f4ab88aa880cfd48a51b/src/keys/mod.rs#L953-L959) was okay:

```rust
    fn test_keys_generate_xprv() {
        let generated_xprv: GeneratedKey<_, miniscript::Segwitv0> =
            bip32::ExtendedPrivKey::generate_with_entropy_default(TEST_ENTROPY).unwrap();


        assert_eq!(generated_xprv.valid_networks, any_network());
        assert_eq!(generated_xprv.to_string(), "xprv9s21ZrQH143K4Xr1cJyqTvuL2FWR8eicgY9boWqMBv8MDVUZ65AXHnzBrK1nyomu6wdcabRgmGTaAKawvhAno1V5FowGpTLVx3jxzE5uk3Q");
    }
```

jesseposner:
@notmandatory The only other reason not to create a mnemonic first is that a mnemonic is limited to 256 bits of entropy, whereas BIP32 allows for 512 bits of entropy. However, anything over 256 bits is overkill and BIP32 recommends 256 bits, so as a practical matter I don't think much is being lost with the 256 limit.

So the benefits of the additional flexibility of not requiring a mnemonic are pretty minor. It's nice to have but I also agree that it's a reasonable simplification to require a mnemonic.

PR #219, opened 2022-10-28
Added Mnemonic Interface

dhruv-1001:
### Description
This PR adds `interface Mnemonic` which will make the API to generate new DescriptorSecretKey have type-safe arguments.

### Notes to the reviewers

This PR doesn't have any issue linked to it, as it was discusses on a call during the implementation of `DescriptorSecretKey` (PR #154). It was discussed to make `Mnemonic` an interface and use that instead of string `Mnemonic` so that the API to generate `DescriptorSecretKey` doesn't have any potential failure (like in case it's provided with incorrect Mnemonic words).

APIs added
```
// generates and returns Mnemonic with random entropy
Mnemonic::new(word_count: WordCount) -> Self { ... } 
// converts string Mnemonic to Mnemonic type with error (in case of incorrect string Mnemonic)
Mnemonic::from_str(mnemonic: String) -> Result<Self, BdkError> { ... }
// generates and returns Mnemonic with given entropy
Mnemonic::from_entropy(entropy: Vec<u8>) -> Result<Self, BdkError> {...}
// view mnemonic as string
Mnemonic::as_string(&self) -> String { ... }
```
Along with some changes to `DescriptorSecretKey::new()` to fit these new APIs

### Changelog notice
```
- Added Struct Mnemonic with following methods [#219]
  - new(word_count: WordCount) generates and returns Mnemonic with random entropy
  - from_str(mnemonic: String) converts string Mnemonic to Mnemonic type with error
  - from_entropy(entropy: Vec<u8>) generates and returns Mnemonic with given entropy
  - as_string() view Mnemonic as string
- API removed [#219]
  - generate_mnemonic(word_count: WordCount)

[#219](https://github.com/bitcoindevkit/bdk-ffi/pull/219)
```
#### All Submissions:

#### New Features:

#### Bugfixes:

    * Top level function `generate_mnemonic(...)` was removed

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,7 +15,7 @@ use bdk::blockchain::{Blockchain as BdkBlockchain, Progress as BdkProgress};
 use bdk::database::any::{AnyDatabase, SledDbConfiguration, SqliteDbConfiguration};
 use bdk::database::{AnyDatabaseConfig, ConfigurableDatabase};
 use bdk::descriptor::DescriptorXKey;
-use bdk::keys::bip39::{Language, Mnemonic, WordCount};
+use bdk::keys::bip39::{Language, Mnemonic as BdkMnemonic, WordCount};
 use bdk::keys::{
     DerivableKey, DescriptorPublicKey as BdkDescriptorPublicKey,
     DescriptorSecretKey as BdkDescriptorSecretKey, ExtendedKey, GeneratableKey, GeneratedKey,
@@ -857,10 +857,40 @@ impl BumpFeeTxBuilder {
     }
 }
 
-fn generate_mnemonic(word_count: WordCount) -> Result<String, BdkError> {
-    let mnemonic: GeneratedKey<_, BareCtx> =
-        Mnemonic::generate((word_count, Language::English)).unwrap();
-    Ok(mnemonic.to_string())
+/// Mnemonic phrases are a human-readable version of the private keys.
+/// Supported number of words are 12, 15, 18, 21 and 24.
+struct Mnemonic {
+    internal: BdkMnemonic,
+}
+
+impl Mnemonic {
+    /// Generates Mnemonic with a random entropy
+    fn new(word_count: WordCount) -> Self {
+        let generated_key: GeneratedKey<_, BareCtx> =
+            BdkMnemonic::generate((word_count, Language::English)).unwrap();
+        let mnemonic = BdkMnemonic::parse_in(Language::English, generated_key.to_string()).unwrap();
+        Mnemonic { internal: mnemonic }
+    }
+
+    /// Parse a Mnemonic with given string
+    fn from_str(mnemonic: String) -> Result<Self, BdkError> {
+        BdkMnemonic::from_str(&mnemonic)
+            .map(|m| Mnemonic { internal: m })
+            .map_err(|e| BdkError::Generic(e.to_string()))
+    }
+
+    /// Create a new Mnemonic in the specified language from the given entropy.
+    /// Entropy must be a multiple of 32 bits (4 bytes) and 128-256 bits in length.
+    fn from_entropy(entropy: Vec<u8>) -> Result<Self, BdkError> {
+        BdkMnemonic::from_entropy(entropy.as_slice())
+            .map(|m| Mnemonic { internal: m })
+            .map_err(|e| BdkError::Generic(e.to_string()))
+    }
+
+    /// Returns Mnemonic as string
+    fn as_string(&self) -> String {
+        self.internal.to_string()
+    }
 }
 
 struct DerivationPath {
@@ -882,19 +912,18 @@ struct DescriptorSecretKey {
 }
 
 impl DescriptorSecretKey {
-    fn new(network: Network, mnemonic: String, password: Option<String>) -> Result<Self, BdkError> {
-        let mnemonic = Mnemonic::parse_in(Language::English, mnemonic)
-            .map_err(|e| BdkError::Generic(e.to_string()))?;
-        let xkey: ExtendedKey = (mnemonic, password).into_extended_key()?;
+    fn new(network: Network, mnemonic: Arc<Mnemonic>, password: Option<String>) -> Self {
+        let mnemonic = mnemonic.internal.clone();
+        let xkey: ExtendedKey = (mnemonic, password).into_extended_key().unwrap();
         let descriptor_secret_key = BdkDescriptorSecretKey::XPrv(DescriptorXKey {
             origin: None,
             xkey: xkey.into_xprv(network).unwrap(),
             derivation_path: BdkDerivationPath::master(),
             wildcard: bdk::descriptor::Wildcard::Unhardened,
         });
-        Ok(Self {
+        Self {
             descriptor_secret_key_mutex: Mutex::new(descriptor_secret_key),
-        })
+        }
     }
 
     fn derive(&self, path: Arc<DerivationPath>) -> Result<Arc<Self>, BdkError> {
@@ -1118,9 +1147,8 @@ mod test {
     }
 
     fn get_descriptor_secret_key() -> DescriptorSecretKey {
-        let mnemonic =
-        "chaos fabric time speed sponsor all flat solution wisdom trophy crack object robot pave observe combine where aware bench orient secret primary cable detect".to_string();
-        DescriptorSecretKey::new(Testnet, mnemonic, None).unwrap()
+        let mnemonic = Mnemonic::from_str("chaos fabric time speed sponsor all flat solution wisdom trophy crack object robot pave observe combine where aware bench orient secret primary cable detect".to_string()).unwrap();
+        DescriptorSecretKey::new(Testnet, Arc::new(mnemonic), None)
     }
 
     fn derive_dsk(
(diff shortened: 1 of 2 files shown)


notmandatory:
Thanks for the PR, but after some discussions I had with @thunderbiscuit last week I think we may not need the "Mnemonic" data structure anymore.. since the user can use a String for it. Originally I thought we needed the error checking to happen in the "Mnemonic" constructor but we can also throw the error in an alternative constructor for "DescriptorSecretKey" that takes a mnemonic String.

notmandatory:
This looks good, can you also add a `Mnemonic.from_entropy()` constructor? See also https://github.com/bitcoindevkit/bdk-ffi/pull/218#issuecomment-1296340467.

dhruv-1001:
Added the `Mnenomic.from_entropy()` constructor

src/lib.rs
```diff
@@ -857,10 +857,39 @@ impl BumpFeeTxBuilder {
     }
 }
 
-fn generate_mnemonic(word_count: WordCount) -> Result<String, BdkError> {
-    let mnemonic: GeneratedKey<_, BareCtx> =
-        Mnemonic::generate((word_count, Language::English)).unwrap();
-    Ok(mnemonic.to_string())
+struct Mnemonic {
+    mnemonic_mutex: Mutex<BdkMnemonic>,
```
notmandatory, line 861: I think we may not need to wrap this inner `BdkMnemonic` in a Mutex. But I still need to test it out to be sure.
```suggestion
    inner: BdkMnemonic,
```

dhruv-1001: Okay, so it should be like this instead
```
struct Mnemonic {
    inner: BdkMnemonic
}

impl Mnemonic {
    fn new(...) -> Arc<Self> {...}
    fn from_string(...) -> Result<Arc<Self>, BdkError> {...}
    fn from_entropy(...) -> Result<Arc<Self>, BdkError> {...}
    fn as_string(...) -> String {...}
}
```
Right?

notmandatory: Thanks, yes but the API should look the same, uniffi-rs is smart enough to know that since it's marked as a constructor in the UDL file that it will add an `Arc` to the result in the glue code it makes.
```rust
struct Mnemonic {
    inner: BdkMnemonic
}

impl Mnemonic {
    fn new(...) -> Self {...}
    fn from_string(...) -> Result<Self, BdkError> {...}
    fn from_entropy(...) -> Result<Self, BdkError> {...}
    fn as_string(...) -> String {...}
}
```

dhruv-1001: Updated `Mnemonic` struct according to suggestion. Resolving conversation.

src/lib.rs
```diff
@@ -857,10 +857,33 @@ impl BumpFeeTxBuilder {
...
+    fn new(word_count: WordCount) -> Self {
+        let generated_key: GeneratedKey<_, BareCtx> =
+            BdkMnemonic::generate((word_count, Language::English)).unwrap();
+        let mnemonic = BdkMnemonic::parse_in(Language::English, generated_key.to_string()).unwrap();
+        Mnemonic { internal: mnemonic }
+    }
+
+    fn from_string(mnemonic: String) -> Result<Self, BdkError> {
+        BdkMnemonic::parse_in(Language::English, mnemonic)
+            .map(|x| Mnemonic { internal: x })
```
notmandatory, line 874: this is a small nit.. but can you use `m` here? similar to how you used `e` for error below.
```suggestion
            .map(|m| Mnemonic { internal: m })
```

src/lib.rs
```diff
@@ -857,10 +857,33 @@ impl BumpFeeTxBuilder {
...
+            .map_err(|e| BdkError::Generic(e.to_string()))
+    }
+
+    fn from_entropy(entropy: Vec<u8>) -> Result<Self, BdkError> {
+        BdkMnemonic::from_entropy(entropy.as_slice())
+            .map(|x| Mnemonic { internal: x })
```
notmandatory, line 880: same comment as above:
```suggestion
            .map(|m| Mnemonic { internal: m })
```

notmandatory:
I think this one is ready to go. I tested it by running also a quick test in `test.kts` below and all works as expected. We don't need to make any permanent change to `test.kts` but this was helpful for me to confirm that everything is OK without the `Mutex`.

```kotlin
import org.bitcoindevkit.*

val network = Network.TESTNET

val mnemonic = Mnemonic(WordCount.WORDS12)

println("mnemonic = ${mnemonic.asString()}")

val descriptorSecKey = DescriptorSecretKey(network, mnemonic, null)

println("descriptorSecKey = ${descriptorSecKey.asString()}")

println("mnemonic = ${mnemonic.asString()}")
```

notmandatory:
Also since this is calling `bdk` code that's already tested I don't think you need to add any addition testing for these new bindings APIs.

src/lib.rs
```diff
@@ -857,10 +857,33 @@ impl BumpFeeTxBuilder {
...
 
-fn generate_mnemonic(word_count: WordCount) -> Result<String, BdkError> {
-    let mnemonic: GeneratedKey<_, BareCtx> =
-        Mnemonic::generate((word_count, Language::English)).unwrap();
-    Ok(mnemonic.to_string())
+struct Mnemonic {
+    internal: BdkMnemonic,
+}
+
+impl Mnemonic {
```
thunderbiscuit, line 864: We're starting to enforce that documentation be written for new PRs. In general it's not too hard because you can more or less use what is in the Rust docs if the constructs match.

src/lib.rs
```diff
@@ -857,10 +857,33 @@ impl BumpFeeTxBuilder {
...
+    fn new(word_count: WordCount) -> Self {
+        let generated_key: GeneratedKey<_, BareCtx> =
+            BdkMnemonic::generate((word_count, Language::English)).unwrap();
+        let mnemonic = BdkMnemonic::parse_in(Language::English, generated_key.to_string()).unwrap();
```
thunderbiscuit, line 868: I can't quite figure out if there is a better way to create this mnemonic but for some reason these 3 lines look and feel a bit weird/complex to me, when there should be some sort of `Mnemonic::create()` method that does it all in one go. Instead we have to use the `Mnemonic::generate` which actually returns a `GeneratedKey` _which has a mnemonic inside_, and then we feed that  mnemonic back to the `Mnemonic` struct?

But I'm looking at the docs and for some reason I can't find a better/cleaner way. But as is, doesn't it feel like we're not quite using the API as intended? Surely this is a very convoluted way to get at the idea of "creating a mnemonic".

thunderbiscuit (changes requested):

thunderbiscuit:
Just to be clear the "changes" I'm requesting are regarding the documentation. The mnemonic generation is a bit weird to me but I don't want to block this PR on it. Just hoping to start a discussion.

src/lib.rs
thunderbiscuit: I think this might be due to the fact that bdk does not turn on the `rand` feature on the bip39 crate. From the docs and some of their examples, I gather that the "standard" way to do what we're trying to do here would be to use the `Mnemonic::generate()` or `Mnemonic::generate_in()` methods ([docs here](https://docs.rs/bip39/1.0.1/src/bip39/lib.rs.html#275-295)). This is where it would just use the RNG and create the mnemonic directly for you. 

Going through the `GeneratedKey<_ BareCtx> struct is a workaround that I guess, but I'm not as clear as to what this type does/purpose.

dhruv-1001: I've found a workaround for this
It is to use the `Mnemonic::from_entropy_in()` ([docs here](https://docs.rs/bip39/1.0.1/src/bip39/lib.rs.html#180)).  And the method would look like this :
```rust
fn new(word_count: WordCount) -> Self {
    let mut entropy = [0u8; 32];
    thread_rng().fill(&mut entropy[..]);
    let entropy = &entropy.as_ref()[..(word_count as usize / 8)];
    let mnemonic = BdkMnemonic::from_entropy_in(Language::English, entropy).unwrap();
    Mnemonic { internal: mnemonic }
}
```

Here we cannot use `Mnemonic::generate_in_with()` ([docs here](https://docs.rs/bip39/1.0.1/src/bip39/lib.rs.html#245-261)) because `bip39::Mnemonic` only supports 12, 18 & 24 words. While BDK supports 12, 15, 18, 21 & 24 words. So we can implement it the same way it's done in BDK and have support for 12, 15, 18, 21 & 24 words.

notmandatory: Good research, even though it's a little convoluted it looks like the current approach is the only way to get what we need without copy/pasting code. I think you're right @thunderbiscuit we need to push for  bitcoindevkit/bdk#644 and make sure it provides a better API to do what we need here.

dhruv-1001: So do we continue as it is right now, and cleanup the API with different PR once [bitcoindevkit/bdk#644](https://github.com/bitcoindevkit/bdk/pull/644) gets merged?
Or should we wait for it to get merged?

thunderbiscuit: I suggest we go ahead with what you currently have. Add the docstrings and it's a go from me.

dhruv-1001: Updated the PR according to suggestions. Resolving Conversation

thunderbiscuit (approved):
ACK 9866649fdc5ffe0203f5a76082d789a22e5214bf.

notmandatory (approved):
ACK 9866649fdc5ffe0203f5a76082d789a22e5214bf

PR #228, opened 2022-11-10
Merge python ffi

thunderbiscuit:
This PR brings the bdk-python repo into the bdk-ffi repository.

This is part of https://github.com/bitcoindevkit/bdk-ffi/issues/206.

## Note to the reviewers
To simplify things, I have removed all files at the root but for the `.gitignore` (license, readme, etc.). I simply brought them in the bdk-ffi directory when I created it, to make sure everything was kept really simple and separate. I plan on bringing those back once bdk-android and bdk-jvm are in. In any case the readme will require a fairly big rewrite.


.github/workflows/build-python-wheels.yaml
```diff
@@ -0,0 +1,133 @@
+name: Build Python wheels
+on: [push, pull_request]
```
notmandatory, line 2: Don't need to do it in this PR.. but we should be able to tune the workflows to only run when bdk-ffi changes, or the there's a [push to a file](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-affects-specific-files) for a particular language sub-directory.
```suggestion
on: 
      push:
         paths:
            - 'bdk-ffi/**'
            - 'bdk-python/**'
      pull_request
```

.github/workflows/cont_integration.yml
```diff
@@ -1,7 +1,6 @@
+name: bdk-ffi CI
 on: [push, pull_request]
```
notmandatory, line 2: As above, but only need to run when bdk-ffi files change.
```suggestion
on:
    push:
             paths:
                - 'bdk-ffi/**'
          pull_request
             paths:
                - 'bdk-ffi/**'
```

thunderbiscuit: Oh those are fantastic changes, I figured I'd look into it after but it's too easy not to include!

notmandatory:
I added one more small change to the workflows to set the default working directory, then we don't need to do a `cd bdk-xxx` for each `run` step.

notmandatory (approved):
ACK 4f544e465bc8c1af13733e17af575de3115425ca

PR #230, opened 2022-11-14
bdk-kotlin ffi merge

thunderbiscuit:
Merge bdk-kotlin into bdk-ffi.


notmandatory:
I added a couple commits to move the bdk-ffi rust "workspace" to the top level of the `bdk-ffi` project, but still leaving all the rust code in the `bdk-ffi/bdk-ffi` directory. This fixes the audit workflow.

notmandatory:
> It's too bad that it leaves the repo just that little bit muddier (the target directory will show up at the root with the Cargo files, but the `src` is one deep into the bdk-ffi.

I agree, it's not ideal for anyone unfamiliar with how rust work spaces work.. but it's not terrible since all the submodules depend on /target and that's now one level less buried.

notmandatory (approved):
ACK 063f69222aa0087850d63fc186acb6eb8cbbb3ef

notmandatory:
I had to relax the signed commit rule since some of the old commits aren't signed. Will put it back on after this is merged.

PR #260, opened 2022-11-22
Add descriptor templates

### LINKED ISSUE #165: Create a type safe descriptor builder
As discussed in the team chat with kirillzhukov, the `bdk` lib uses rust macros to provide a type safe way to construct valid descriptors. Since the macro approach won't work for `bdk-ffi` language bindings we should try to build a simple library, possibly as a new repo, that uses simple structs and enums to accomplish something similar to the `bdk::descriptor::dsl` module.

thunderbiscuit:
Fixes #165.

Also relevant is bitcoindevkit/bdk#817

### Description
This PR adds the descriptor templates as described in #165.

### Links
[Docs for BIP84 template](https://docs.rs/bdk/latest/bdk/descriptor/template/struct.Bip84.html)

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -203,7 +203,7 @@ dictionary ScriptAmount {
 
 interface Wallet {
   [Throws=BdkError]
-  constructor(string descriptor, string? change_descriptor, Network network, DatabaseConfig database_config);
+  constructor(Descriptor descriptor, Descriptor? change_descriptor, Network network, DatabaseConfig database_config);
 
   [Throws=BdkError]
   AddressInfo get_address(AddressIndex address_index);
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -357,6 +357,33 @@ interface DescriptorPublicKey {
   string as_string();
 };
 
+interface Descriptor {
+  // [Throws=Error]
+  // constructor(string descriptor);
+
+  [Name=new_bip44]
+  constructor(DescriptorSecretKey secret_key, KeychainKind keychain, Network network);
+
+  [Name=new_bip44_public]
+  constructor(DescriptorPublicKey public_key, string fingerprint, KeychainKind keychain, Network network);
+
+  // [Name=new_bip49]
+  // constructor(DescriptorSecretKey secret_key, KeychainKind keychain);
+
+  // [Name=new_bip49_public]
+  // constructor(DescriptorPublicKey public_key, string fingerprint, KeychainKind keychain);
+
+  // [Name=new_bip84]
+  // constructor(DescriptorSecretKey secret_key, KeychainKind keychain);
+
+  // [Name=new_bip84_public]
+  // constructor(DescriptorPublicKey public_key, string fingerprint, KeychainKind keychain);
+
+  string as_string();
+
+  string as_string_private();
+};
+
 interface Address {
   [Throws=BdkError]
   constructor(string address);
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -367,17 +367,17 @@ interface Descriptor {
   [Name=new_bip44_public]
   constructor(DescriptorPublicKey public_key, string fingerprint, KeychainKind keychain, Network network);
 
-  // [Name=new_bip49]
-  // constructor(DescriptorSecretKey secret_key, KeychainKind keychain);
+  [Name=new_bip49]
+  constructor(DescriptorSecretKey secret_key, KeychainKind keychain, Network network);
 
-  // [Name=new_bip49_public]
-  // constructor(DescriptorPublicKey public_key, string fingerprint, KeychainKind keychain);
+  [Name=new_bip49_public]
+  constructor(DescriptorPublicKey public_key, string fingerprint, KeychainKind keychain, Network network);
 
-  // [Name=new_bip84]
-  // constructor(DescriptorSecretKey secret_key, KeychainKind keychain);
+  [Name=new_bip84]
+  constructor(DescriptorSecretKey secret_key, KeychainKind keychain, Network network);
 
-  // [Name=new_bip84_public]
-  // constructor(DescriptorPublicKey public_key, string fingerprint, KeychainKind keychain);
+  [Name=new_bip84_public]
+  constructor(DescriptorPublicKey public_key, string fingerprint, KeychainKind keychain, Network network);
 
   string as_string();
 
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -358,8 +358,8 @@ interface DescriptorPublicKey {
 };
 
 interface Descriptor {
-  // [Throws=Error]
-  // constructor(string descriptor);
+  [Throws=BdkError]
+  constructor(string descriptor, Network network);
 
   [Name=new_bip44]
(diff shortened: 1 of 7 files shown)


notmandatory:
API design looks good to me.

bdk-ffi/src/bdk.udl
```diff
@@ -335,3 +335,34 @@ interface Address {
...
+  [Name=new_bip84_public]
+  constructor(DescriptorPublicKey public_key, string? fingerprint, KeychainKind keychain);
+
+  [Name=new_bip86]
+  constructor(DescriptorSecretKey secret_key, KeychainKind keychain);
+
+  [Name=new_bip86_public]
+  constructor(DescriptorPublicKey public_key, string? fingerprint, KeychainKind keychain);
+
+  string to_string();
```
notmandatory, line 367: We had conflicts in Kotlin I think with `to_string()`, should be:
```suggestion
  string as_string();
```

thunderbiscuit:
This PR is surprisingly more complex to pull off than one might think at first glance. The core issue is that descriptors in BDK are built on fairly complex types and generics coming from rust-miniscript and rust-bitcoin.

Here are some notes I wrote for myself to keep track of the task at hand and the tradeoffs:

## Types related to extended keys and descriptors in bdk, rust-bitcoin, and miniscript:
```rust
// bdk::keys::SinglePriv                 (miniscript) <- Struct
// bdk::keys::SinglePub                  (miniscript) <- Struct
// bdk::keys::DescriptorSecretKey        (miniscript) <- Enum
// bdk::keys::DescriptorPublicKey        (miniscript) <- Enum
// bdk::descriptor::DescriptorXKey       (miniscript) <- Struct
// bdk::descriptor::Descriptor           (miniscript) <- Enum

// bdk::keys::ExtendedKey                (bdk)        <- Enum
// bdk::keys::DerivableKey               (bdk)        <- Trait
// bdk::descriptor::IntoWalletDescriptor (bdk)        <- Trait
// bdk::keys::DescriptorKey              (bdk)        <- Enum
// bdk::descriptor::ExtendedDescriptor   (bdk)        <- Alias for Descriptor<DescriptorPublicKey>
// bdk::descriptor::template::DescriptorTemplateOut   <- Alias for (ExtendedDescriptor, KeyMap, ValidNetworks)

// bitcoin::util::bip32::ExtendedPrivKey (bitcoin)    <- Struct
// bitcoin::util::bip32::ExtendedPubKey  (bitcoin)    <- Struct
```

## Bindings `Wallet` vs Rust BDK `Wallet`
In the bindings the `Wallet` currently takes descriptors in the form of strings whereas in Rust BDK, the main constructor is a generic function on the descriptor type with a trait bound of `IntoWalletDescriptor`:
and the `Wallet::new()` constructor requires a descriptor that implements the `IntoWalletDescriptor` trait:
```rust
impl Wallet {
    pub fn new<E: IntoWalletDescriptor>(
        descriptor: E,
        change_descriptor: Option<E>,
        // ...
    ) -> Result<Self, Error> {}
}
```

This trait is implemented on a few types:
1. `&str`
2. `&String`
3. `(ExtendedDescriptor, KeyMap)`
4. `DescriptorTemplateOut` -> that's just an alias for `(ExtendedDescriptor, KeyMap, ValidNetworks)`
5. `ExtendedDescriptor`    -> that's just an alias for `Descriptor<DescriptorPublicKey>`, so only works with xpubs

## Currently unsolved
The example `DescriptorPR260` type I added here showcases one way to do this: we could have a `Descriptor` type which would simply wrap a `DescriptorTemplateOut`. We could even change the argument type on the Wallet to take this `Descriptor` type if we wanted. The problem is that I don't see a "clean" way to retrieve the descriptor as a string _after_ we've used the template. The type that comes out of the `Template::build()` method is `(ExtendedDescriptor, KeyMap, ValidNetworks)`, segreggating the extended public key and the private keys. Not clear to me how you'd just "print the private descriptor" yet, although I'm still playing around with it.

thunderbiscuit:
Good progress was made today. BIP44, 49, and 84 "private" templates work. A few outstanding discussion items:

1. The fingerprint argument on the public version of the templates is mandatory, which means we'll need a way for users to get this fingerprint out of the `DescriptorSecretKey` struct if we want to keep the API the same.
2. I think we need 2 `as_string()` methods: one that returns the public version of the descriptor, and one that returns the private version if there is one
3. We can either ask users to always build a string out of the `Descriptor` type and give that to the wallet, or we could make the wallet take a descriptor of type `Descriptor`. This would be cleaner and nearer to the BDK API, but implies we need to have a `from_string()` constructor on the `Descriptor` type. All of this is a bit different than in BDK, where the descriptor provided to the Wallet struct are generics and simply have a trait bound of `IntoWalletDescriptor`.
4. The Bip86 template is not something in BDK yet. I assume the best way to go about this is to open an issue/PR on Rust BDK to add it upstream, and so for now I have simply omitted it.

thunderbiscuit:
If bitcoindevkit/bdk/pull/821 gets merged I assume we might be ok with adding the dev-dependency on assert_matches as well. It feels much cleaner to me, and I have prepared an extra commit that brings that in.

```rust
// asserting on the string
assert_eq!(wallet2.unwrap_err().to_string(), "Descriptor(Key(InvalidNetwork))");

// assert_matches!
assert_matches!(wallet2.unwrap_err(), bdk::Error::Descriptor(Key(InvalidNetwork)))
```

bdk-ffi/src/lib.rs
```diff
@@ -1419,4 +1612,142 @@ mod test {
         assert!(tx_builder_result.psbt.fee_amount().is_some());
         assert_eq!(tx_builder_result.psbt.fee_amount().unwrap(), 220);
     }
+
+    #[test]
+    fn test_descriptor_templates() {
```
notmandatory, line 1617: Need to remove or at least comment out the `println!` lines and add back in the `assert!` lines at the end.

notmandatory: Oh forgot about the issue with bitcoindevkit/bdk#817, ok to leave the asserts commented out until we update to bdk 0.26 which should fix the issue.

bdk-ffi/src/lib.rs
```diff
@@ -1419,4 +1612,142 @@ mod test {
...
+    }
+
+    #[test]
+    fn test_descriptor_from_string() {
+        let descriptor1 = Descriptor::new("wpkh(tprv8hwWMmPE4BVNxGdVt3HhEERZhondQvodUY7Ajyseyhudr4WabJqWKWLr4Wi2r26CDaNCQhhxEftEaNzz7dPGhWuKFU4VULesmhEfZYyBXdE/0/*)".to_string(), Network::Testnet);
+        let descriptor2 = Descriptor::new("wpkh(tprv8hwWMmPE4BVNxGdVt3HhEERZhondQvodUY7Ajyseyhudr4WabJqWKWLr4Wi2r26CDaNCQhhxEftEaNzz7dPGhWuKFU4VULesmhEfZYyBXdE/0/*)".to_string(), Network::Bitcoin);
+
+        // Creating a Descriptor using an extended key that doesn't match the network provided will throw and InvalidNetwork Error
+        assert!(descriptor1.is_ok());
+        assert_eq!(
```
notmandatory, line 1722: Shouldn't this be updated to use the new `assert_match!` macro?

notmandatory (approved):
ACK bada9b82e08dd97842facc391fec5f43b6591142

PR #264, opened 2022-11-25
Add uniffi-bindings-template link to readme

thunderbiscuit:
This PR adds a small paragraph pointing to the uniffi-bindings-template repo as an example and starting point for other projects wishing to leverage a similar tech stack.


notmandatory (approved):
ACK 832387d32d540b0f0ffa9f833a24ca28da2d4735

PR #269, opened 2022-11-29
Fix artifact upload path in JVM CI publishing workflow

thunderbiscuit:
The JVM publishing CI workflow was uploading artifacts from an incorrect path. This PR fixes that.


notmandatory (approved):
ACK f834da11c57ca0da37ac7f2c2780f7a00dabd517

PR #271, opened 2022-11-30
Fix jvm and android library version defined in two places

### LINKED ISSUE #234: Group and version properties are defined in three places
Once in the root build.gradle.kts file (which is required by the nexus publish-plugin and once in each of the publishing blocks in the build files of the individual libraries.

I played around with the delegated properties and the Gradle extra properties but could not find a solution I found elegant and satisfactory. Opening this issue to keep track of the problem.

thunderbiscuit:
This PR fixes #234.

The library version for bdk-android and bdk-jvm is now defined in the `gradle.properties` file, and picked up in the gradle build scripts using property delegates.

This PR also adds documentation on how to use the snapshot versions of bdk-jvm and bdk-android.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,8 @@ dependencies {
 }
 ```
 
+_Note:_ We also publish snapshot versions of bdk-jvm and bdk-android. See the specific readmes for instructions on how to use those.
+
 ### bdk-python
 ```shell
 pip3 install bdkpython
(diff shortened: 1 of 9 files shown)


notmandatory (approved):
ACK 5dd5828e8893615336f344f3077f7de110e4b78b

PR #279, opened 2022-12-07
Change TxBuilder.drain_to argument to Script instead of address String

### LINKED ISSUE #261: Investigate `to_script_pubkey(address)`
We have this method which we use in the TxBuilder, but now that we have the `Script` type I'm not sure if we need this method anymore or if the TxBuilder should be refactored.  Just noticed it while working on something else so I'm opening this now to keep it in mind.

notmandatory:
### Description

Change `TxBuilder.drain_to()` argument to Script instead of address String. Also remove unneeded `fn to_script_pubkey()` fixes #261.

### Notes to the reviewers

This change aligns bdk-ffi with bdk API which also takes a script instead of an address string for the `TxBuilder.drain_to()`. 

### Changelog notice

* APIs Changed
  * `TxBuilder.drain_to()` argument is now `Script` instead of address `String`.

#### All Submissions:

### DIFF
--- a/api-docs/kotlin/src/main/kotlin/org/bitcoindevkit/bdk.kt
+++ b/api-docs/kotlin/src/main/kotlin/org/bitcoindevkit/bdk.kt
@@ -410,7 +410,7 @@ class TxBuilder() {
      * [drainWallet] to spend all of them. When bumping the fees of a transaction made with this option,
      * you probably want to use [BumpFeeTxBuilder.allowShrinking] to allow this output to be reduced to pay for the extra fees.
      */
-    fun drainTo(address: String): TxBuilder {}
+    fun drainTo(script: Script): TxBuilder {}
 
     /** Enable signaling RBF. This will use the default `nsequence` value of `0xFFFFFFFD`. */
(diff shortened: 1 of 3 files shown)


thunderbiscuit (approved):
ACK [7d1a450](https://github.com/bitcoindevkit/bdk-ffi/pull/279/commits/7d1a4500ef88d0d817af2abfbe4e77007ad70a8e).

How do you feel about fixing up the API docs as we go? Would you prefer we keep all the fixes for one big PR fix every release, or require people update as they open PRs that change the API?

notmandatory:
> How do you feel about fixing up the API docs as we go? Would you prefer we keep all the fixes for one big PR fix every release, or require people update as they open PRs that change the API?

Yes good catch, we should keep the api-docs up to date as we make PR changes. I don't think for this one I only needed to change the method signature for Kotlin since the docs already match the rust docs.  Doese it needs updating anywhere else?

thunderbiscuit:
Yep this is good to go. Ok will keep the docs in mind for future PRs!

PR #281, opened 2022-12-14
Compile libraries before tests

### LINKED ISSUE #236: Build JVM and Android libs tasks do not run as part of the standard build process
### Description

Currently, when running tests directly from CLI or an IDE, we don't execute `buildJvmLib` and `buildAndroidLib` tasks automatically. Usually codegen-like tasks similar to `buildJvmLib` and `buildAndroidLib` are executed automatically before `assemble` task is executed, it would be nice to set that up for this project as well.

Pros:
1. No need to manually run `buildJvmLib` or `buildAndroidLib` when building project (which happens as part of running tests for example)
2. Easier to develop when making changes to bindings as they changes will be picked up automatically which avoids manually running `buildJvmLib` and `buildAndroidLib`.

Cons:
1. I didn't get a chance to test that out but it would be good that `buildJvmLib` and `buildAndroidLib` tasks can be cachable to avoid unnecessary execution, since we would be trying to run the task on each build.

### How to Reproduce

Steps:
1. Delete generated bindings in `jvm/src/main/kotlin/org/bitcoindevkit`
2. Run `./gradlew :jvm:test`
3. Fails since native bindings were not auto generated, requires manually running `buildJvmLib`
4. Same for the `:android` project

### Expected Result

The custom `buildJvmLib` and `buildAndroidLib` tasks are executed automatically when assembling `:jvm` and `:android` projects.

### Version

Latest `0.8.0-SNAPSHOT`.

### Artifact

- [X] bdk-jvm
- [X] bdk-android

### Platform

- [ ] ARM64 Android
- [ ] 64-bit x86 Android
- [ ] 32-bit x86 Android
- [ ] 64-bit x86 Linux (kernel 2.6.32+, glibc 2.11+)
- [ ] 64-bit x86 macOS (10.7+, Lion+)
- [ ] ARM64 macOS (11.0+, Big Sur+)

### Relevant log output

_No response_

thunderbiscuit:
### Description
This PR fixes #236 by making the `KotlinCompile` task depend on the `buildJvmLib` and `buildAndroidLib`. Thanks @kirillzh for the suggestion!

The tests can now be run directly without requiring the building of the libraries first.

#### All Submissions:

### DIFF
--- a/.github/workflows/test-android.yaml
+++ b/.github/workflows/test-android.yaml
@@ -46,11 +46,6 @@ jobs:
       - name: Install rust android targets
         run: rustup target add x86_64-linux-android aarch64-linux-android armv7-linux-androideabi
 
-      - name: Build bdk-android library
-        run: |
-          cd bdk-android
-          ./gradlew buildAndroidLib
-
       - name: Run Android tests
         run: |
           cd bdk-android
--- a/.github/workflows/test-jvm.yaml
+++ b/.github/workflows/test-jvm.yaml
@@ -31,11 +31,6 @@ jobs:
           distribution: temurin
           java-version: 11
 
-      - name: Build bdk-jvm library
-        run: |
-          cd bdk-jvm
-          ./gradlew buildJvmLib
-
       - name: Run JVM tests
         run: |
           cd bdk-jvm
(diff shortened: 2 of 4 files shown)


notmandatory (approved):
ACK 8193cc56b9ba1d6f0b2526c7de56f013cf7504f9

Issue #242, opened 2022-08-17
Streamline the NDK version used in Android build and CI


thunderbiscuit:
When using the ubuntu-latest (ubuntu-20.04), the CI has also migrated to NDK version 25 (see issue [here](https://github.com/actions/runner-images/issues/5930)).

We should try to use a more recent version of NDK.

thunderbiscuit:
Updating to the latest NDK is a long-standing issue with uniffi-rs and Rust in general. Here are some links to keep track of some of the conversation and progress around this:

1. https://github.com/mozilla/application-services/pull/5442
2. https://github.com/mozilla/application-services/issues/5436
3. https://github.com/mozilla/application-services/issues/5165
4. https://github.com/mozilla/rust-android-gradle/issues/75
5. https://github.com/rust-lang/rust/pull/85806
6. https://github.com/rust-lang/rust/pull/102332

thunderbiscuit:
Currently updating to the latest NDK works only for some builds:

The x86_64 version of bdk-android throws a runtime error when used from an x86_64 machine (but not from an M1).
```
Native library (android-x86-64/libbdkffi.so) not found in resource path (.)
```

PR #282, opened 2022-12-15
Use emulator to run connected tests for bdk-android in CI

### LINKED ISSUE #243: Fix Android instrumentation tests in CI
The CI should run the instrumented tests.

thunderbiscuit:
This PR fixes the test-android workflow to use an emulator and run the connected tests for bdk-android. The workflow as it stands does not actually run any tests!

Fixes #243.

### DIFF
--- a/.github/workflows/test-android.yaml
+++ b/.github/workflows/test-android.yaml
@@ -10,11 +10,11 @@ on:
       - "bdk-ffi/**"
       - "bdk-android/**"
 
-# The default Android NDK on the ubuntu-22.04 image is 25.2.9519653
+# The default Android NDK on the macos-12 image is 25.2.9519653
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: macos-12
     steps:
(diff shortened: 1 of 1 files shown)


notmandatory:
Is this one still WIP for release 0.27 ? I think it is breaking because our Kotlin builds need to use 21.4.X not the default 25.x.

thunderbiscuit:
Useful links:
1. https://github.com/bbqsrc/cargo-ndk/issues/76
2. https://github.com/termux/termux-packages/discussions/8214

The problem _might_ be coming from the plugin, where we define the following:
```kotlin
// if ANDROID_NDK_ROOT is not set then set it to github actions default
if (System.getenv("ANDROID_NDK_ROOT") == null) {
    environment(
        Pair("ANDROID_NDK_ROOT", "${System.getenv("ANDROID_SDK_ROOT")}/ndk-bundle")
    )
}
```

thunderbiscuit:
See #242 for related links.

PR #290, opened 2023-01-10
Prepare libraries for 0.26.0 release

thunderbiscuit:
### Description
This PR reflects the changes required to bump all libraries to their new release version.

I ran all tests and looked at examples, as well as updated the API docs for Kotlin, which did not include docs and samples for the new descriptor templates (that was my fault they should have been in PR #260 😅).

### Notes to the reviewers
It's a lot of small stuff all over the place, let me know if I forgot anything.

The reason the tests passed in #288 is that we don't run the Python tests in the CI, nor the Swift ones, nor the Android ones! cc #291 and #282.

### Changelog notice
There is not changelog notice for this PR.

#### All Submissions:

#### New Features:

### DIFF
--- a/api-docs/kotlin/src/main/kotlin/org/bitcoindevkit/bdk.kt
+++ b/api-docs/kotlin/src/main/kotlin/org/bitcoindevkit/bdk.kt
@@ -350,15 +350,15 @@ data class BlockTime (
  * @constructor Create a BDK wallet.
  *
  * @param descriptor The main (or "external") descriptor.
- * @param changeDescriptor The change (or "internal") descriptor.
+ * @param changeDescriptor? The change (or "internal") descriptor.
  * @param network The network to act on.
  * @param databaseConfig The database configuration.
  *
  * @sample org.bitcoindevkit.walletSample
  */
 class Wallet(
-    descriptor: String,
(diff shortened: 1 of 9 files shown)


api-docs/kotlin/src/main/kotlin/org/bitcoindevkit/bdk.kt
```diff
@@ -578,6 +578,62 @@ class DescriptorPublicKey(network: Network, mnemonic: String, password: String?)
     fun asString(): String
 }
 
+/**
+ * A output descriptor.
+ *
+ * @param descriptor The descriptor in string format.
+ * @param network The network this DescriptorPublicKey is to be used on.
```
notmandatory, line 585: I think you may have a copy/paste error here, should be Descriptor right?
```suggestion
 * @param network The network this Descriptor is to be used on.
```

thunderbiscuit: Oh yep! Thanks for pointing it out.

thunderbiscuit: Fixed.

PR #298, opened 2023-01-23
Fix Python release workflow

thunderbiscuit:
This PR fixes the Python release workflows for macOS and Windows.

## Notes for reviewers
I simply added the absolute path to the correct directories so that the upload task would find the tarballs, but I wonder if I should instead work with relative paths and the `with:` key. Happy to revise if you think there is significant advantages to this approach.

### DIFF
--- a/.github/workflows/build-python-wheels.yaml
+++ b/.github/workflows/build-python-wheels.yaml
@@ -90,7 +90,7 @@ jobs:
       - uses: actions/upload-artifact@v2
         with:
           name: bdkpython-macos-${{ matrix.python }}
-          path: dist/*.whl
+          path: /Users/runner/work/bdk-ffi/bdk-ffi/bdk-python/dist/*.whl
 
   build-windows-wheel:
     name: 'Build windows wheel'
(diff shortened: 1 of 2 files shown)


.github/workflows/build-python-wheels.yaml
```diff
@@ -87,10 +87,12 @@ jobs:
         env:
           ARCHFLAGS: "-arch x86_64 -arch arm64"
         run: python3 setup.py -v bdist_wheel
+      - name: Display structure of macOS files
```
notmandatory, line 90: Did you mean to keep this in the workflow?

.github/workflows/build-python-wheels.yaml
```diff
@@ -120,7 +122,9 @@ jobs:
       - run: pip install --user wheel
       - name: build wheel
         run: python setup.py -v bdist_wheel
+      - name: Display structure of Windows files
```
notmandatory, line 125: Also this one, looks like it's here for debugging

notmandatory (approved):
ACK 06b0f5e1559a652d8fd272183984dd1360ed6990

Looks good to go with or without the steps I suggested removing.

.github/workflows/build-python-wheels.yaml
thunderbiscuit: Oh yeah I'll remove those actually.

PR #304, opened 2023-01-31
Expose/get internal address

xsats:
### Description

Until this PR,  bdk-ffi get_address did not support internal/change addresses. Any sufficiently advanced wallet will need this functionality, or be indistinguishable from an insufficiently-advanced wallet.

### Notes to the reviewers

I created a separate mapping for get_internal_address instead of adding another field to get_address, which matches the existing bdk naming pattern and keeps the exposed get_{x_}address methods from being cluttered.

### Changelog notice

Add `wallet::get_internal_addresses` API

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/wallet.rs
+++ b/bdk-ffi/src/wallet.rs
@@ -577,6 +577,14 @@ mod test {
 
         assert_eq!(
             wallet
+                .get_address(crate::AddressIndex::LastUnused)
+                .unwrap()
+                .address,
+            "bcrt1q0xs7dau8af22rspp4klya4f7lhggcnqfun2y3a"
+        );
+
+        assert_eq!(
+            wallet
                 .get_internal_address(crate::AddressIndex::New)
                 .unwrap()
                 .address,
@@ -589,6 +597,14 @@ mod test {
                 .unwrap()
                 .address,
             "bcrt1qaux734vuhykww9632v8cmdnk7z2mw5lsf74v6k"
+        );
+
+        assert_eq!(
+            wallet
+                .get_internal_address(crate::AddressIndex::LastUnused)
+                .unwrap()
+                .address,
+            "bcrt1qpmz73cyx00r4a5dea469j40ax6d6kqyd67nnpj"
         );
     }
 }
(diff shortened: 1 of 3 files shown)


thunderbiscuit (changes requested):
Approach ACK. Great addition to the API. It reminded me that I had in mind of taking all methods on the Wallet type one by one and asking myself "is there a reason to keep this out of the bindings?". In this case clearly there are none. Happy to see it added. 

Similar to 305, you'll need to remove the `to_string()` calls in the test to pass the Clippy CI, and I'd like to have the Kotlin API docs fixed before merging. As mentioned in #305, I'm happy to do it, just let me know.

xsats:
Thanks again, happy to fix this too. Also happy help adding other absent APIs. 

From my standpoint this and #305 are definite candidates, and simple to implement so I jumped straight to PRs. But will open issues for any 'contentious' additions in terms of whether they're needed, or there're multiple clear implementation variants etc.

xsats:
> and I'd like to have the Kotlin API docs fixed before merging. As mentioned in #305, I'm happy to do it, just let me know.

Similarly, if you don't mind adding for now that would be great thanks.

bdk-ffi/src/wallet.rs
```diff
@@ -522,4 +540,71 @@ mod test {
...
+                .address,
+            "bcrt1qaux734vuhykww9632v8cmdnk7z2mw5lsf74v6k"
+        );
+
+        assert_eq!(
+            wallet
+                .get_internal_address(crate::AddressIndex::LastUnused)
+                .unwrap()
+                .address,
+            "bcrt1qpmz73cyx00r4a5dea469j40ax6d6kqyd67nnpj"
```
thunderbiscuit, line 607: The tests are failing because the address here should be `bcrt1qaux734vuhykww9632v8cmdnk7z2mw5lsf74v6k`. Makes sense since it's not being used in between the two assertions, and so `LastUnused` should return the same address as the previous call on 596

thunderbiscuit (changes requested):
One small fix to the tests is required.

bdk-ffi/src/wallet.rs
xsats: You're absolutely right. Sorry, I'll try to check the tests more closely going forwards.

thunderbiscuit (approved):
ACK aa842c3844fa14c646fa616b9febe5c62db28635.

notmandatory (approved):
ACK aa842c3844fa14c646fa616b9febe5c62db28635

Concept and code looks good!

Issue #300, opened 2023-01-26
Migration to the procedural macros system of uniffi-rs


thunderbiscuit:
I'm opening this issue to keep track of documentation, references, and discussions around the migration to the new uniffi-rs system.

1. Documentation page on using proc-macros: [here](https://mozilla.github.io/uniffi-rs/proc_macro/index.html)
2. Issue on readiness of migration and further documentation: [here](https://github.com/mozilla/uniffi-rs/issues/1459)
3. PR for the `0.23.0` migration: #299

aki-mizu:
It seems rust-nostr already successfully migrated their ffi part to proc-macro
https://github.com/rust-nostr/nostr/pull/204

reez:
Thunder how interested are you in doing this at the current moment, just wondering what you're current thoughts would be for this

thunderbiscuit:
Looking into it now. I'm not sure if we'd be able to migrate everything or if some things would stay in the UDL and some would be proc macros.

But the one prerequisite to this is that we use the `--library` flag to create the library, which [is recommended anyway](https://mozilla.github.io/uniffi-rs/tutorial/foreign_language_bindings.html#running-uniffi-bindgen-using-a-library-file-recommended) and is a good move forward no matter what if we can do it, so I think we can start with that and see where it leads us.

reez:
awesome, thanks for those thoughts, and --library makes sense to start like you're saying

reez:
I'm going to close this since we started this migration in #677 and #678 . Maybe we can start a new checklist of everything that still needs to be migrated

PR #283, opened 2022-12-15
Add MVP Python API docs

thunderbiscuit:
Not ready, but the core machinery is there. I just don't really know how to make them look really "Pythonesque" yet.

bdkpython docs

Issue #251, opened 2022-10-19
Add Swift Docs


notmandatory:
https://developer.apple.com/documentation/Xcode/documenting-apps-frameworks-and-packages

Publish generated docs on bitcoindevkit.org website.

notmandatory:
https://www.swift.org/documentation/docc

thunderbiscuit:
My feeling is that we're better off waiting on proper uniffi support for inline docs. We're currently maintaining the Kotlin ones at not-too-great a cost, but adding Python and Swift feels like extra work I don't know if we have capacity for at the moment. What do you think @notmandatory @reez? 

I suggest closing this but open to other approaches if you're getting enough pressure to support Swift docs and think we should take this on.

reez:
> My feeling is that we're better off waiting on proper uniffi support for inline docs. We're currently maintaining the Kotlin ones at not-too-great a cost (now that we have the whole workflow and setup dealt with), but adding Python and Swift feels like extra work I don't know if we have capacity for at the moment, particularly because the extra gains we get from it are small (the Swift/Python syntax is slightly different thank Kotlin but the APi is exaclty the same). What do you think @notmandatory @reez?
> 
> I suggest closing this but open to other approaches if you're getting enough pressure to support Swift docs and think we should take this on.

That makes sense, I defer to ya'll on this since you have first hand experience on what it's been like maintaining the Kotlin docs, etc.

thunderbiscuit:
Well to be fair, once we got the pipeline working for Kotlin it's been ok. You write a few lines every time you PR, basically copying what's in Rust BDK. Not the worst. But we looked into the Swift API docs and we just didn't have the expertise to tell if it was going to be easy to create/deploy or just another rabbit hole not worth our time...

If you're super familiar with Swift API docs and you think it's only a small amount of work I do know that people would love to see them; I just don't know where that lands on your priority list. How about I add it to the discussion for the dev call next week? We can see where people are then and make the call. Uniffi has also been looking at merging one or two PRs that would make this possible directly from the Rust docs, so that would make it much easier, but I don't have a clear timeline for that feature to be merged/usable.

PR #305, opened 2023-01-31
Features/getaddress peek + reset

xsats:
### Description

The `get_address` 'Peek' `AddressIndex` option is useful to obtain specific addresses by their indices, without affecting the internal wallet database.

### Notes to the reviewers

Matched bdk patterns.

### Changelog notice

Add `AddressIndex::Peek(index)` and `get_address(Peek(index))` API

#### All Submissions:

#### New Features:

### DIFF
--- a/api-docs/kotlin/src/test/kotlin/org/bitcoindevkit/Samples.kt
+++ b/api-docs/kotlin/src/test/kotlin/org/bitcoindevkit/Samples.kt
@@ -72,10 +72,6 @@ fun addressIndexSample() {
         databaseConfig = DatabaseConfig.Memory
     )
 
-    fun getLastUnusedAddress(): AddressInfo {
-        return wallet.getAddress(AddressIndex.LastUnused)
-    }
-
     fun peekAddress100(): AddressInfo {
         return wallet.getAddress(AddressIndex.Peek(100u))
     }
@@ -89,11 +85,7 @@ fun addressInfoSample() {
         databaseConfig = DatabaseConfig.Memory
     )
 
-    fun getLastUnusedAddress(): AddressInfo {
-        return wallet.getAddress(AddressIndex.New)
-    }
-
-    val newAddress: AddressInfo = getLastUnusedAddress()
+    val newAddress: AddressInfo = wallet.getAddress(AddressIndex.New)
 
     println("New address at index ${newAddress.index} is ${newAddress.address}")
 }
--- a/api-docs/kotlin/src/test/kotlin/org/bitcoindevkit/Samples.kt
+++ b/api-docs/kotlin/src/test/kotlin/org/bitcoindevkit/Samples.kt
@@ -73,7 +73,11 @@ fun addressIndexSample() {
     )
 
     fun getLastUnusedAddress(): AddressInfo {
-        return wallet.getAddress(AddressIndex.LAST_UNUSED)
+        return wallet.getAddress(AddressIndex.LastUnused)
+    }
+
+    fun peekAddress100(): AddressInfo {
+        return wallet.getAddress(AddressIndex.Peek(100u))
     }
 }
 
@@ -86,7 +90,7 @@ fun addressInfoSample() {
     )
 
     fun getLastUnusedAddress(): AddressInfo {
-        return wallet.getAddress(AddressIndex.NEW)
+        return wallet.getAddress(AddressIndex.New)
     }
 
     val newAddress: AddressInfo = getLastUnusedAddress()
--- a/bdk-ffi/src/wallet.rs
+++ b/bdk-ffi/src/wallet.rs
@@ -575,6 +575,7 @@ mod test {
             "bcrt1q0xs7dau8af22rspp4klya4f7lhggcnqfun2y3a"
         );
 
+        // new index still 0
         assert_eq!(
             wallet
                 .get_address(crate::AddressIndex::New)
@@ -583,6 +584,7 @@ mod test {
             "bcrt1qqjn9gky9mkrm3c28e5e87t5akd3twg6xezp0tv"
         );
 
+        // new index now 1
         assert_eq!(
             wallet
                 .get_address(crate::AddressIndex::New)
@@ -591,12 +593,22 @@ mod test {
             "bcrt1q0xs7dau8af22rspp4klya4f7lhggcnqfun2y3a"
         );
 
+        // new index now 2
+        assert_eq!(
+            wallet
+                .get_address(crate::AddressIndex::New)
+                .unwrap()
+                .address,
+            "bcrt1q5g0mq6dkmwzvxscqwgc932jhgcxuqqkjv09tkj"
+        );
+
+        // peek index 1
         assert_eq!(
             wallet
                 .get_address(AddressIndex::Peek { index: 0 })
                 .unwrap()
                 .address,
-            "bcrt1qqjn9gky9mkrm3c28e5e87t5akd3twg6xezp0tv"
+            "bcrt1q0xs7dau8af22rspp4klya4f7lhggcnqfun2y3a"
         );
     }
 
--- a/bdk-ffi/src/wallet.rs
+++ b/bdk-ffi/src/wallet.rs
@@ -542,7 +542,7 @@ mod test {
     }
 
     #[test]
-    fn test_peek_address() {
+    fn test_peek_reset_address() {
         let test_wpkh = "wpkh(tprv8hwWMmPE4BVNxGdVt3HhEERZhondQvodUY7Ajyseyhudr4WabJqWKWLr4Wi2r26CDaNCQhhxEftEaNzz7dPGhWuKFU4VULesmhEfZYyBXdE/0/*)";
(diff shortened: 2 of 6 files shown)


thunderbiscuit (changes requested):
Thanks for the contribution! The `Peek` variant is definitely useful and I think we somehow just forgot about it.

I have added 1 small comment below, but overall this looks like a simple and useful feature to add. Let's try to get it in the next release. 4 little things while you're at it: 
1. Would you mind exposing the `Reset(u32)` (the last variant on the `AddressIndex` enum) as part of this PR? Might as well have the whole thing available.
2. The last thing left would be to add this new API to the [Kotlin API docs](https://github.com/bitcoindevkit/bdk-ffi/blob/master/api-docs/kotlin/src/main/kotlin/org/bitcoindevkit/bdk.kt). It's a bit of a weird file and temporary setup for our API docs and I don't know if you are familiar with Kotlin or not, so let me know if you prefer I just add that commit to the PR myself I'm happy to do it.
3. The JVM tests are failing because in Kotlin the new types are now sealed classes instead of enums, and so their capitalization changes. [Line 49 of the tests file](https://github.com/bitcoindevkit/bdk-ffi/blob/master/bdk-jvm/lib/src/test/kotlin/org/bitcoindevkit/JvmLibTest.kt) should read
```kotlin
val address = wallet.getAddress(AddressIndex.New).address
```
4. For some reason clippy in the CI for `1.63.0` is failing/warning [at the `.to_string()` method on the address in the tests](https://github.com/bitcoindevkit/bdk-ffi/actions/runs/4058927597/jobs/6988183685). I don't get this warning locally myself (I'm on 1.64.0) but it looks like you'll need to remove that redundant `to_string()` call to pass the clippy CI.

xsats:
Thanks a lot for your review and detailed suggestions! I'll fix the PR later today.

xsats:
1, 3, 4 should hopefully be resolved now.

> 2. The last thing left would be to add this new API to the [Kotlin API docs](https://github.com/bitcoindevkit/bdk-ffi/blob/master/api-docs/kotlin/src/main/kotlin/org/bitcoindevkit/bdk.kt). It's a bit of a weird file and temporary setup for our API docs and I don't know if you are familiar with Kotlin or not, so let me know if you prefer I just add that commit to the PR myself I'm happy to do it.

I'm reasonably familiar with Kotlin but not 100% sure how this doc generation workflow functions so if you don't mind adding that commit so I can see the diff for any future changes that would be great, thanks.

bdk-ffi/src/wallet.rs
```diff
@@ -522,4 +524,90 @@ mod test {
...
+                .get_address(crate::AddressIndex::New)
+                .unwrap()
+                .address,
+            "bcrt1q5g0mq6dkmwzvxscqwgc932jhgcxuqqkjv09tkj"
+        );
+
+        // peek index 1
+        assert_eq!(
+            wallet
+                .get_address(AddressIndex::Peek { index: 0 })
```
thunderbiscuit, line 592: Should be index `1` here to pass the test.

thunderbiscuit (changes requested):
One small test issue but otherwise looks good to me.

xsats:
> Since you're a new contributor to the bindings, I'd like to take the opportunity to ask you some questions about our repo/setup/docs if you don't mind. Feel free to not answer them if you prefer not to, or send me a DM on Discord/Twitter if you prefer to respond that way.
> 
> 1. How did you find the documentation for the bindings? Were you able to build locally on first try or did you struggle with it for a while? Let us know if you would add anything to the documentation or if anything was more confusing than helping.

I was able to build locally on first try. I think the repo/docs are fairly self explanatory, though I'm not 100% certain about the jvm doc generation workflow (as discussed above), although should be able to work it out if I look closer at past issues/commits. Some more direction on this in the docs could be helpful, but I appreciate you mentioned this is a temporary workaround.

> 2. Were you able to add the APIs in [Expose/get internal address #304](https://github.com/bitcoindevkit/bdk-ffi/pull/304) and [Features/getaddress peek #305](https://github.com/bitcoindevkit/bdk-ffi/pull/305) easily and test them locally in a project?

Yep, these API additions are quite simple and straightforward, and found there to be plenty of guidance in the existing code/previous PRs.

> 3. What library are you using directly yourself? Swift, Kotlin/JVM, Kotlin/Android, or Python?

Mainly using the Swift and Kotlin/Android at the moment.

> 4. Did you easily find our supporting samples and docs, or are they hidden too deep in the readme? (i.e. the [Devkit Wallet](https://github.com/thunderbiscuit/devkit-wallet), the [Kotlin API docs](https://bitcoindevkit.org/android/), the [BdkSwiftSample](https://github.com/futurepaul/BdkSwiftSample), the [Bindings Notes for Devs](https://hackmd.io/@thunderbiscuit/languagebindingsnotesfordevs)?

I had previously found and had a look over the Rust/Kotlin docs and some of the examples on the [BDK website](https://bitcoindevkit.org) but I suppose it wouldn't be a bad idea to add the examples, maybe as another column in the [supported languages/platforms table](https://github.com/bitcoindevkit/bdk-ffi/blob/master/README.md#supported-target-languages-and-platforms).

Are there any plans to add a Swift API reference to the BDK site?

thunderbiscuit (approved):
ACK f1c522496280bd82d1e57dc3fff2e3086967d277.

thunderbiscuit:
>Are there any plans to add a Swift API reference to the BDK site?
This is something we've been asked for for a while and something we'd love to offer, but at the moment we're not sure it's worth the effort. You can see discussions on this in a few places: 
1. #251
2. The closing of PR #283
3. #300 

The gist of it is that they require manual upkeep and are error-prone, plus the Kotlin API is identical except for the small syntactic differences between the languages, and so at the moment we're putting our hopes and efforts into the new uniffi-rs system that might allow us to have direct, clean inline docs in the future.

xsats:
Makes sense! As you say, the Kotlin API is functionally identical so definitely not a priority. Thanks

api-docs/kotlin/src/test/kotlin/org/bitcoindevkit/Samples.kt
```diff
@@ -86,7 +90,7 @@ fun addressInfoSample() {
     )
 
     fun getLastUnusedAddress(): AddressInfo {
```
notmandatory, line 92: Looks like a copy/paste error, should be renamed here and on line 96.
```suggestion
    fun getNewAddress(): AddressInfo {
```

notmandatory:
Looks like a good change. One suggestion, since `BdkAddressIndex` and the `AddressIndex` in this lib are now identical I think you can remove the bdk-ffi `AddressIndex` and just use the bdk `AddressIndex`.

api-docs/kotlin/src/test/kotlin/org/bitcoindevkit/Samples.kt
thunderbiscuit: Yeah thanks for pointing this out I found a few things in there I didn't like so I ended up simplifying the sample a bit. 👍

xsats:
> Looks like a good change. One suggestion, since `BdkAddressIndex` and the `AddressIndex` in this lib are now identical I think you can remove the bdk-ffi `AddressIndex` and just use the bdk `AddressIndex`.

The bindings generated by uniffi seem to demand struct variant enums, as opposed to tuple variant. This is why I implemented the `Peek(u32)` and `Reset(u32)` as struct types in the bdk-ffi `AddressIndex` enum.

I'm not too familiar with uniffi, but is there a way to force it to generate tuple variant bindings? In that case we can just use the bdk `AddressIndex` as you suggested.

In case it's helpful, here's an example of the error when attempting to compile using the bdk tuple variant enum:

```
   Compiling bdk-ffi v0.26.0 (/Users/edball/code/bdk/bdk-ffi/bdk-ffi)
error[E0769]: tuple variant `AddressIndex::Peek` written as struct variant
   --> ~/code/bdk/bdk-ffi/target/debug/build/bdk-ffi-5c24d076d38b6cb9/out/bdk.uniffi.rs:321:13
    |
321 |             r#AddressIndex::r#Peek { r#index, } => {
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
help: use the tuple variant pattern syntax instead
    |
321 |             r#AddressIndex::r#Peek(r#index) => {
    |                                   ~~~~~~~~~
```

thunderbiscuit:
Darn that's too bad. For a moment there when I read @notmandatory's comment I also thought it would just work right out of the gate. But yes reading their docs you can see on the [last line of this page](https://mozilla.github.io/uniffi-rs/udl/enumerations.html#enumerations): `Only enums with named fields are supported by this syntax.` So the tuple variants mean that the whole enum does require the translation between the BDK lib and the ffi layer.

notmandatory:
@xsats thanks for checking, I forgot about the only-named fields issue with uniffi enums, so it's good as you have it.

notmandatory (approved):
ACK f2fc586807265e0d257d822d5040fdf073709ebb

Looks good, you'll just need a final rebase so we can do a clean merge.

Issue #294, opened 2023-01-12
Add miniscript policy compiler function


notmandatory:
**Describe the enhancement**  

Add function to construct a `Descriptor` from a miniscript policy string.   See the `bdk-cli` [compiler docs] for usage and  [handle_compile_subcommand()] for example code. 

**Use case** 

Supporting wallets or tools that let a user create a descriptor from a miniscript policy string.

**Additional context**

This enhancement was inspired by a conversation with @rob1ham who could use this feature with python.

[compiler docs]: https://bitcoindevkit.org/bdk-cli/compiler/
[handle_compile_subcommand()]: https://github.com/bitcoindevkit/bdk-cli/blob/54dd1b21a5430d34fd1689725acbf404ca032f2e/src/handlers.rs#L497

Rob1Ham:
I have some code that I'm using to do this currently as a basic rust function that @vnprc helped me out with. It takes a policy string as an input like this:

```let policy = get_policy().map_err(|err| err.to_string())?;

    let policy = Concrete::<String>::from_str(&policy)?;
    let segwit_policy: Miniscript<String, Segwitv0> = policy
        .compile()
        .map_err(|e| Error::Generic(e.to_string()))?;

    let descriptor = Descriptor::new_wsh(segwit_policy).unwrap().to_string();
```

Its been a few days so I'm fuzzy on the details, were we talking about adding additional functionality around this?

Rob1Ham:
Ah yes sorry - yes Python bindings that supports this functionality would be great. For now we have a rust binary that we call to execute this code, which provides us with sufficient utility for an MVP. Having python bindings would allow me to have faster iterations in updating code, but I've been tag teaming learning the rust bdk library with @vnrpc to learn how its done in rust.

vnprc:
Here's the rust script I slapped together for Rob. It's very rough but apparently super useful. https://github.com/vnprc/rob_script

thunderbiscuit:
Very cool. Wondering if you'd be interested in adding your script to the [bitcoindevkit-scripts](https://github.com/thunderbiscuit/bitcoindevkit-scripts) repo @vnprc?

vnprc:
Oh cool. It needs some cleanup work but sure!

thunderbiscuit:
This issue has gone a bit stale and we have not had discussions or request to add this. Will close for now but feel free to reopen if need be!

PR #306, opened 2023-02-01
Add from_miniscript_policy() method on Descriptor type

thunderbiscuit:
### Description
This PR is meant to allow us to continue the discussion opened in #294.

It currently doesn't compile. I'm basically trying to implement something like [what we have in the bdk example](https://github.com/bitcoindevkit/bdk/blob/master/examples/compiler.rs). One of the issue is that we don't have the `Descriptor` type directly, and so we can't call the methods we have in Rust like `new_wpkh()` and `new_wsh()`. We either need to expose that type somehow, or change the API slightly and add the choice of what type of script to use for the descriptor by passing it as an enum to the method I created here. Overall not a straightforward "port" of BDK directly, so we should look to get feedback on how people might use it before moving into a specific direction.

### Changelog notice
No changelog notice yet.

### Checklists

#### New Features:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -160,6 +160,7 @@ version = "0.26.0"
 dependencies = [
  "assert_matches",
  "bdk",
+ "miniscript",
  "uniffi",
  "uniffi_build",
  "uniffi_macros",
--- a/bdk-ffi/Cargo.toml
+++ b/bdk-ffi/Cargo.toml
@@ -11,6 +11,7 @@ name = "bdkffi"
 
 [dependencies]
 bdk = { version = "0.26", features = ["all-keys", "use-esplora-ureq", "sqlite-bundled", "rpc"] }
+miniscript = { version = "8.0", features = ["serde", "compiler"] }
 uniffi_macros = { version = "0.21.0", features = ["builtin-bindgen"] }
 uniffi = { version = "0.21.0", features = ["builtin-bindgen"] }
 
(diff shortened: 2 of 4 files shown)


thunderbiscuit:
You can test this PR with this Kotlin script:
```
kotlin@file:Repository("file:///~/.m2/repository/")
@file:DependsOn("org.bitcoindevkit:bdk-jvm:0.27.0-PR306")

import org.bitcoindevkit.*

val miniscriptPolicy = "and(pk([22739455/84'/1'/0']tpubDDEqu57tdMjDiQhEobb2P2X8G6XMH1Vrrq3yhJmNSJtRT5gLLzAsXpFiKLGHPTDREkkeFaAmuzaDkCF4Kj9iMJggXLb48QyBwwP9CK94iZa/0/*),or(99@pk([3a686ab9/84'/1'/0']tpubDDAf2xGr2RqMHQwJBaYqYDr4dA3pYtgM1aCw9PeHSoUEQd9RYPKcjvZW42QT2cvNHHxa74NYcfw3jbyfZGWWwFJNWYHqXRVkp32jG2q1UjB/0/*),after(5)))"
val descriptor: Descriptor = Descriptor.fromMiniscriptPolicy(
    policy = miniscriptPolicy,
    network = Network.TESTNET,
    scriptType = "wsh"
)
println("The descriptor for miniscript policy $miniscriptPolicy is\n${descriptor.asStringPrivate()}")
```

PR #320, opened 2023-02-24
Add external signers

### LINKED ISSUE #222: Custom Signer
**Describe the enhancement**  
We'd like to have a custom signer interface similar to https://docs.rs/bdk/latest/bdk/wallet/signer/index.html

**Use case**  
We have a 2-of-3 multisig wallet setup where one key is stored directly on the mobile device. We'd like to store the private key separate from other wallet metadata in a secure store (i.e. Keychain / KeyStore) and only access it when it's needed (for signing) and otherwise not retain it in any way in memory.

Right now to accomplish that, we can maintain one instance of a `Wallet` object in memory with all 3 public keys and then create a separate `Wallet` object on demand with 2 public keys and the app private key and use that for signing.

This will work, but it would be better if we didn't have to instantiate an entire `Wallet` object to accomplish this and rather have access to a customer signer API like what exists at the bdk-core layer. One thing that will be important though is for the ergonomics of using the custom signer to be simple (i.e. just need to provide the raw private key if possible).

We also might be interested in similar functionality for other parts of project (like server signing) cc @alexatblock for more info

thunderbiscuit:
This PR will bring the ability to add external signers to the wallet.

Very much in draft mode. I also think we'll need to really dig into the exact uses for it to make sure we expose all of the APIs required for users (`InputSigner` vs `TransactionSigner` traits, `SignOptions`, etc.)

This PR closes #222.

### Changelog notice
```md
- APIs Added
  - Add external signers to the `Wallet` type [#320]

[#320]: https://github.com/bitcoindevkit/bdk-ffi/pull/320
```

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -3,6 +3,11 @@ namespace bdk {
 };
 
 [Error]
+enum CustomError {
+  "SignerError",
+};
+
+[Error]
 enum BdkError {
   "InvalidU32Bytes",
   "Generic",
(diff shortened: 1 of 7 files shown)


thunderbiscuit:
The questions I have remaining for this are around the InputSigner vs TransactionSigner. On the [Rust side](https://docs.rs/bdk/latest/bdk/wallet/signer/trait.InputSigner.html) you get a TransactionSigner for free if you implement the InputSigner, but I'm not sure how/if I should translate that to the FFI layer.

PR #321, opened 2023-03-08
Streamline CI workflows

thunderbiscuit:
## Description
This PR fixes a lot of the small differences between the styles of the different GitHub Actions workflows we have. Examples of the changes proposed:
1. Capitalize the first letter of the name of the CI task
2. Add a newline between tasks for easier readability
3. Pin the version of the image used (`ubuntu-20.04` instead of `ubuntu-latest`, etc.)

In this PR all changes are purely stylistic, so as not to introduce a change in behaviour at the same time as changes in style.

#### All Submissions:

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -1,4 +1,4 @@
-name: bdk-ffi CI
+name: Rust layer CI
 on:
   push:
     paths:
@@ -10,7 +10,7 @@ on:
 jobs:
   build-test:
     name: Build and test
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
         rust:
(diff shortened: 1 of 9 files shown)


.github/workflows/cont_integration.yml
```diff
@@ -1,4 +1,4 @@
-name: bdk-ffi CI
+name: Rust Layer CI
```
notmandatory, line 1: small nit.. but for consistency with having only the first word capitalized (CI is ok since it's an acronym):
```suggestion
name: Rust layer CI
```

notmandatory (approved):
ACK a5facf875a915e01323970899b688e0d8a0b1e85

.github/workflows/cont_integration.yml
thunderbiscuit: Good catch.

PR #326, opened 2023-03-21
Add SignOptions to Wallet.sign() params

### LINKED ISSUE #323: Expose SignOptions struct
Would be cool be able to set SignOptions like [allow_all_sighashes](https://docs.rs/bdk/latest/bdk/wallet/signer/struct.SignOptions.html#structfield.allow_all_sighashes) when a psbt needs to be signed by the wallet. An use would case be when we need import a psbtBase64 from other wallet that contain a custom sighash

notmandatory:
### Description

Fixes #323 

### Notes to the reviewers

I didn't include the `tap_leaves_options: TapLeavesOptions` because it would require more deeply nested structures and is not currently required.

### Changelog notice

- Add SignOptions to Wallet.sign() params.

#### All Submissions:

#### New Features:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,9 +21,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.69"
+version = "1.0.70"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


thunderbiscuit (approved):
ACK 85a420731097805aaf1593aa9ef359f7e31d139c.

PR #333, opened 2023-03-28
Use Address type in address field on AddressInfo

### LINKED ISSUE #332: Mirror `AddressInfo` API from Rust BDK
Our current `AddressInfo` is defined as such:
```rust
pub struct AddressInfo {
    pub index: u32,
    pub address: String,
}
```

Whereas the Rust BDK AddressInfo is
```rust
pub struct AddressInfo {
    pub index: u32,
    pub address: Address,
    pub keychain: KeychainKind,
}
```

We previously didn't have all the ingredients to mirror the Rust API but we now do; let's fix this type.

This will ensure that the `Wallet.get_address()` method returns an `AddressInfo` with a proper `Address` inside, with all goodies that come with it.

thunderbiscuit:
This PR leverages the newly exposed methods on the `Address` type and uses that in the AddressInfo type instead of a address string.

## Description
Fixes #332.

## Notes to the reviewers
As with the descriptor type, I'm changing the `to_string()` method to `as_string()` in the FFI. I don't remember why we made this choice way back when. Is that because of the name clash with the toString() methods on Kotlin and Swift? Just looking for validation before locking the API in.

### Changelog notice
```md
API Changes
- address field on `AddressInfo` type is now of type `Address` [#334]

[#334]: https://github.com/bitcoindevkit/bdk-ffi/pull/334
```

#### All Submissions:

#### New Features:

* [x] I've fixed the tests for the new feature

### DIFF
--- a/bdk-ffi/src/lib.rs
+++ b/bdk-ffi/src/lib.rs
@@ -47,17 +47,20 @@ pub struct ScriptAmount {
 
 /// A derived address and the index it was found at.
 pub struct AddressInfo {
-    /// Child index of this address
+    /// Child index of this address.
     pub index: u32,
-    /// Address
-    pub address: String,
+    /// Address.
+    pub address: Arc<Address>,
+    /// Type of keychain.
+    pub keychain: KeychainKind,
 }
 
 impl From<BdkAddressInfo> for AddressInfo {
-    fn from(x: bdk::wallet::AddressInfo) -> Self {
+    fn from(x: BdkAddressInfo) -> Self {
         AddressInfo {
             index: x.index,
-            address: x.address.to_string(),
+            address: Arc::new(Address::from(x.address)),
+            keychain: x.keychain,
         }
     }
 }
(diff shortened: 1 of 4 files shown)


bdk-ffi/src/lib.rs
```diff
@@ -47,17 +47,20 @@ pub struct ScriptAmount {
...
-    pub address: String,
+    /// Address.
+    pub address: Arc<Address>,
+    /// Type of keychain.
+    pub keychain: KeychainKind,
 }
 
 impl From<BdkAddressInfo> for AddressInfo {
-    fn from(x: bdk::wallet::AddressInfo) -> Self {
+    fn from(x: BdkAddressInfo) -> Self {
```
notmandatory, line 59: nit: can we use something more descriptive instead of `x`, something like `ai`:
```suggestion
    fn from(ai: BdkAddressInfo) -> Self {
```

notmandatory (approved):
ACK 5ee8698e0ab2976d52b45a9cd926b6c11b938b5c

Thanks for the extra cleanups !

PR #335, opened 2023-04-05
Use version 21 of the Android NDK in the CI test, build, and publish workflows

thunderbiscuit:
This PR reverts an update to our Android CI workflows.

For info on why this is needed, see issues #242 and #243.


thunderbiscuit:
@notmandatory this is the first we should merge IMO, so we get to test the CI on other PRs before merging them. Following that would be 337 and 334 if they're ready in time.

notmandatory (approved):
ACK beb75dd5524c0f16f0715d979b685c56ba95c15d

PR #345, opened 2023-04-18
Fix Android build in CI

thunderbiscuit:
Not yet sure why the Android build breaks. Might have to do with the update to the ubuntu-22.04 image.

### DIFF
--- a/.github/workflows/publish-jvm.yaml
+++ b/.github/workflows/publish-jvm.yaml
@@ -46,7 +46,7 @@ jobs:
   build-jvm-full-library:
     name: Create full bdk-jvm library
     needs: [build-jvm-macOS-M1-native-lib]
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-20.04
     steps:
       - name: Checkout publishing branch
         uses: actions/checkout@v2
--- a/.github/workflows/publish-jvm.yaml
+++ b/.github/workflows/publish-jvm.yaml
@@ -24,6 +24,9 @@ jobs:
           distribution: temurin
           java-version: 11
 
+      - name: "Set default Rust version to 1.67.0"
+        run: rustup default 1.67.0
+
       - name: Install aarch64 Rust target
         run: rustup target add aarch64-apple-darwin
 
@@ -67,6 +70,9 @@ jobs:
         with:
           distribution: temurin
           java-version: 11
+
+      - name: "Set default Rust version to 1.67.0"
+        run: rustup default 1.67.0
 
       - name: Build bdk-jvm library
         run: |
--- a/.github/workflows/publish-android.yaml
+++ b/.github/workflows/publish-android.yaml
@@ -37,6 +37,9 @@ jobs:
           distribution: temurin
           java-version: 11
 
+      - name: "Set default Rust version to 1.67.0"
+        run: rustup default 1.67.0
+
       - name: "Install Rust Android targets"
         run: rustup target add x86_64-linux-android aarch64-linux-android armv7-linux-androideabi
 
--- a/.github/workflows/publish-android.yaml
+++ b/.github/workflows/publish-android.yaml
@@ -10,7 +10,7 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-20.04
     steps:
       - name: "Install Android NDK 21.4.7075529"
         run: |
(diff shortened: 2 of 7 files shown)


yellowHatpro:
I might be super wrong, but isn't ubuntu 20.04 also affected due to NDK replacement?

thunderbiscuit:
Yep indeed they both need the manual download of NDK 21. I wondered whether there was something else about the ubuntu-22.04 image that might have caused the issue, but indeed both versions return the same error. So I don't think the upgrade to the 22.04 version was the problem in the end.

notmandatory:
Give this a try to fix the build: https://github.com/bitcoindevkit/bdk-ffi/pull/348/commits/afa284f0b9d45b78de1a973f8851e6cf7e62eb34

I think it broke because of an issue in rust 1.68 which is default in ubuntu 20.04 and 22.04. Sounds like this issue: https://github.com/cross-rs/cross/issues/1222

thunderbiscuit:
I have added a workflow_dispatch to all test workflows so as to be able to trigger them manually in the future (so as to not need the 1 line changes in the readmes just to run specific workflows).

bdk-jvm/README.md
```diff
@@ -98,3 +98,5 @@ dependencies {
 
 [`bdk`]: https://github.com/bitcoindevkit/bdk
 [`bdk-ffi`]: https://github.com/bitcoindevkit/bdk-ffi
+
+Temporary line: needed to run CI
```
notmandatory, line 102: Did you forgot to remove this line?

bdk-python/README.md
```diff
@@ -46,3 +46,5 @@ tox -vv
 ```shell
 pip install ./dist/bdkpython-<yourversion>-py3-none-any.whl
 ```
+
+Temporary line: needed to run CI
```
notmandatory, line 50: Did you forgot to remove this line?

bdk-swift/README.md
```diff
@@ -50,3 +50,5 @@ own GitHub repository use the following steps:
 [Xcode]: https://developer.apple.com/documentation/Xcode
 [`bdk`]: https://github.com/bitcoindevkit/bdk
 [`bdk-ffi`]: https://github.com/bitcoindevkit/bdk-ffi
+
+Temporary line: needed to run CI
```
notmandatory, line 54: Did you forgot to remove this line?

.github/workflows/publish-jvm.yaml
```diff
@@ -43,7 +46,7 @@ jobs:
   build-jvm-full-library:
     name: Create full bdk-jvm library
     needs: [build-jvm-macOS-M1-native-lib]
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-20.04
```
notmandatory, line 49: Should the `publish-android`, `publish-python` and `test-python` or any other places with `ubuntu-22.04` also be changed to `ubuntu-20.04`?

notmandatory (commented):
ACK 52b6dfb9ff901dd6471f0d244a43a7653b4fa8b1, with a few comments.

thunderbiscuit:
Sorry man I left this PR not in its final state this weekend. Today I'll squash all commits into one, and remove the extra lines in the readmes (I still needed them yesterday to make sure I ran all the workflows with the latest changes on the yaml, but now that they have run and succeeded, I can remove them).

The idea of running everything on 20.04 I think might be a good one for the short term. I'll fix those everywhere as well.

bdk-swift/README.md
thunderbiscuit: I had a few more changes to the yaml files so I still needed them yesterday for my last push, but now that everything has passed I will remove them.

bdk-python/README.md
thunderbiscuit: Will remove now.

thunderbiscuit:
The build for the older versions of Python breaks, but I'm also realizing that we are supporting probably way too old versions; Python 3.6 is 8 years old at this point, and even Debian stable (version 10, Bullseye) ships with 3.9. I feel it fair to expect users of the bitcoin development kit Python bindings use a recent-ish version of Python, and propose we deprecate support for 3.6 and 3.7.

thunderbiscuit:
Note that I also removed the PyPy releases in favour of just the CPython ones. I'd be happy to bring them back if we get demand for it, but at this point I don't know that it's really useful, and the versions it was even supporting were 3.7 and 3.8.

I think the cleaner approach is to start by supporting recent versions of CPython, and address demands for older/esoteric versions if and as they come.

notmandatory (approved):
ACK 871a06d1ce39456a2283cca5a927379a10db22ef

Nice well organized PR!

PR #348, opened 2023-04-22
Fix android build

notmandatory:
### Description

Don't merge. Testing android workflow.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/.github/workflows/test-android.yaml
+++ b/.github/workflows/test-android.yaml
@@ -18,7 +18,8 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-22.04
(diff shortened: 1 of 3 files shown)


thunderbiscuit:
Thank you Steve!

The fact that they change the image defined in the yaml CI workflow under you is mind-boggling to me. Any run is likely to break because of some update pushed by GitHub on any given day, and any of those can cost your devs days of work. Better to have explicit upgrades IMO.

notmandatory:
I agree, much better to manually set our default rust version and only upgrade it when we need to.

PR #352, opened 2023-05-01
Fix bdk-swift/build-local-swift.sh for aarch64-apple-ios-sim

notmandatory:
### Description

Fix bdk-swift/build-local-swift.sh for aarch64-apple-ios-sim, this matches the same fix made for the `bdk-swift` publishing workflow here: bitcoindevkit/bdk-swift#44

### Notes to the reviewers

I was able to remove `-Z build-std` because `std` is now included with the `aarch64-apple-ios-sim` target, so we don't need to build it ourselves. 

Also added `--profile release-smaller` with `aarch64-applie-ios-sim`. It used to not work with `nightly` but looks like it's working now and saves about 20MB on the final xcframework zip file.

### Changelog notice

None

#### All Submissions:

### DIFF
--- a/.github/workflows/test-swift.yaml
+++ b/.github/workflows/test-swift.yaml
@@ -36,7 +36,7 @@ jobs:
       - name: Build bdk-ffi for aarch64-apple-darwin
         run: cargo build --package bdk-ffi --profile release-smaller --target aarch64-apple-darwin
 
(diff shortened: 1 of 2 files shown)


thunderbiscuit (approved):
ACK d3895441d34dd1f11f887ff61442594e73a1b660.

PR #358, opened 2023-05-25
Expose the `add_foreign_utxo` method on TxBuilder type

thunderbiscuit:
### Description
This PR fixes #329.

### Notes to the reviewers
Not ready for review yet.

### Changelog notice
```md
APIs Added:
    - Add TxBuilder.add_foreign_utxo method [#358]

[#358]: https://github.com/bitcoindevkit/bdk-ffi/pull/358
```

* [x] I've signed all my commits
* [ ] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [ ] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:
* [ ] I've added tests for the new feature
* [ ] I've added docs for the new feature

PR #362, opened 2023-06-01
Add Wallet::get_psbt_input function and Descriptor::max_satisfaction_weight

notmandatory:
### Description

Add `Wallet::get_psbt_input` and `Descriptor::max_satisfaction_weight` functions and required structs and enums needed by #358.

Also added `Input` `from_json` constructor and `json_serialize` functions.

### Notes to the reviewers

~~After the `TxBuilder::add_foreign_utxo` function is added the `Input._inner` field should be renamed to `inner`.~~

### Changelog notice

- Add Input, PsbtSighashType, EcdsaSighashType and SchnorrSighashType
- Add Input from_json constructor and json_serialize functions
- Add Wallet::get_psbt_input function
- Add Descriptor::max_satisfaction_weight function

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/src/psbt.rs
+++ b/bdk-ffi/src/psbt.rs
@@ -83,12 +83,26 @@ impl PartiallySignedTransaction {
 /// transaction.
 #[derive(Debug)]
 pub(crate) struct Input {
-    _inner: BdkInput,
+    inner: BdkInput,
+}
+
+impl Input {
+    /// Create a new PSBT Input from a JSON String.
+    pub(crate) fn from_json(input_json: String) -> Result<Self, BdkError> {
+        let input = serde_json::from_str(input_json.as_str())?;
+        Ok(Self { inner: input })
+    }
+
+    /// Serialize the PSBT Input data structure as a JSON String.
+    pub(crate) fn json_serialize(&self) -> String {
+        let input = &self.inner;
+        serde_json::to_string(input).unwrap()
+    }
 }
 
 impl From<BdkInput> for Input {
     fn from(input: BdkInput) -> Self {
-        Input { _inner: input }
+        Input { inner: input }
     }
 }
 
--- a/bdk-ffi/src/psbt.rs
+++ b/bdk-ffi/src/psbt.rs
@@ -1,5 +1,8 @@
 use bdk::bitcoin::hashes::hex::ToHex;
+use bdk::bitcoin::util::psbt::Input as BdkInput;
 use bdk::bitcoin::util::psbt::PartiallySignedTransaction as BdkPartiallySignedTransaction;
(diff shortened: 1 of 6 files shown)


bdk-ffi/src/psbt.rs
```diff
@@ -76,6 +79,48 @@ impl PartiallySignedTransaction {
...
+    }
+}
+
+/// A Signature hash type for the corresponding input. As of taproot upgrade, the signature hash
+/// type can be either [`EcdsaSighashType`] or [`SchnorrSighashType`] but it is not possible to know
+/// directly which signature hash type the user is dealing with. Therefore, the user is responsible
+/// for converting to/from [`PsbtSighashType`] from/to the desired signature hash type they need.
+#[derive(Debug)]
+pub(crate) struct PsbtSighashType {
+    inner: BdkPsbtSighashType,
```
thunderbiscuit, line 101: I like the idea of using `inner` as the variable name, and wonder if we should eventually refactor other structs that use the same design pattern (simple wrappers over a BDK types).

thunderbiscuit (approved):
ACK 1e1b794380abce59ed7e3e20f65da94416da3526. Only thing is it appears your commits are not signed anymore..?

notmandatory:
I fixed commit signatures, and added `Input` `from_json` constructor and `json_serialize` functions.

thunderbiscuit:
ACK 7b79f3919429d6eaa2e4b16cdd1acb2708d4d772 with some questions.

Is this JSON serialization/deserialisation deterministic? Like can someone arrive from an psbt input built from a different tool and exported in json and simply import it here? Or are there corner cases? I'm not clear enough on the Rust serialization system. It looks like its not done by the rust-bitcoin crate, but just the sort of "standard" serde crate (it's a derived method). Sorry these feel like beginner questions but I was thinking it'd be cool if we can make sure we add the methods that allow us to be compatible with other implementations. Just not clear on whether a input in JSON format coming from Sparrow (Java) or Blue (bitcoin-js) would just work here, or if it's just the "jsonization" of the Rust struct itself.

notmandatory:
As far as I could find there's no standard serialization format for just the PSBT Output, only the whole PSBT has a standard cross-implementation serialization format. So you have a good point that this JSON approach would not be compatible across other wallets. That makes me think this could be a dead-end approach and that we should encourage users to collaboratively add inputs and outputs by constructing PSBTs and using `PartiallySignedTransaction.combine` instead of `Wallet.add_foreign_utxo`, but I'm not sure if that solves the required use-cases or not.

https://docs.rs/bitcoin/0.29.2/bitcoin/util/psbt/struct.PartiallySignedTransaction.html#method.combine

notmandatory:
Closing for now, may bring it back later after we finish updating to BDK 1.0.

PR #363, opened 2023-06-05
Use new bitcoin-ffi crate for Network type

thunderbiscuit:
Just playing around with the idea of using a new bitcoin-ffi crate where rust-bitcoin types are defined.


thunderbiscuit:
This branch is old and not used anymore. But the work to use a "bitcoin-ffi" crate continues!

Issue #350, opened 2023-04-25
Python libraries all build using the same generic name


thunderbiscuit:
The attempt at the publishing of the `0.28.0` release of bdkpython [failed](https://github.com/bitcoindevkit/bdk-ffi/actions/runs/4790684509).

This is because the new, cleaner build workflow built libraries that all had the same name, and their upload to PyPI therefore failed (you can't publish more than one library with the same name). The library they all attempted to upload was
```txt
bdkpython-0.28.0-py3-none-any.whl
```

Whereas the names we're looking for (from previous versions) are similar to the following:
```txt
bdkpython-0.27.1-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
bdkpython-0.27.1-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
bdkpython-0.27.1-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl

bdkpython-0.27.1-cp310-cp310-macosx_12_0_universal2.whl
bdkpython-0.27.1-cp39-cp39-macosx_12_0_universal2.whl
bdkpython-0.27.1-cp38-cp38-macosx_12_0_universal2.whl

bdkpython-0.27.1-cp310-cp310-win_amd64.whl
bdkpython-0.27.1-cp39-cp39-win_amd64.whl
bdkpython-0.27.1-cp38-cp38-win_amd64.whl
```

After exploring solutions, the cleanest approach seems to be what is described [here](https://stackoverflow.com/a/64921892/10523694).

The one slight issue with that is that for macOS, even if you build a universal wheel, it still names it according to the platform onto which is was built. This is fixed by specifying the command line flag `--plat-name macosx_12_0_universal2`.

I created this issue to explain how I came to the fix I used for the 0.28.0 release, and will link to this issue in the script that uses the trick.

thunderbiscuit:
Follow up small issue on [this run](https://github.com/bitcoindevkit/bdk-ffi/actions/runs/4792823684/jobs/8524991761)

```sh
 ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/        
         Binary wheel 'bdkpython-0.28.1-cp310-cp310-linux_x86_64.whl' has an    
         unsupported platform tag 'linux_x86_64'.        
```

thunderbiscuit:
[...]

Issue #339, opened 2023-04-11
PyPi does not load BDK correctly.


Rob1Ham:
**Describe the bug**  
While compiling from source it works appropriately, but when downloading from PyPi it does not work.

**To Reproduce**  
Here is the logs from my terminal:

```
 rob@macbook workbench % pip install bdkpython==0.27.1
ERROR: Could not find a version that satisfies the requirement bdkpython==0.27.1 (from versions: 0.0.1, 0.0.2, 0.0.3, 0.0.4)
ERROR: No matching distribution found for bdkpython==0.27.1
 rob@macbook workbench % pip install bdkpython
Collecting bdkpython
  Using cached bdkpython-0.0.4-py3-none-any.whl (10.1 MB)
Installing collected packages: bdkpython
Successfully installed bdkpython-0.0.4
 rob@macbook workbench % python3
Python 3.9.13 (main, Aug 25 2022, 18:29:29) 
[Clang 12.0.0 ] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import bdkpython as bdk
>>> descriptor = bdk.Descriptor("wpkh(tprv8ZgxMBicQKsPcx5nBGsR63Pe8KnRUqmbJNENAfGftF3yuXoMMoVJJcYeUw5eVkm9WBPjWYt6HMWYJNesB5HaNVBaFc1M6dRjWSYnmewUMYy/84h/0h/0h/0/*)", bdk.Network.TESTNET) 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'bdkpython' has no attribute 'Descriptor'
```

Was running on my MacOSXv 13.2.1 with intel chips.

notmandatory:
@thunderbiscuit I'd like to get this figured out for the 0.28 release so tagging it there.  It sounds like somehow the python bindings code is generated correctly but the corresponding rust binary libs aren't being added the the packages.

thunderbiscuit:
If it does work, I'm not sure of the best approach to provide it out of the box. The CI currently builds with the latest possible image for GitHub Actions for macOS, `macos-12`. There might be ways to go around that and make the CI build macOS 13 wheels, but... not sure yet how one would go about doing that. We'd need to dig a bit more.

thunderbiscuit:
I believe this is now resolved with some of the new build processes! If it isn't, feel free to reopen.

PR #374, opened 2023-06-15
Use macos-13 image to test and publish Python libraries

thunderbiscuit:
This PR brings support for macOS 13 for bdkpython.

See related issues #339 and #350

Note that as is, this doesn't actually support multiple versions of macOS for the library. I think the issue is in the naming convention used for the file (see issue #350). We're currently setting the name of the file using [the --plat_name argument](). If I bump it to `macosx13_0_universal2`, the library cannot be imported by pip on my macOS 12, and if we leave it as is (`macosx12_0_universal2`), the library doesn't work on 13 (see issue #339).

I'm wondering if more experienced Python devs might have suggestions for us? Pinging @andreasgriffin...


thunderbiscuit:
This is finally ready for review! This PR turned into much more than supporting macOS 13. There were some issues with the Windows tests, and I learned a ton about wheel naming. We are now publishing separate macOS wheels for the two platforms, and so each library is smaller (3.6 MB on average) rather than having 1, universal wheel that sometimes pip didn't like.

bdk-python/scripts/generate-linux.sh
```diff
@@ -0,0 +1,18 @@
...
+set -euo pipefail
+${PYBIN}/python --version
+${PYBIN}/pip install -r requirements.txt
+
+echo "Generating bdk.py..."
+cd ../bdk-ffi/
+cargo run --bin uniffi-bindgen generate src/bdk.udl --language python --out-dir ../bdk-python/src/bdkpython/ --no-format
+
+echo "Generating native binaries..."
+rustup default 1.67.0
```
notmandatory, line 12: In the future this should probably come from en env variable set by the github action, but this is fine until we need to update it to a new version.

notmandatory (approved):
ACK dc79b78b2dea15be409c94315392208f43ca1ce8

PR #376, opened 2023-06-17
Update changelog, bump snapshot and dev versions of libraries

thunderbiscuit:
This PR does four things:

1. Bump the development versions of bdk-android, bdk-jvm, and bdk-python.
2. Update the changelog.
3. Update the release_minor_version issue template
4. Remove the release_patch_version issue template.

I propose to remove the release patch template because at the moment it's just a copy of the minor release, but that's not really the workflow when we release a patch (which is often because one of the builds/publishing tasks failed and we need a new version to publish).

I'm not sure really how to create a set workflow for those patch releases, and rather than having a file that is not representative or reality I figured maybe we could just remove it for now. If we end up wanting to define exactly our workflow for patches we can craft one from scratch in a separate PR.


notmandatory (approved):
ACK d0a7315c9d48e907aed0fad9ebff8006122807bd

Looks good!

PR #380, opened 2023-07-03
Add Windows target to Kotlin/JVM published library

thunderbiscuit:
This PR adds the `x86_64-pc-windows-msvc` Windows target to our bdk-jvm library.

I try to be careful any time I touch publishing CI jobs, and particularly those I can't test prior to actually releasing. I think this is clean. If it passes review, I might release a 0.29.1 version of the library (which would also allow us to test the new Python workflow).


notmandatory (approved):
ACK 3a9400e6c738a73c65fe93a8868bd199a655efd6

PR #382, opened 2023-07-04
Fix CI bdk-jvm

thunderbiscuit:
Small fixes to the workflow that releases the bdk-jvm library to enable Windows support.


notmandatory (approved):
ACK 5fc189717db0c6d3ae60536f6c7b6f26a52fdfb0

PR #392, opened 2023-08-04
refactor(keys): remove unused mutex

### LINKED ISSUE #390: Inner of DescriptorSecretKey, DescriptorPublicKey don't need to have a mutex
**Describe the enhancement**  

The `DescriptorSecretKey` and `DescriptorPublicKey` types are not mutable so a mutex should not be needed the structures.

```rust
pub(crate) struct DescriptorSecretKey {
    pub(crate) descriptor_secret_key_mutex: Mutex<BdkDescriptorSecretKey>,
}
```

```rust
#[derive(Debug)]
pub(crate) struct DescriptorPublicKey {
    pub(crate) descriptor_public_key_mutex: Mutex<BdkDescriptorPublicKey>,
}
```

Creating an issue so we don't forget to fix this.

thunderbiscuit:
The DescriptorSecretKey and DescriptorPublicKey types used unnecessary mutexes on their inner types.

Fixes #390

### Changelog notice
None.

#### All Submissions:

### DIFF
--- a/bdk-ffi/src/descriptor.rs
+++ b/bdk-ffi/src/descriptor.rs
@@ -1,3 +1,4 @@
+use std::ops::Deref;
 use crate::{BdkError, DescriptorPublicKey, DescriptorSecretKey};
 use bdk::bitcoin::secp256k1::Secp256k1;
 use bdk::bitcoin::util::bip32::Fingerprint;
@@ -11,7 +12,6 @@ use bdk::template::{
     DescriptorTemplate,
 };
 use bdk::KeychainKind;
-use std::ops::Deref;
 use std::str::FromStr;
 use std::sync::Arc;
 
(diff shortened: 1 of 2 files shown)


thunderbiscuit:
Do not merge. Bunch of errors still.

notmandatory (approved):
ACK 106d31c9c361908851c0cd0d5cb5c1c8ac991739

Good refactor, much simpler this way!

bdk-ffi/src/descriptor.rs
```diff
@@ -36,9 +35,9 @@ impl Descriptor {
         keychain_kind: KeychainKind,
         network: Network,
     ) -> Self {
-        let derivable_key = secret_key.inner_mutex.lock().unwrap();
+        let derivable_key = &secret_key.inner;
 
-        match derivable_key.deref() {
+        match derivable_key {
             BdkDescriptorSecretKey::XPrv(descriptor_x_key) => {
```
thunderbiscuit, line 41: @notmandatory I forgot to ask but does this seem like the best approach to you?

I'm now borrowing the `secret_key.inner`, and the `derivable_key` doesn't need to be dereferenced anymore. I just want to make sure I'm not messing up something with the ownership rules.

notmandatory: Your approach looks good to me, since you only need to read the value borrowing it is the right thing to do.

Issue #245, opened 2022-11-15
Update CI audit task


notmandatory:
The current audit task we're using ([actions-rs/audit-check@v1](https://github.com/actions-rs/audit-check)) isn't maintained. We should find one that is being maintained. For example:

[actions-rust-lang/audit@v1](https://github.com/actions-rust-lang/audit)

notmandatory:
If there aren't any other good options then I'd rather stick with what we have then remove it completely. Another option is to manually run the `cargo audit` tool, as long as we can make sure it properly fail the workflow if one of the audit checks fails. The [`cargo audit`](https://github.com/RustSec/rustsec/tree/main/cargo-audit) tool itself is well supported by the rust community.

notmandatory:
The `cargo-audit` page recommends this for using it with github actions: https://github.com/rustsec/audit-check

PR #402, opened 2023-09-13
Update bdk version to 1.0.0-alpha.2

thunderbiscuit:
### Description
This is the beginning of the migration to the new 1.0 API for the bitcoin development kit bindings.

### Notes to the reviewers
This is likely to be a fairly big PR; let's review it one commit at a time.

### Changelog notice
Oh lordy lord the changelog on this thing


.github/workflows/audit.yml
reez, line 1: Is #245 the one that this change is for, if so we can link that Issue to this PR possibly.

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/AndroidLibTest.kt
reez, line 1: Ill open an issue to match these tests on the Swift/iOS side of things post-merge of this PR

bdk-jvm/lib/src/test/kotlin/org/bitcoindevkit/JvmLibTest.kt
reez, line 1: These match Android tests nicely, thumbs up.

bdk-ffi/Cargo.toml
```diff
@@ -17,12 +17,30 @@ path = "uniffi-bindgen.rs"
...
+uniffi = { version = "=0.23.0", features = ["bindgen-tests"] }
 assert_matches = "1.5.0"
+
+[profile.release-smaller]
+inherits = "release"
+opt-level = 'z'     # Optimize for size.
+lto = true          # Enable Link Time Optimization
+codegen-units = 1   # Reduce number of codegen units to increase optimizations.
+panic = "abort"     # Abort on panic
+strip = "debuginfo" # Partially strip symbols from binary
```
reez, line 46: I didn't have this change on my branch and also can't remember going over this, so it stood out to me that's all. I generally understand what this is trying to do but is this related to any specific Issue? Just for my own understanding or for reference is the only reason I'm asking.

bdk-ffi/src/bdk.udl
```diff
@@ -116,260 +105,20 @@ enum WordCount {
...
-  TxBuilder drain_to(Script script);
-
-  TxBuilder enable_rbf();
-
-  TxBuilder enable_rbf_with_sequence(u32 nsequence);
-
-  TxBuilder add_data(sequence<u8> data);
-
-  TxBuilder set_recipients(sequence<ScriptAmount> recipients);
+  string to_qr_uri();
```
reez, line 114: I'm cool leaving this in but I think it is unused at the moment I think?

reez (commented):
***I have tested this branch locally***

Just calling out a couple high level changes here for reference:

- Related adding of `esplora.rs` and removal of `blockchain.rs`
- Some changes to scripts and .sh files related to removal of workspace
- `database.rs` removed and will be reimplemented in some different form in follow up PR's
- `psbt.rs` fully commented out but stays because we know it will be implemented fully in follow up PR's

.github/workflows/audit.yml
thunderbiscuit: Yep that's the one. Good idea to link it. I don't see it as a clean fix of it however, so not sure what I should do there.

reez: ok sweet, makes sense. I think its good enough now that we have it mentioned in this PR and also there is a reference to it now in the Issue per our mention of it here!

bdk-ffi/Cargo.toml
thunderbiscuit: Here do you mean the release-smaller profile, or the change of the `strip = "debuginfo"` line?

The profile got moved from the root Cargo.toml because of the removal of the cargo workspace, but the strip = "debuginfo" is because our old strip = true prevents the `--library` flag from building the scaffolding code on Linux machines. This is explored more in thunderbiscuit/uniffi-examples#3, and it was timely that this guy happen to open the PR when he did because I ran into the issue just a few weeks later for this update.

But come to think of it this is not required _yet_, because we're not updating to the 0.24.3 version of uniffi-rs... I could revert that. No strong opinion on it yet; I've been meaning to test the difference in size for the binaries when the two flags are used.

reez: ahhh, that answers it. I was unsure if it was an all new block of code and a totally new addition to the project, but now that you mention it just got moved from a different file (the root `Cargo.toml`) because of the removal of the workspace that clears it up for me (not new code, but new spot!). Thanks!

bdk-ffi/src/bdk.udl
thunderbiscuit: Oh this might be a weird issue of how it looks in the GitHub UI I think (which makes it appear like the `to_qr_uri()` is part of the TxBuilder), but it's actually part of the `Address` type. So yes it's used.

bdk-ffi/Cargo.toml
thunderbiscuit: Also related is #405.

reez:
nice job thunder!!! 🚀

bdk-ffi/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "bdk-ffi"
-version = "0.30.0"
+version = "1.0.0-alpha.2"
 authors = ["Steve Myers <steve@notmandatory.org>", "Sudarsan Balaji <sudarsan.balaji@artfuldev.com>"]
```
notmandatory, line 4: In a future PR I want to remove this line and replace it with:
```toml
homepage = "https://bitcoindevkit.org"
repository = "https://github.com/bitcoindevkit/bdk"
```

notmandatory (approved):
ACK 7463fa7720f333f995c1fcc9c01669d4c72311d6

Looks great!  1.0 here we come.

PR #409, opened 2023-10-24
Add scan method on esplora client

thunderbiscuit:
This PR adds a temporary scan method on the Esplora client. Note that the long term solution to this involves not passing the wallet directly to the client. More discussion on this in bitcoindevkit/bdk#1153.


notmandatory (approved):
ACK 00cd55bb4650997d6cd9298de1be2258373089c4

reez:
✅ nice!

PR #410, opened 2023-10-25
tests: update all tests

reez:
### Description

Match all 4 platforms with tests.

### Notes to the reviewers

All 4 platforms now have 3 matching tests, testing:

1. Descriptor Bip-86
2. New Address
3. Balance

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-android/lib/build.gradle.kts
+++ b/bdk-android/lib/build.gradle.kts
@@ -53,7 +53,8 @@ dependencies {
     androidTestImplementation("com.github.tony19:logback-android:2.0.0")
     androidTestImplementation("androidx.test.ext:junit:1.1.3")
     androidTestImplementation("androidx.test.espresso:espresso-core:3.4.0")
-    androidTestImplementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.1")
+    androidTestImplementation("org.jetbrains.kotlin:kotlin-test:1.6.10")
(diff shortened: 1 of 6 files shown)


thunderbiscuit (commented):
This is good, ready for merge probably (although I think you could remove the commented out `testMemoryWalletNewAddress()`). 

But now that we're into the tests... this got me looking into the JVM tests and the Android tests, and they're both not super well done (oops... my fault... 3 of them don't actually test anything and just print to console lol). Are you able to fix those locally? (I know Android has extra requirements for the emulator and whatnot).

Overall, how do you feel about making this PR the one that starts us right and sets the tone on the new tests approach? A few thoughts to start this discussion:
- I think we should wait on anything that requires syncing an actual blockchain
- As discussed before a fair coverage of most of the "happy paths" would be a super good place to start
- Because those tests carry two objectives (testing and our currently _only_ documentation) I favour using explicit types everywhere (even thought the IDEs might flag this as redundant in many cases). I think this help readers who are simply glancing at the code in a web browser on GitHub rather than in a full IDE synced up with all dependencies and tooling and whatnot.

thunderbiscuit (approved):
ACK [372f79a](https://github.com/bitcoindevkit/bdk-ffi/pull/410/commits/372f79a10f6054937b70191d9fe79dfa1deabcff).

PR #406, opened 2023-10-19
Add TxBuilder, Script, Transaction, and PartiallySignedTransaction

thunderbiscuit:
This PR brings back some of the important types after the MVP 1.0 migration performed in #402, namely:
- `TxBuilder`
- `Transaction`
- `Script`
- `PartiallySignedTransaction`

### Notes to the reviewers
The biggest change here is the use of the `ScriptBuf` type instead of the `Script` type. This is new for this codebase and I need to dig a bit more into it.

This PR also introduces a new file, `bitcoin.rs`, where I added the types that are coming from rust-bitcoin and reexported by bdk. In theory this should allow for a smoother transition to a bitcoin-ffi crate.

### Changelog notice
```txt
APIs Added:
    - New TxBuilder type
    - New Transaction type
    - New Script type
    - New PartiallySignedTransaction type
```

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:
* [x] I've added tests for the new feature
* [x] I've added docs for the new feature

PR #411, opened 2023-10-31
Add workflow for live tests with manual trigger and weekly schedule

thunderbiscuit:
This PR adds the ability to trigger all live tests with a workflow dispatch, and will also run them once a week on master just for sanity checks.

The goal here is not to test on all possible architectures and Python versions, but rather to simply be able to run the tests that hit the testnet endpoints when/if we want to. Note that #406 add live and offline tests and only the offline tests are run the CI. Live tests must be run locally by contributors and reviewers.

See example run here: https://github.com/thunderbiscuit/bdk-ffi/actions/runs/6710735935

### DIFF
--- a/.github/workflows/live-tests.yaml
+++ b/.github/workflows/live-tests.yaml
@@ -1,0 +1,93 @@
+name: Run All Live Tests
+on: 
+  workflow_dispatch:
+  schedule:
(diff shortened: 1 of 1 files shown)


notmandatory (approved):
ACK 8eefbaf529babce2788df5f501d5e042433fe7d4

PR #414, opened 2023-11-09
Bump Rust compiler to 1.73.0 and update uniffi to 0.25.1

### LINKED ISSUE #242: Streamline the NDK version used in Android build and CI

### LINKED ISSUE #243: Fix Android instrumentation tests in CI

### LINKED ISSUE #413: Issues with updating the Rust compiler for the Android target
The Android library builds on Rust 1.67.0, but not on Rust 1.68.0 and above.

This issue will keep track of the work done to address this.

thunderbiscuit:
This PR fixes a number of longstanding issues with uniffi, the Rust compiler, and the Android NDK. In short, it updates all 3 to new versions:
1. Bumps NDK to 25.2.9519653
2. Bumps Rust compiler to 1.73.0
3. Bumps uniffi to 0.25.1
4. Removes the need to use nightly for the Swift builds.

Fixes #413
Fixes #243
Fixes #242
 Will help get #282 in.

We have two main reasons for bumping the uniffi version:
1. The 0.24 version of uniffi bring in async support, which we want to use for the async Esplora client.
2. The 0.25.X version of uniffi will bring inline API docs, and this is an important feature for the 1.0 release. We have decided to deprecate and not keep supporting the custom-built API docs maintained in the [api-docs](https://github.com/bitcoindevkit/bdk-ffi/tree/master/api-docs) directory. Progress on the merging of this can be tracked mostly on mozilla/uniffi-rs#1498

And one reason to bump the NDK:
1. Android emulator CI tests must be run on macos images, and we must use a later version of the NDK with them.

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -17,10 +17,8 @@ jobs:
     strategy:
       matrix:
         rust:
-          - version: 1.67.0
+          - version: 1.73.0
             clippy: true
-          # TODO 1: Should we keep this? We'll need to pin dependencies
-          # - version: 1.61.0 # MSRV
     steps:
       - name: "Checkout"
         uses: actions/checkout@v3
@@ -50,12 +48,6 @@ jobs:
       - name: "Update toolchain"
         run: rustup update
 
-      - name: "Pin dependencies for MSRV"
-        if: matrix.rust.version == '1.61.0'
-        run: |
-          cargo update -p hashlink --precise "0.8.1"
-          cargo update -p tokio --precise "1.29.1"
-          cargo update -p flate2 --precise "1.0.26"
       - name: "Build"
         run: cargo build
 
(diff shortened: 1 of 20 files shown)


thunderbiscuit:
I have tested the library locally on a version of the Devkit Wallet. This is a fairly important update across 3 different dimensions; happy to report that everything seems to work well.

reez:
I was able to build the bdk-swift library locally (on my M1 😢).

I then tested the library locally on a simplified version of BDKSwiftExampleWallet, able to create the wallet and sync the balance properly, so seems to be working well.

notmandatory (approved):
ACK f2cd561f25fd5d0b0aa7f2a54fdd8d802b39f280

reez:
ACK f2cd561f25fd5d0b0aa7f2a54fdd8d802b39f280

PR #417, opened 2023-11-16
fix: handle parsing and network errors in address new

reez:
### Description

Fixes "TODO 11".

Instead of panicking if the parsing fails, handle the error by propagating; avoids panicking and allows the caller of Address::new to handle the BdkError as necessary.

### Notes to the reviewers

Already using the ? operator so it seemed more idiomatic to continue with this style (instead of `match`), but open to changing this if needed.

#### All Submissions:

#### New Features:

#### Bugfixes:


thunderbiscuit (approved):
ACK e79ce982958b912c82de2e2eb6065438a98e8e62.

PR #428, opened 2023-11-29
Add sent_and_received method on Wallet

reez:
### Description

Adds the functionality for [sent_and_received](https://docs.rs/bdk/1.0.0-alpha.2/bdk/wallet/struct.Wallet.html#method.sent_and_received)

`sent_and_received` was a new method added in [BDK](https://github.com/bitcoindevkit/bdk/pull/1048).

Computes total input value going from script pubkeys in the wallet (sent) and the total output value going to script pubkeys in the wallet (received) in the tx passed into the function.

For _sent_ the function needs historical information about the outputs being spent (which are assumed to be indexed), while for _received_ it can rely solely on the tx's outputs without needing prior indexing of spent outputs.

### Notes to the reviewers

At the moment I decided to not add a code comment above the `sent_and_received` function in the `wallet.rs` file; my thinking was I would like to add a code comment similar to what is in the BDK documentation, but since the BDK documentation is still evolving I wouldn't want to add a code comment in bdk-ffi now but never get back around to updating it as the BDK documentation changes, and thus have a misleading code comment in the future.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/lib.rs
+++ b/bdk-ffi/src/lib.rs
@@ -23,6 +23,7 @@ use crate::types::AddressInfo;
 use crate::types::Balance;
 use crate::types::LocalUtxo;
 use crate::types::ScriptAmount;
+use crate::wallet::SentAndReceivedValues;
 use crate::wallet::TxBuilder;
 use crate::wallet::Update;
 use crate::wallet::Wallet;
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -107,6 +107,8 @@ interface Wallet {
 
   [Throws=BdkError]
   boolean sign(PartiallySignedTransaction psbt);
+
+  SentAndReceivedValues sent_and_received([ByRef] Transaction tx);
 };
 
 interface Update {};
@@ -245,6 +247,11 @@ interface EsploraClient {
 dictionary ScriptAmount {
   Script script;
   u64 amount;
+};
+
+dictionary SentAndReceivedValues {
+    u64 sent;
+    u64 received;
 };
 
 // ------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


bdk-ffi/src/lib.rs
```diff
@@ -39,6 +39,11 @@ pub struct ScriptAmount {
     pub amount: u64,
 }
 
+pub struct SentAndReceivedResult {
```
notmandatory, line 42: I think it'd be a little clearer to call this `SendAndReceivedValues` instead of `Result` since result has special meaning in rust. Also I'd put this struct in `wallet.rs` since that's the only place it's used.

reez: Thanks for the feedback! 

I was unsure of what to name that struct so I did my best guess based any other types I could find that might be similar-ish in the existing codebase (like`TxBuilderResult`, etc), so it's good to get your thoughts on what to name it. 

Also thanks for the suggestion of putting the struct in `wallet.rs`.

I'll make these suggested changes and update the PR:
- `SentAndReceivedResult` -> `SentAndReceivedValues`
- move `SentAndReceivedValues` struct to `wallet.rs`

notmandatory: No problem, I've also struggled to figure out what to call the types we use to replace the rust tuples!

reez:
_rebased on master and tests pass_

notmandatory:
For this function it's safe to go ahead and copy the BDK docs, I don't expect they will change for the 1.0 release.

thunderbiscuit (approved):
Tested ACK 46e0324f28dbd5d508cea862edf201c28431ee25.

I opened a new issue (#431) regarding API docs. I'm ok with not having any in this PR, and a further PR should remove them all so we can do a proper reset.

PR #429, opened 2023-12-06
feat: add new types module

thunderbiscuit:
This PR moves "floating" types from the `lib.rs` to a new, `types.rs` file.

My objective is to have a lib.rs file that simply has imports and uses.

This PR also brings back the `TxOut` and `LocalUtxo` types.


reez (approved):
Great idea that is in sync with how [bdk](https://github.com/bitcoindevkit/bdk/blob/master/crates/bdk/src/types.rs) handles this.

ACK 6b177f089359e48af43c2add2a945f615be924c7

PR #432, opened 2023-12-07
Add transactions method on wallet

reez:
### Description

Adds [transactions](https://docs.rs/bdk/1.0.0-alpha.2/bdk/wallet/struct.Wallet.html#method.transactions) method on `Wallet`.

### Notes to the reviewers

Simplified return type to list of `Transaction`s.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-python/tests/test_live_wallet.py
+++ b/bdk-python/tests/test_live_wallet.py
@@ -5,7 +5,7 @@ class TestLiveWallet(unittest.TestCase):
 
     def test_synced_balance(self):
         descriptor: bdk.Descriptor = bdk.Descriptor(
-            "wpkh([c258d2e4/84h/1h/0h]tpubDDYkZojQFQjht8Tm4jsS3iuEmKjTiEGjG6KnuFNKKJb5A6ZUCUZKdvLdSDWofKi4ToRCwb9poe1XdqfUnP4jaJjCB2Zwv11ZLgSbnZSNecE/0/*)",
+            "wpkh(tprv8ZgxMBicQKsPf2qfrEygW6fdYseJDDrVnDv26PH5BHdvSuG6ecCbHqLVof9yZcMoM31z9ur3tTYbSnr1WBqbGX97CbXcmp5H6qeMpyvx35B/84h/1h/0h/0/*)",
             bdk.Network.TESTNET
         )
         wallet: bdk.Wallet = bdk.Wallet.new_no_persist(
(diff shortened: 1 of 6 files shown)


thunderbiscuit (commented):
I think this works, but I have 2 questions I'm still wondering about:

1. The API doc for the method is one line: `Iterate over the transactions in the wallet.`. I wonder if (a) this is meant to be used for more than just returning the transactions, and (b) only returns transactions "owned" by the wallet. For example I see that it's basically traversing the TxGraph... but if I understand correctly you can add any transactions to this TxGraph (?)
2. The type of CanonicalTx.TxNode.tx is `&'a T`, which I don't understand well. Will you always be able to get a `Transaction` type out of it when you apply the .into() method on it?

reez:
Great Q’s/Feedback Thunder! I think I caught some of the conversation about this on the dev call this morning and then heard some suggestions, so I made some updates based on what I heard as what needed to happen which was to filter the transactions based on `isMine`.

So the changes I made: I’m filtering transactions based on `isMine`; I’m checking the input script and the output script, if either of those `isMine` I include that transaction in the final Vec.

Questions:
- is my understanding of what I heard on the dev call correct/complete?
- if so, does my updated code look like what you were thinking?
- would you like me to update to the unit tests I added `transactions` logic for based on this new `isMine` filtering?
- my only other thought was to add `isMine` as a bool parameter in the function signature so a user could say whether they wanted to filter on that or not (and if I did add that parameter I’d update the code inside of the function to reflect that new logic)?

thunderbiscuit:
Yeah roughly my take on it is that isMine is implied in the the `transaction` method at the moment. We can add isMine and tell people to filter, but that feels like a weird API, in part because not all users will get the message that they in fact need to do both, and also because we don't actually offer any methods (at the moment at least) to add anything to your TxGraph, so at the moment the isMine filter would be redundant.

My current take on it is this (happy to hear your thoughts): we should add the isMine filter to the method directly. This ensures the method is explicit in returning only wallet-related txs (even thought that's really all it can return at the moment). We open an issue on the bdk repo to discuss this weird API and see what might have been the thinking behind it. One good option proposed by @notmandatory was to add an argument `onlyMine` to the method (at the rust layer) or something like that, to make it explicit that the method can return much more than that if that flag is not set.

thunderbiscuit:
I'm changing my mind on this. I think we should expose the API as is on the Rust side, and open issues/fix it upstream if there are things to improve. For now, exposing it as is is the better/cleaner way. We also don't offer ways to impact the TxGraph manually at this point on the ffi side, so the problems that this could cause are limited.

bdk-python/tests/test_live_wallet.py
```diff
@@ -23,7 +23,16 @@ def test_synced_balance(self):
...
+        print(f"Transactions count: {len(wallet.transactions())}")
+        transactions = wallet.transactions()[:3]
+        for tx in transactions:
+            sent_and_received = wallet.sent_and_received(tx)
+            print(f"Transaction: {tx.txid()}")
+            print(f"Sent {sent_and_received.sent}")
+            print(f"Received {sent_and_received.received}")
+
+
+def test_broadcast_transaction(self):
```
thunderbiscuit, line 35: Small nit: your indentation here should be 4 spaces to the right so that the def is inside the class definition.

thunderbiscuit (changes requested):
Just a small nit. This is ready to go afterwards IMO.

bdk-python/tests/test_live_wallet.py
reez: good call, updated now ✅

reez:
@thunderbiscuit PR is updated now per suggestions and tests pass, thanks!

PR #441, opened 2024-01-10
Upgrade Rust bdk to alpha 3

thunderbiscuit:
This PR updates the version of bdk and associated libraries to the alpha 3 version.

The majority of the changes come from the introduction of new error enums. At the moment they are all erased into this one `Alpha3Error::Generic` error type. Further commits will pull the required errors one at a time and implement them until the generic variant is not required anymore.


reez (changes requested):
This looks great other than 1 requested change-

I'd like to not include the dependency `thiserror` since I think we can do without it, and then use something like this in `error.rs` instead:

```
#[derive(Debug)]
pub enum Alpha3Error {
    Generic,
}

impl fmt::Display for Alpha3Error {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        match self {
            Alpha3Error::Generic => write!(f, "Error in FFI"),
        }
    }
}

impl std::error::Error for Alpha3Error {}
```

That matches for `Alpha3Error` what I've got for `CalculateFeeError` (which matching my code isn't really the important thing, but not including the dependency `thiserror` is more what I'm going for). With those changes `thiserror` can be removed and things compile on my end and `cargo test --lib` still passes locally. But let me know if you have any thoughts related to that or if I'm missing anything.

Other than that this looks great!

thunderbiscuit:
Oh yeah great idea!

reez (approved):
LGTM 🚀

ACK 54beb23f13fca04867e024bcfbacde4b3af70395

PR #443, opened 2024-01-11
Upgrade API to match bdk 0.29

thunderbiscuit:
This PR updates the API to adapt for bdk 0.29.0.

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -46,6 +46,7 @@ enum BdkError {
   "Sled",
   "Rusqlite",
   "Rpc",
+  "HardenedIndex"
 };
 
 dictionary AddressInfo {
@@ -361,7 +362,7 @@ interface TxBuilder {
 interface BumpFeeTxBuilder {
   constructor(string txid, float new_fee_rate);
 
-  BumpFeeTxBuilder allow_shrinking(string address);
+  BumpFeeTxBuilder allow_shrinking(Script script_pubkey);
 
   BumpFeeTxBuilder enable_rbf();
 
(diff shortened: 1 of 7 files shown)


thunderbiscuit:
A few things to look for in this PR:

1. `Transaction.serialize()` is removed.
2. The `Address` constructor now requires a `Network` argument

@jurvis can you take a look here and make sure this does what you need it to do? In particular, the Address type is now requiring a Network when building it; does that change your workflow? (you were talking about scanning QR codes and other things). This Address type in Rust bitcoin is generic in whether its network is checked or not, i.e. you can have an `Address<NetworkUnchecked>` and an `Address<NetworkChecked>`, and in this PR we're only exposing the `NetworkChecked` version (open to comments if a slightly different API shape would work better for you).

jurvis:
> The Address constructor now requires a Network argument

This would work for how we use `Address`. However, I'd imagine, other users may prefer being able to instantiate `Address` without knowing a `network` field ahead of time. If possible, I don't think it'll hurt if we exposed an initializer that returns `Address<NetworkUnchecked>` and also expose `require_network` that will make it `Address<NetworkChecked>`. This would leave it up to users to have finer control over when they'd like the address to be validated.

> Transaction.serialize() is removed.

we only use the one for Psbt, so it's fine for us.

thunderbiscuit:
Transaction.serialize() is back in!

As for the Address type, after digging into it a bit more I realize there are a lot of ways we could go about exposing the complexity offered in the rust-bitcoin crate for that type, and I think it needs a bit more discussion before a final decision is reached. I am here implementing the solution we have in the alpha 2 release, with the addition of the `is_valid_for_network()` method.

reez (approved):
Woot! Nice job with this. 

ACK 25a48e0565376c41160e9890391a5783e6b57d39

PR #365, opened 2023-06-06
[swift] suppress keyword escape warnings

notmandatory:
### Description

This `Package.swift` config change suppresses build warnings so that users won't see these (expected) warnings when they include the `bdk-swift` package in their project:

```
/Users/steve/git/notmandatory/bdk-ffi/bdk-swift/Sources/BitcoinDevKit/BitcoinDevKit.swift:2240:13 Keyword 'index' does not need to be escaped in argument list

/Users/steve/git/notmandatory/bdk-ffi/bdk-swift/Sources/BitcoinDevKit/BitcoinDevKit.swift:2241:13 Keyword 'address' does not need to be escaped in argument list

/Users/steve/git/notmandatory/bdk-ffi/bdk-swift/Sources/BitcoinDevKit/BitcoinDevKit.swift:2242:13 Keyword 'keychain' does not need to be escaped in argument list

/Users/steve/git/notmandatory/bdk-ffi/bdk-swift/Sources/BitcoinDevKit/BitcoinDevKit.swift:2321:13 Keyword 'immature' does not need to be escaped in argument list

/Users/steve/git/notmandatory/bdk-ffi/bdk-swift/Sources/BitcoinDevKit/BitcoinDevKit.swift:2322:13 Keyword 'trustedPending' does not need to be escaped in argument list

/Users/steve/git/notmandatory/bdk-ffi/bdk-swift/Sources/BitcoinDevKit/BitcoinDevKit.swift:2323:13 Keyword 'untrustedPending' does not need to be escaped in argument list

/Users/steve/git/notmandatory/bdk-ffi/bdk-swift/Sources/BitcoinDevKit/BitcoinDevKit.swift:2324:13 Keyword 'confirmed' does not need to be escaped in argument list

etc.
```

### Notes to the reviewers

There's currently no way in Swift to suppress individual warning types.

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #449, opened 2024-01-22
chore: remove swiftsettings from bdk-swift

reez:
### Description

This removes the flag to suppress warnings on `bdk-swift`.

The motivation for this PR is to let `bdk-swift` work as a locally imported package, and work stably as a remote package.

### Notes to the reviewers

We originally added this flag in [PR 365](https://github.com/bitcoindevkit/bdk-ffi/pull/365) as the solution to the problem were `bdk-swift` users were getting many warnings in their project from `bdk-swift`. This was around the time of Xcode 13/14 and the correct solution.

However over the past few months I've seen a bunch of issues and weird behavior like in `ldk-swift` where the package would just error out, and being a blocker for anyone to even pull in the package I [opened](https://github.com/lightningdevkit/ldk-swift/issues/120) and [resolved](https://github.com/lightningdevkit/ldk-swift/commit/84d28f6ed731d1e409aa9389930e59eab950e325) it; I didn't see that extreme of behavior in `bdk-swift` but kept monitoring it.

One bad behavior we did get in `bdk-swift` was if you were to use it as a local package Xcode wouldn't let you unless you commented it out this line of code manually, which is pretty frustrating and is why `ldk-node` just removed that code in a [PR recently](https://github.com/lightningdevkit/ldk-node/pull/218#event-11238025550).

So my suggestion/solution for now its to remove it as well, and then monitor the situation to see if anything in Xcode changes that makes me feel comfortable if we put a fix in that it will remain stable.

An alternative solution is to put the line of code removed from bdk-ff `Package.swift` in the bdk-ffi `Package.swift.txt`, but at this point I’d rather wait on that since whatever change caused this in Xcode is not documented so they might keep changing things and potentially break whatever we fix.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-swift/Package.swift
+++ b/bdk-swift/Package.swift
@@ -29,8 +29,7 @@ let package = Package(
         .binaryTarget(name: "bdkFFI", path: "./bdkFFI.xcframework"),
         .target(
             name: "BitcoinDevKit",
-            dependencies: ["bdkFFI"],
-            swiftSettings: [.unsafeFlags(["-suppress-warnings"])]
+            dependencies: ["bdkFFI"]
         ),
         .testTarget(
             name: "BitcoinDevKitTests",


thunderbiscuit (approved):
ACK 6f4b4c9.

PR #456, opened 2024-02-08
Add esplora error

reez:
### Description

Adding the esplora [error](https://docs.rs/esplora-client/0.6.0/esplora_client/enum.Error.html)

### Notes to the reviewers

Added `EsploraError` for blocking client and added unit tests for each case.

Updated `full_scan` to use `EsploraError` instead of generic `Alpha3Error`.

Notable items / open questions:
- My implementation of `EsploraError` and the associated data differs slightly from the esplora_client [Error](https://docs.rs/esplora-client/0.6.0/esplora_client/enum.Error.html)
- [esplora_client](https://docs.rs/esplora-client/0.6.0/esplora_client/index.html) blocking enables [ureq](https://docs.rs/ureq/2.7.1/x86_64-unknown-linux-gnu/ureq/index.html) but async enables [reqwest](https://docs.rs/reqwest/0.11.18/x86_64-unknown-linux-gnu/reqwest/index.html)

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -27,6 +27,21 @@ interface WalletCreationError {
   LoadedNetworkDoesNotMatch(Network expected, Network? got);
 };
 
+[Error]
+interface EsploraError {
+  Ureq(string error_message);
+  UreqTransport(string error_message);
+  Http(u16 status_code);
+  Io(string error_message);
+  NoHeader();
+  Parsing(string error_message);
+  BitcoinEncoding(string error_message);
+  Hex(string error_message);
+  TransactionNotFound();
+  HeaderHeightNotFound(u32 height);
+  HeaderHashNotFound();
+};
+
 // ------------------------------------------------------------------------
 // bdk crate - types module
 // ------------------------------------------------------------------------
@@ -274,7 +289,7 @@ interface Descriptor {
 interface EsploraClient {
   constructor(string url);
 
-  [Throws=Alpha3Error]
+  [Throws=EsploraError]
   Update full_scan(Wallet wallet, u64 stop_gap, u64 parallel_requests);
 
   [Throws=Alpha3Error]
--- a/bdk-ffi/src/error.rs
+++ b/bdk-ffi/src/error.rs
@@ -1,6 +1,7 @@
 use crate::bitcoin::OutPoint;
 
 use bdk::chain::tx_graph::CalculateFeeError as BdkCalculateFeeError;
+use bdk_esplora::esplora_client::Error as BdkEsploraError;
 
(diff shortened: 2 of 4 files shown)


thunderbiscuit (approved):
Untested ACK. Thanks for working on those! Giving me an outline of how to do the persistence errors.

Just one comment: I think because your enum doesn't have the Reqwest() error, the match expression doesn't cover all cases. Did you leave it out on purpose because we don't have the async client yet?

reez:
> Untested ACK. Thanks for working on those! Giving me an outline of how to do the persistence errors.
> 
> Just one comment: I think because your enum doesn't have the Reqwest() error, the match expression doesn't cover all cases. Did you leave it out on purpose because we don't have the async client yet?

Awesome, I think I have the hang of it now and hopefully have an outline of how opening up these Errors works.

Yeah I left the Reqwest error out because we didn't have async yet, and after experimenting it felt much cleaner that way for this PR since we only support blocking atm, especially since the esplora-client crate defines errors with conditional compilation so some errors are only available when certain features are enabled, i.e. [blocking enables ureq but async enables reqwest](https://docs.rs/esplora-client/0.6.0/esplora_client/index.html#features); so when we compile with the `blocking` feature enabled but not the `async` feature the Rust compiler only considers the code paths and enum variants that are relevant for the `blocking` feature, so it covers all cases for `blocking` since it doesn't "know" about `async` Reqwest *yet*. However, once we do add async, at that point the Error will know about the reqwest error and I can loop back around on adding it (as well as was kinda planning on chatting with you on how to best handle that conditionality and get your thoughts to make sure I do that as clean+elegantly as possible, I experimented a bit with it).

reez:
_Rebased from changes merged into `master`_

Also updated error handling in `full_scan` due to Error type having `Box`:

```
/// [`esplora_client::Error`]
type Error = Box<esplora_client::Error>;
```

Let me know if you think there is a better way for me to handle it other than `.map_err(|e| EsploraError::from(*e))?;`, it felt cleaner that way for the moment so I chose to not to implement the `From` trait for `Box<bdk_esplora::esplora_client::Error>` for now (even though in the future I'm open to that or any other suggestion)

notmandatory:
> Let me know if you think there is a better way for me to handle it other than `.map_err(|e| EsploraError::from(*e))?;`, it felt cleaner that way for the moment so I chose to not to implement the `From` trait for `Box<bdk_esplora::esplora_client::Error>` for now (even though in the future I'm open to that or any other suggestion)

This way is generally fine with the `.map_err()?`, if we figure out a way to make it more concise that can be done later without breaking any APIs so not a big deal.

PR #462, opened 2024-02-28
Fix live tests

reez:
### Description

This PR addresses an issue encountered in our test suite where live tests were [failing](https://github.com/bitcoindevkit/bdk-ffi/actions/runs/8034434176).

I believe it was just due to shared state conflicts caused by multiple tests using the same database file (`bdk_persistence.db`). To resolve this I've implemented a strategy where each test case generates a unique database file for its execution. This should ensure test isolation and prevents conflicts that arise from shared database access.

### Notes to the reviewers

Each test case now generates a unique database file by appending a UUID to the `bdk_persistence` filename. This file is created in the `setUpWithError` method before each test runs.

The `tearDownWithError` method has been updated to delete the unique database file after each test completes. This ensures that no leftover state or data persists to affect subsequent tests and also helps manage filesystem usage by removing unnecessary files.

#### All Submissions:

#### New Features:

#### Bugfixes:

## Summary by CodeRabbit

- **Tests**
	- Enhanced setup and cleanup procedures in test files to improve reliability and maintain a clean test environment in BitcoinDevKit.

### DIFF
--- a/bdk-swift/Tests/BitcoinDevKitTests/LiveTxBuilderTests.swift
+++ b/bdk-swift/Tests/BitcoinDevKitTests/LiveTxBuilderTests.swift
@@ -2,12 +2,22 @@ import XCTest
 @testable import BitcoinDevKit
 
 final class LiveTxBuilderTests: XCTestCase {
+    var dbFilePath: URL!
+
+    override func setUpWithError() throws {
+        super.setUp()
+        let fileManager = FileManager.default
+        let documentDirectory = fileManager.urls(for: .documentDirectory, in: .userDomainMask).first!
+        let uniqueDbFileName = "bdk_persistence_\(UUID().uuidString).db"
+        dbFilePath = documentDirectory.appendingPathComponent(uniqueDbFileName)
+
+        if fileManager.fileExists(atPath: dbFilePath.path) {
+            try fileManager.removeItem(at: dbFilePath)
+        }
+    }
+
     override func tearDownWithError() throws {
         let fileManager = FileManager.default
-        let dbFileName = "bdk_persistence.db"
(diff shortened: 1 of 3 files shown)


thunderbiscuit (approved):
ACK. This looks good to me, but I'm wondering why the teardown works in this case but didn't previously?

Is it correct to think that the teardown happens after every function is run in the test class? I might be wrong and it only runs when the class is being torn down?

reez:
> ACK. This looks good to me, but I'm wondering why the teardown works in this case but didn't previously?
> 
> Is it correct to think that the teardown happens after every function is run in the test class? I might be wrong and it only runs when the class is being torn down?

Yeah your understanding is definitely correct, the teardown method (tearDownWithError) is called after each test method executes within the test class, not just at the end when the entire class is being torn down, which helps each test run in a relatively isolated environment by cleaning up or resetting any state that was modified during the test. The nuance ends up around the "isolation" when dealing with shared resources like databases.

So in cases where tests interact with shared resources, isolation can be a bit more squishy, and the effectiveness of `tearDownWithError` can vary based on how the resources are managed. If all tests used a single shared resource (like a `bdk_persistence.db` file) the teardown process might not have effectively isolated tests from each other, any state left in the shared resource by one test could inadvertently affect the outcome of another test, leading to flaky or failing tests, potentially.

So the fix I added here is to try to solve that by having each test interact with its own unique instance of the shared resource (in this case, a unique database file for each test), just to enhance the isolation by preventing tests from interfering with each other's state, trying to ensure it cleans up but also effectively resets the environment for each test.

thunderbiscuit:
Awesome thanks for the great explanation!

Issue #464, opened 2024-03-11
Failed to load Info.plist from bundle at path errors on Xcode 15.3


jurvis:
**Describe the bug**  
As of Xcode 15.3, it looks like `Info.plist`s for embedded third-party frameworks are enforced. 

**To Reproduce**  
Wrap `bdk-swift` in a packaged framework for iOS app access.

You should run into a build error that looks like this:
```
Failed to load Info.plist from bundle at path
```

**Expected behavior**  

Build should work successfully

**Build environment**  
 - BDK tag/commit: v0.29.0
 - OS+version: macOS Sonoma 14.3 + Xcode 15.3

reez:
I couldn't @mention you on #463 but just wanted you to get a heads up on the update [over there](https://github.com/bitcoindevkit/bdk-ffi/issues/463#issuecomment-1995685773)

reez:
This should be fixed now in bdk-swift [0.31.1](https://github.com/bitcoindevkit/bdk-swift/releases/tag/0.31.1) if you'd like to give it a test as well and let me know if anything doesn't work for ya still.

I tested 0.31.1 on [BDKSwiftExampleWallet](https://github.com/bitcoindevkit/BDKSwiftExampleWallet) using Xcode 15.3, and successfully was able to:
- build
- run
- archive
- upload to App Store Connect
- TestFlight that new archive on App Store Connect of BDKSwiftExampleWallet on my physical iOS device

Issue #463, opened 2024-03-11
Missing Info.plist in bdkFFI.framework


TadeasKriz:
**Describe the bug**  

**Xcode 15.3** now requires `Info.plist` file in `.framework` bundles. `bdkFFI.framework` brought in by `bdk-swift` doesn't have it, which results in:
- not being able to install on an iOS device or simulator
- not being able to archive and upload to AppStore Connect

**To Reproduce**  

Add `bdk-swift` package to an Xcode 15.3 project and run on simulator. You should get an error dialog about `bdkFFI.framework` not containing `Info.plist`.

**Expected behavior**  

Should install without issues.

**Build environment**  
 - BDK tag/commit: bdk-swift 0.31.0
 - OS+version: macOS 14.3 (23D56)
 - Xcode 15.3

**Additional context**  

In the meantime, a workaround is possible by creating an `Info.plist` in a `Run Script` phase in Xcode.

reez:
Thanks! Will look into this more from bdk-swift side to see best resolution path. A+ job with your description+context, appreciated @TadeasKriz !

reez:
I’m sure you all have seen a lot of large projects affected by this new Xcode 15.3 behavior, and the workarounds that folks are having to do (like just simply having to use Xcode 15.2), right now we have a [draft PR](https://github.com/bitcoindevkit/bdk-ffi/pull/466) that works for `bdk-swift` in Xcode 15.3 when building locally all the way to uploading a build to App Store Connect.

We will tighten the PR up and do some additional testing, and then plan to cut a 0.31.1 release with the proper fixes in the next week here.

If you all have any other thoughts/feedback definitely let us know in the meantime @TadeasKriz.

reez:
Tightened the PR up and did some additional testing since this Xcode 15.3 issue is quite interesting. Will add some additional comments on the [PR](https://github.com/bitcoindevkit/bdk-ffi/pull/466) directly since there are a few interesting things to note.

reez:
Have done enough testing on this, still no information still from Apple related to this Xcode 15.3 breaking change, so unless anything comes up over the next few days will plan on starting next week cutting a release for `bdk-swift` with this fix.

PR #466, opened 2024-03-13
Update bdk-swift framework info.plist for xcode 15.3 reqs

reez:
### Description

This is for pre-BDK1.0

#### All Submissions:

#### New Features:

#### Bugfixes:


notmandatory (approved):
ACK 4e9ec1c056e33fb903e97576121f7f6ac35aaa65

reez:
TODO:

- Address `macos-arm64_x86_64`'s `Info.plist`
- Rethink hardcoding version strings in `Info.plist` files
- Look at .xcframework `Info.plist` again (`macos-arm64_x86_64` section has "17.0")
- Re-look at [bdk-swift](https://github.com/bitcoindevkit/bdk-swift) repo

TadeasKriz:
Noticed the minimum iOS version is 17. Could we go as low as 15.0 (we support down to 15 in our app).

Also the macos is minimum 17, but latest version is 14.

reez:
> Noticed the minimum iOS version is 17. Could we go as low as 15.0 (we support down to 15 in our app).
> 
> Also the macos is minimum 17, but latest version is 14.

Absolutely. updated iOS min to 15.0, and macOS min to 12.0. 

And yeah good eye on that macOS min of 17, I saw it too and had it in my TODO's to get back around to it on this PR even though it "worked".

Thanks for following up with your feedback/help/thoughts!

reez:
Have made some additional changes and done some additional testing, specifically in tandem with our BDK iOS app (minimum deployment iOS 17.0). This testing included:
- building+running on Xcode 15.3
- building+running on iOS simulator
- archiving and sending a TestFlight build to App Store Connect
- running the TestFlight build on a physical iOS device

Due to the nature of this issue I've tinkered with many parameters, and here's what I set:
- using [MinimumOSVersion](https://developer.apple.com/documentation/bundleresources/information_property_list/minimumosversion) for versions relating to iOS, iPadOS, tvOS, and watchOS, and setting it to "15.0".
- using [LSMinimumSystemVersion](https://developer.apple.com/documentation/bundleresources/information_property_list/lsminimumsystemversion) for versions relating to macOS, and setting it to "12.0".
*note: using `MinimumOSVersion` in these places did not seem to break anything though.

### Notable issues/details:

As I've done testing, adding Info.plist's was important to getting things to build+run, but also having this specific item in `bdk-swift/bdkFFI.xcframework/ios-arm64/bdkFFI.framework/Info.plist` was the item that was a bit tricky in order to getting the archiving and sending to TestFlight/AppStoreConnect successfully.

When running the BDK iOS app, if I made the `bdk-swift/bdkFFI.xcframework/ios-arm64/bdkFFI.framework/Info.plist` minimum "15.0", it will build and run fine, but when archiving I'll get a “Invalid Bundle. The bundle BDKSwiftExampleWallet.app/Frameworks/bdkFFI.framework does not support the minimum OS Version specified in the Info.plist.” 

This appears to be an overly strict enforcement of the compatibility check, which doesn't take into account the implicit support for newer OS versions by frameworks that meet the app's minimum OS version requirement? To work around this, if I make the `bdk-swift/bdkFFI.xcframework/ios-arm64/bdkFFI.framework/Info.plist` minimum "100" or "20" everything works fine and I'm able to archive. I'm still not sure if this exact behavior Xcode 15.3 has is what Apple intends and there is no information about all of this in [Xcode 15.3 release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-15_3-release-notes)? It obviously doesn't make sense that you can set the minimum to "100" and it work.

### TODO:
While this PR seems to resolve the Xcode 15.3 issue, one more thing I'd like to follow up on is whether we want to be manually setting these versions (0.31.1, etc) in the plists or if we'd like to automate the way we do this in bdk-ffi/bdk-swift instead, since that could be a potential foot gun.

notmandatory:
What happens if someone with xcode 15.3 tries to use an old version of our package without this fix?

reez:
> What happens if someone with xcode 15.3 tries to use an old version of our package without this fix?

Would be the same scenario as reported [here](https://github.com/bitcoindevkit/bdk-ffi/issues/463#issuecomment-2004688034) and [here](https://github.com/bitcoindevkit/bdk-ffi/issues/464). So the error like `Failed to load Info.plist from bundle at path` and not being able to install on an iOS device or simulator & not being able to archive and upload to AppStore Connect.

Workarounds are just not using Xcode 15.3 (using Xcode 15.2, etc), or things like what TadeasKriz mentioned.

reez:
Have done enough testing on this, still no information still from Apple related to this Xcode 15.3 breaking change, so unless anything comes up over the next few days will plan on starting next week cutting a release for bdk-swift with this fix.

thunderbiscuit:
@reez let me know when this is ready to merge in your opinion. I read the changes and I'm happy to hear it fixes the issue but will leave it to you for the final word on it.

reez:
> @reez let me know when this is ready to merge in your opinion. I read the changes and I'm happy to hear it fixes the issue but will leave it to you for the final word on it.

@thunderbiscuit I believe this is ready to merge into a branch to create a 0.31.1 release for bdk-swift

PR #468, opened 2024-03-21
Bump bdk to alpha 7

thunderbiscuit:
This PR:
1. Bumps the Rust bdk library to its latest version, alpha 7, the esplora client to 0.9.0, and the file store to 0.7.0.
~2. Finally closes a long-standing comment/issue where we could not use rustls for the async esplora client, preventing us from building for Android. The commit enabling the feature simply enables it, but does not build the async client itself.~ This is currently causing runtime issues. Will fix in a further PR.

### Notes to the reviewers
This bump was fairly simple.

#### All Submissions:

### DIFF
--- a/bdk-ffi/Cargo.lock
+++ b/bdk-ffi/Cargo.lock
@@ -138,9 +138,9 @@ dependencies = [
 
 [[package]]
 name = "bdk"
-version = "1.0.0-alpha.6"
+version = "1.0.0-alpha.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4d2d7e0429201a21a4fcd586995b8dd452a15d7c31e88a68a978441a1c17dc8f"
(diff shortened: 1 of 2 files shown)


thunderbiscuit:
The failure of the Swift tests _maybe_ look like what is referenced in [this issue](https://github.com/mozilla/uniffi-rs/issues/1291).

If so, it might mean that it's an architecture mix issue. The CI runs on x86_64, but it somehow build for arm64? I know the tests run cleanly on my local M1.

reez:
ACK bf3942154f5536f6598a2774d3f517649a20b484

changing from `macos14` to `macos12` did the trick?

thunderbiscuit:
No changing it didn't solve it unfortunately (but makes more sense so I'm happy it didn't).

We're still getting linker errors for iOS when attempting to use the reqwest dependency.

reez (approved):

PR #471, opened 2024-03-26
Enable using swift local build script from bdk-swift directory

### LINKED ISSUE #423: Swift build scripts should be callable from the bdk-swift directory
This would bring them in line with the other 3 libraries.

thunderbiscuit:
This PR fixes #423.

Our local build script for Swift needed to be called from the root directory, unlike the build workflows for Kotlin and Python. This PR fixes that.

One thing to check: does that impact the bdk-swift release workflow? Should confirm because I think the CI workflows might need to be adjusted there as well.

### DIFF
--- a/bdk-swift/build-local-swift.sh
+++ b/bdk-swift/build-local-swift.sh
@@ -3,18 +3,15 @@
 # The results of this script can be used for locally testing your SPM package adding a local package
 # to your application pointing at the bdk-swift directory.
 
-# Run the script from the repo root directory, ie: ./bdk-swift/build-local-swift.sh
-
 rustup install 1.73.0
 rustup component add rust-src
-rustup target add aarch64-apple-ios      # iOS ARM64
+rustup target add aarch64-apple-ios      # iOS arm64
 rustup target add x86_64-apple-ios       # iOS x86_64
 rustup target add aarch64-apple-ios-sim  # simulator mac M1
 rustup target add aarch64-apple-darwin   # mac M1
 rustup target add x86_64-apple-darwin    # mac x86_64
 
-pushd bdk-ffi
-mkdir -p Sources/BitcoinDevKit
+cd ../bdk-ffi/ || exit
 cargo run --bin uniffi-bindgen generate src/bdk.udl --language swift --out-dir ../bdk-swift/Sources/BitcoinDevKit --no-format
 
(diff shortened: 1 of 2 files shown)


bdk-swift/build-local-swift.sh
```diff
@@ -13,8 +11,8 @@ rustup target add aarch64-apple-ios-sim  # simulator mac M1
 rustup target add aarch64-apple-darwin   # mac M1
 rustup target add x86_64-apple-darwin    # mac x86_64
 
-pushd bdk-ffi
-mkdir -p Sources/BitcoinDevKit
+cd ../bdk-ffi/
+# mkdir -p temp/Sources/BitcoinDevKit
```
thunderbiscuit, line 15: I don't think this line did anything because the directory is not referenced anywhere after.

bdk-swift/build-local-swift.sh
```diff
@@ -40,4 +37,4 @@ cp ../bdk-ffi/target/lipo-macos/release-smaller/libbdkffi.a bdkFFI.xcframework/m
 rm Sources/BitcoinDevKit/bdkFFI.h
 rm Sources/BitcoinDevKit/bdkFFI.modulemap
 #rm bdkFFI.xcframework.zip || true
-#zip -9 -r bdkFFI.xcframework.zip bdkFFI.xcframework
\ No newline at end of file
+#zip -9 -r bdkFFI.xcframework.zip bdkFFI.xcframework
```
thunderbiscuit, line 40: Are there still useful?

reez:
> This PR fixes #423.
> 
> Our local build script for Swift needed to be called from the root directory, unlike the build workflows for Kotlin and Python. This PR fixes that.
> 
> One thing to check: does that impact the bdk-swift release workflow? Should confirm because I think the CI workflows might need to be adjusted there as well.

ACK 10241e6b2159ea94a8054b0ad28df240cf3ab867

Tested locally.

Re "One thing to check: does that impact the bdk-swift release workflow? Should confirm because I think the CI workflows might need to be adjusted there as well." I think you+I chatted about this since then about how it shouldn't impact that workflow from my recollection. When looking at that workflow I don't see that it uses the `./build-local-swift.sh` anywhere in it so I feel like that's why maybe we chatted about how it wouldn't impact it.

reez (approved):

PR #480, opened 2024-03-28
Gradle updates

thunderbiscuit:
This PR is a big update to the Gradle plugins used in the Android and JVM libraries. It also updates the Kotlin version to 1.9.23 from a now quite old version (1.6.10).

It also uses the now recommended plugin block structure. We haven't touched these gradle build scripts in a long time so they were overdue for a bit of a refresh. They also address a few of the warnings we're getting when building and releasing.

Note: merge after #479.


reez:
Whats the best/smartest way for me to test this out before ACK?

thunderbiscuit:
You could just build both libraries, which will verify most of the changes. The only potential sticky place is with the version of Kotlin, which you might not have installed on your local machine. But I think since you have Android Studio it might actually be there anyway... not sure.

Try
```shell
cd bdk-jvm
just build

cd ../bdk-android/
just build
```

And let me know if it compiles for you.

reez:
> You could just build both libraries, which will verify most of the changes. The only potential sticky place is with the version of Kotlin, which you might not have installed on your local machine. But I think since you have Android Studio it might actually be there anyway... not sure.
> 
> Try
> 
> ```shell
> cd bdk-jvm/
> just build
> 
> cd ../bdk-android/
> just build
> ```
> 
> And let me know if it compiles for you.

==

`kotlin -version`
Kotlin version 1.9.23-release-779 (JRE 21.0.2)

==

```shell
cd bdk-jvm/
just build
```
BUILD SUCCESSFUL in 1m 34s

==

```shell
cd ../bdk-android/
just build
```
*I do have android studio and the emulator open*
```
./gradlew buildAndroidLib

FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'bdk-android'.
> Could not determine the dependencies of task ':plugins:compileKotlin'.
   > Unknown Kotlin JVM target: 21

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

BUILD FAILED in 443ms
error: Recipe `build` failed on line 8 with exit code 1
```

I'll look into this build failure for me on bdk-android

thunderbiscuit:
Weird... I don't know why it's expecting a Kotlin JVM target of 21...

reez (approved):
ACK 742c56d8d4d8cd269954aecaec52eb4e29abb6b2

Tested locally.

PR #481, opened 2024-03-28
Release for Python 3.11 and 3.12

thunderbiscuit:
Just trying to see if it passes the test workflows...

### DIFF
--- a/.github/workflows/publish-python.yaml
+++ b/.github/workflows/publish-python.yaml
@@ -24,6 +24,8 @@ jobs:
           - cp38-cp38
           - cp39-cp39
           - cp310-cp310
+          - cp311-cp311
+          - cp312-cp312
     steps:
       - name: "Checkout"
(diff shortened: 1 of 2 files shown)


reez:
ACK

My only question is since the PR touches Python+JVM+Android (and I see this follow on PR about JVM+Android #480 ) I was wondering how you were going to approach the commit history? Since it would be nice to have the Python related ones in one PR and all the JVM+Android ones in the other PR (so like separate the PR's starting with the commit related to `build: use jdk 17 for android and jvm builds`)

reez:
> Yes I had in mind of merging #480 first, then rebasing this one on top of it, which would then reduce this PR to only the [dec1ca5](https://github.com/bitcoindevkit/bdk-ffi/commit/dec1ca53110cfde58aaa37fb8a0919f342339d69) commit if I do my things correctly I think.

great!

PR #482, opened 2024-04-01
Bump bdk to alpha 8

thunderbiscuit:
This PR upgrades bdk to alpha 8 as well as bdk_esplora to 0.10.0 and bdk_file_store to 0.8.0.

It required the FeeRate object to be fiddled with a little bit because bdk now uses the bitcoin::FeeRate, which uses u64s instead of f32 in most places. The names of the methods are also changed a little bit, and the `from_sat_per_vb` constructor actually returns None when the input produces an arithmetic overflow. I decided to return an error instead, to keep the struct simple (having a None in there would require the FeeRate object to be different in nature than what we currently have, and I think an error is appropriate in this case anyway). Happy to hear questions/thoughts on this, because it's a slight deviation from the Rust API.

Merge after 471, 479, 480 and 481.


reez (approved):
ACK aa035588a0c2f1c5a81770f6f0f2898b9d679bfe

PR #483, opened 2024-04-02
Enable Esplora async client

thunderbiscuit:
This is just to open up the work on async Esplora.

### DIFF
--- a/bdk-ffi/Cargo.lock
+++ b/bdk-ffi/Cargo.lock
@@ -3,6 +3,15 @@
 version = 3
 
 [[package]]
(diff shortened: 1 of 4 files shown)


reez:
Do you want to close this for now (since I think it was for experimentation purposes?) and then we can bring it back from the dead or use it as a reference or open up a new PR (rebasing this current PR might stink) when we are ready to do async?

Definitely glad you experimented with it, I'm sure I'll reference it in the future.

thunderbiscuit:
Yeah let's close this for now and stay lean on the PRs.

PR #484, opened 2024-04-03
Add test for `WalletCreationError`

reez:
### Description

Originally had this in #460, but once I created a new error type instead of using `WalletCreationError`, it made sense to not have this test in that PR since it wasn't directly related to my changes in it anymore. 

But I do like this test being included in `bdk-ffi` still so I'm just doing a small follow up PR to add it since it didn't make sense being included in the final version of #460.

#### All Submissions:

#### New Features:

#### Bugfixes:


thunderbiscuit (approved):
ACK a8541ecd400d98c6df464b4320fd30779e410bf8.

PR #488, opened 2024-04-10
Expose new methods on wallet

thunderbiscuit:
This PR exposes:
1. The [get_tx](https://docs.rs/bdk/1.0.0-alpha.7/bdk/wallet/struct.Wallet.html#method.get_tx) method on the wallet. It was an easy one because it's basically the "single tx" version of the `transactions()` method, which returns a vector of transactions.
2. The [list_unspent()](https://docs.rs/bdk/1.0.0-alpha.6/bdk/wallet/struct.Wallet.html#method.list_unspent) method. This is useful for coin selection, and would be used with the TxBuilder.add_utxo() method. We already had all the types required to expose this, hence the small diff. Easy win.
3. The [list_output()](https://docs.rs/bdk/1.0.0-alpha.6/bdk/wallet/struct.Wallet.html#method.list_output) method. Useful for showing history and just general querying of the wallet. Same as (2), we had everything we needed for it so it was an easy one to add.

One difference between this API and the Rust API is that with the `get_tx()` method, we must combine the parsing of the Txid type and the get_tx method, which means that the method can now throw (if the txid string cannot be parsed into the rust-bitcoin Txid type), whereas the bdk method does not throw. Another solution to this would be to expose the rust-bitcoin Txid type directly, but I'm not sure we'd have much use for it other than externalizing this throwing of the error... let me know what you think. If we used it in a few places it might make sense to expose the type completely; just not sure it's useful now.

### Changelog notice

```md
Added:
  - New get_tx method on the Wallet type [#488]
  - New list_output method on the Wallet type [#488]
  - New list_unspent method on the Wallet type [#488]

[#488]: https://github.com/bitcoindevkit/bdk-ffi/pull/488
```

#### All Submissions:

## Summary by CodeRabbit

- **New Features**
	- Enhanced transaction retrieval with improved error handling for parsing transaction IDs, ensuring better reliability and user feedback.

### DIFF
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -104,6 +104,11 @@ interface DescriptorError {
     Hex(string e);
 };
 
+[Error]
+interface TxidParseError {
+  InvalidTxid(string txid);
+};
+
 // ------------------------------------------------------------------------
 // bdk crate - types module
 // ------------------------------------------------------------------------
@@ -211,6 +216,9 @@ interface Wallet {
   SentAndReceivedValues sent_and_received([ByRef] Transaction tx);
 
   sequence<CanonicalTx> transactions();
+
+  [Throws=TxidParseError]
+  CanonicalTx? get_tx(string txid);
 
   [Throws=CalculateFeeError]
   u64 calculate_fee([ByRef] Transaction tx);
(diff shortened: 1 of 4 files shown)


reez (approved):
ACK c7f9f8608e7299bb0d0f394ce29db684a15ff43f

Yup these are good additions and nice wins, and I agree with you're thoughts around `get_tx()` throwing.

PR #499, opened 2024-04-19
Add weight method on transaction type

thunderbiscuit:
The new `weight()` method returns a `Weight` type, which might be interesting to expose at some point, but I figured in this case it was useful to have at least what we had before (the method just returning a u64), so I pulled that weight out of the weight type and return it raw.

Easy win.


reez (approved):
ACK 2c2667f08786230cfdec20c040c7523f11ff29e0

win!

PR #503, opened 2024-04-19
Bump Rust compiler to 1.77.1

thunderbiscuit:
This PR is a cherry pick of the commits I needed to make in order to make the alpha 9 release work.

It does the following:
- Upgrades all Rust compiler versions for builds, tests, CI, documentation to 1.77.1. 1.77 is the stable version, and 1.77.1 is the version Mozilla currently uses in their CI.
- Fixes some new clippy warnings that showed up because of the version bump.
- Updates the cargo lockfile with the latest versions of the dependencies.

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - version: 1.73.0
+          - version: 1.77.1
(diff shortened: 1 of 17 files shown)


reez (approved):
ACK 98f7f0ebd271d4a1dc0a018846c4cbb9f5073392

Tested locally, thanks for doing this update-

PR #512, opened 2024-04-24
Move feerate type to bitcoin module

### LINKED ISSUE #510: Refactor: move FeeRate type to bitcoin module
BDK now uses the FeeRate type from rust bitcoin, but we didn't move it to its correct location when updating.

thunderbiscuit:
Fixes #510.


reez (approved):
ACK 4c48883fc83fe0fbedacfa4a6a3144fb7d1ef26c

Nice!

Issue #509, opened 2024-04-23
Exposing Errors 


thunderbiscuit:
It turns out that errors are not easy to move from Rust to the target languages when using uniffi. In short, it looks like one must choose between a range of options, each with pros and cons, and none particularly perfect. This issue is an attempt at quantifying and categorizing these options and the choices we made for the bdk-ffi language bindings.

thunderbiscuit:
## Option 1
One requirement you might have for errors is that their fields be exposed to the bindings. For example, the `TxidParseError` would like to return to the user not only it's variant's name, but the txid that was not parsable:
```rust
#[derive(Debug, thiserror::Error)]
pub enum TxidParseError {
    #[error("invalid txid: {txid}")]
    InvalidTxid { txid: String },
}
```

In the UDL, this shows up as:
```txt
[Error]
interface TxidParseError {
  InvalidTxid(string txid);
};
```

When building bindings code for this however, the resulting exception thrown will not use the message you might expect (the one built by `thiserror`). Instead, a custom `message` field is built which simply prints the names of the fields in the enum and their value:
```kotlin
sealed class TxidParseException: Exception() {
    class InvalidTxid(val txid: String) : TxidParseException() {
        override val message
            get() = "txid=${ `txid` }"
}
```

This means that we cannot customize the message given to the user; the `thiserror` library does not help us, and following this approach, one must choose between a custom error message and the fields one might want to expose. For example, instead of a message like
```txt
invalid HTTP header value: {value}
```
You only send back to the Koltin/Swift user an exception with a field `message`, which would print as
```txt
value=${value}
```

This is inconvenient, and overall doesn't provide the sort of information one might want to convey. We really want to customize our error messages, but we also need to bring in our data! This leads us to option 2.

## Option 2

reez:
This is an awesome breakdown. 1 or 3 definitely are our best bets.

thunderbiscuit:
So it turns out there is a 4th option. If you use a struct as an error type instead of an enum, your struct can implement the Display trait which will populate the `toString()` method in Kotlin and Swift, and also contain the full typed data inside. The [Mozilla test-fixtures have a good example of this](https://github.com/mozilla/uniffi-rs/tree/main/fixtures/error-types).

This is maybe less applicable for us because it means your errors are not variants of an enum and are simply objects by themselves. It turns out you can combine them into an enum if you want I think, but then you loose the nice Display features.

[...]

PR #513, opened 2024-04-26
Add ADR document on errors

thunderbiscuit:
This PR is very much up for review and discussion. I figured that a good way to lay out correctly a solution would be to write it out. We can hash out the specifics directly on this PR.

### DIFF
--- a/docs/adr/03-errrors.md
+++ b/docs/adr/03-errrors.md
@@ -1,0 +1,59 @@
+# Errors
+
+Returning meaningful errors is important for users of our libraries. The Rust bitcoin ecosystem uses descriptive and data-driven errors, and we would like to stay as close to them as possible when building language bindings for these libraries.
+
+## Context and Problem Statement
+
+The tool we use to produce language bindings for bdk-ffi libraries is [uniffi]. While the library is powerful, it also comes with some caveats. Those come into play when attempting to expose Rust errors; we must choose between simple enums that have variants but not data associated with them (for example a `TxidError` would not be able to return the specific txid that triggered the error), or more complex objects that do have data fields, but no ability to control the error message returned to the user. What's more, while Kotlin and Java users expect a `message` field on an `Exception` type, this field becomes an empty string if the object has no data, and just a strignified version of the fields if it does have any.
+
+## Decision Drivers
+
+Some options were considered and explored in detail in [issue #509]. Some aspects of this decisions include:
+- Expectations from devs using the libraries in different languages (for example Kotlin users expect a `message` field on the exception, whereas Swift users do not)
+- Ease of maintaining the errors as they evolve and we expose more and more of the Rust BDK ecosystem
+- Important vs nice-to-have features (some fields can be stringified without loss of information, some cannot)
+
+## Decision Outcome
+
+We have decided to leverage two different approaches that uniffi offers for exposing errors, and using the most appropriate one for each situation.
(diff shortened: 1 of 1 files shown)


reez:
Great idea. I'm going to ruminate on this a bit over the next week (and add comments/thoughts as needed)

docs/adr/03-errrors.md
```diff
@@ -0,0 +1,59 @@
+# Errors
+
+Returning meaningful errors is important for users of our libraries. The Rust bitcoin ecosystem uses descriptive and data-driven errors, and we would like to stay as close to them as possible when building language bindings for these libraries.
+
+## Context and Problem Statement
+
+The tool we use to produce language bindings for bdk-ffi libraries is [uniffi]. While the library is powerful, it also comes with some caveats. Those come into play when attempting to expose Rust errors; we must choose between simple enums that have variants but not data associated with them (for example a `TxidError` would not be able to return the specific txid that triggered the error), or more complex objects that do have data fields, but no ability to control the error message returned to the user. What's more, while Kotlin and Java users expect a `message` field on an `Exception` type, this field becomes an empty string if the object has no data, and just a strignified version of the fields if it does have any.
```
reez, line 7: if you want to touch on swift real quick here we can add something like this at the end of the `Context and Problem Statement` section:

"Swift, in contrast, can optionally leverage something like the `LocalizedError` protocol to provide custom and localized descriptions."

docs/adr/03-errrors.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Decision Drivers
+
+Some options were considered and explored in detail in [issue #509]. Some aspects of this decisions include:
+- Expectations from devs using the libraries in different languages (for example Kotlin users expect a `message` field on the exception, whereas Swift users do not)
```
reez, line 12: "whereas Swift users do not" could be "whereas Swift users do not expect `message` but might expect `localizedDescription`"

reez (approved):
ACK c198a3476feb0c3bfdd1a583841757265dee479c

Added 2 comments you can optionally include, and then coderabbitai added some good grammatical suggestions I think should be included.

docs/adr/03-errrors.md
reez: verbiage on this suggestion might change now with https://github.com/mozilla/uniffi-rs/pull/2116

reez:
Still interested in adding this as an ADR? I think we are solidified enough per #509 that the way we do it now is the "standard", so I'm cool with getting this merged if you are good with that (obviously after rebase, maybe fixing small nits/comments, etc).

docs/adr/03-errrors.md
thunderbiscuit: Done.

docs/adr/03-errrors.md
thunderbiscuit: Done.

docs/adr/03-errrors.md
```diff
@@ -0,0 +1,59 @@
+# Errors
+
+Returning meaningful errors is important for users of our libraries. The Rust bitcoin ecosystem uses descriptive and data-driven errors, and we would like to stay as close to them as possible when building language bindings for these libraries.
```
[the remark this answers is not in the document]
thunderbiscuit: Looks good!

PR #518, opened 2024-04-29
Use Rust compiler 1.77.1 on python ci runs

### LINKED ISSUE #404: Python CI workflows use actions-rs/toolchain@v1 but the scripts define the Rust compiler version explicitly
We should streamline how we use and define what Rust compiler version is being used in all CI workflows. The Python workflows set the version differently than the Android/JVM/Swift workflows.

thunderbiscuit:
This PR initially tried to remove the "unused" action toolchain@v1. This turned out not to be true, because while the exact Rust compiler version is now defined in the [scripts](https://github.com/bitcoindevkit/bdk-ffi/tree/master/bdk-python/scripts), Rust must be installed already for those steps to have access to rustup in the first place.

The one improvement done here is that the correct version of Rust is downloaded in the first place on the CI, so we don't download the sable version only to then download a second version when the shell script starts. We save one round of downloads. Note that the rustup command to use 1.77.1 must stay in the shell scripts for people building locally.

Fixes #404

### DIFF
--- a/.github/workflows/live-tests.yaml
+++ b/.github/workflows/live-tests.yaml
@@ -74,9 +74,11 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: true
-      - uses: actions-rs/toolchain@v1
+
+      - name: "Install Rust 1.77.1"
+        uses: actions-rs/toolchain@v1
         with:
-          toolchain: stable
+          toolchain: 1.77.1
 
       - name: "Generate bdk.py and binaries"
         run: bash ./scripts/generate-linux.sh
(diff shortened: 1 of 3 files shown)


thunderbiscuit:
The workflow that tests these changes [is here](https://github.com/bitcoindevkit/bdk-ffi/actions/runs/8884141178)

thunderbiscuit:
Example test run of this PR triggered manually here: https://github.com/bitcoindevkit/bdk-ffi/actions/runs/9033520723

reez (approved):
ACK ac15ed7380266e52c9ef50559e38738a68a8fba9

PR #519, opened 2024-04-29
Set rust compiler version before running cargo commands in python

### LINKED ISSUE #404: Python CI workflows use actions-rs/toolchain@v1 but the scripts define the Rust compiler version explicitly

thunderbiscuit:
Small issue I discovered while attempting to fix #404.

The Rust compiler version was defined after running the first cargo command, whereas it should of course be set before, to ensure all cargo commands are run with the same compiler version.


reez (approved):
ACK 9a61229dc21fcebf2e355a45526f73d002196e1a

LGTM!

Issue #522, opened 2024-04-30
Still need to be brought back for 1.0


thunderbiscuit:
This issue compiles a list of structs and methods that were commented out in the move to the 1.0 early releases in order to keep track of what we probably want to bring back, without having to keep them as comments in the codebase.

- [x] TxIn type
- [x] Payload type (doesn't exist anymore)
- [x] Address.from_script() method
- [ ] Address.payload() method
- [x] Transaction.lock_time()
- [x] Transaction.input()
- [x] Transaction.output()
- [x] Psbt.txid() (not there anymore)
- [x] Psbt.combine()
- [x] Psbt.fee_amount() (renamed to fee())
- [x] Psbt.fee_rate() (not there anymore)
- [x] Psbt.json_serialize()
- [ ] SignOptions
- [x] TxBuilder.add_data()

jurvis:
hi @thunderbiscuit, we still need [Transaction – input() and output()](https://github.com/proto-at-block/bitkey/blob/1f9f2298919dac77e6791aa3f1dbfd67efe7f83c/app/ios/Wallet/Sources/BdkBindings/BdkTransactionImpl.swift#L32-L38) and conversely from 1, the `TxIn` type.

does `.finalize_psbt` not require SignOptions as a parameter anymore?

thunderbiscuit:
We didn't have the Wallet.finalize_psbt() method before (or at least I can't find it anywhere), but you might be referring to the Wallet.sign() method? The `sign()` method still takes a `SignOptions` parameter in Rust, but for the bindings we simply pass the default version of that (and have removed it as a parameter). Did you use it to customize the signer options? I'm happy to try to add it back for you next week if you'd like and need it.

andreasgriffin:
`get_height` (the current blockchain tip)  is missing. It also seems to have vanished from the esplora and electrum client. This is however very useful.

andreasgriffin:
The RPC client would be great!

reez:
Note: Have a user on Discord that might need `SignOptions` back, just making a note of it here and I'll follow up with more information about their specific use case and if that satisfies what they might need.

reez:
<img width="867" alt="Image" src="https://github.com/user-attachments/assets/341f660b-de2a-4002-a97d-7919fde79107" />

The last two items left are complete now I believe, I just added a replacement for `Address.payload()` and thunder added `SignOptions` https://github.com/bitcoindevkit/bdk-ffi/pull/650

[...]

PR #523, opened 2024-04-30
Remove pre-1.0 code

thunderbiscuit:
Instead of closing an issue, this PR opens one: #522.

I just thought I'd be cleaner to keep track of what we want to bring back in an issue (or maybe multiple issues?) rather than just old commented out code that's getting very stale at this point. The basic idea is that every codeblock that is removed here gets an associated todo item in #522.

Let me know what you think.

## Summary by CodeRabbit

- **Refactor**
  - Streamlined wallet functionalities by removing unused methods related to address and transaction handling.
  - Enhanced transaction builder by modifying the finish method and removing redundant data addition capabilities.
  - Removed outdated comments and unused code across various modules to improve code clarity and maintenance.


reez (approved):
ACK 5e8271e15804b0fca2cf50ff288bac5848dbdd6f

Awesome

PR #445, opened 2024-01-16
(WIP) Add sync to esplora client

reez:
### Description

_Depends on [bdk PR 1194](https://github.com/bitcoindevkit/bdk/pull/1194) so is currently pointed at that fork branch_

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:

* [ ] I've added tests for the new feature
* [ ] I've added docs for the new feature

#### Bugfixes:

* [ ] This pull request breaks the existing API
* [ ] I've added tests to reproduce the issue which are now passing
* [ ] I'm linking the issue being fixed by this PR

PR #526, opened 2024-05-03
Bump to BDK alpha 10

thunderbiscuit:
temp

### DIFF
--- a/bdk-ffi/src/esplora.rs
+++ b/bdk-ffi/src/esplora.rs
@@ -1,13 +1,16 @@
+use std::ops::{Deref};
 use crate::error::EsploraError;
-use crate::wallet::{Update, Wallet};
-
+use crate::wallet::{Update};
 use crate::bitcoin::Transaction;
+use crate::types::FullScanRequest;
+
 use bdk::bitcoin::Transaction as BdkTransaction;
-use bdk::wallet::Update as BdkUpdate;
+use bdk::chain::spk_client::FullScanResult as BdkFullScanResult;
 use bdk_esplora::esplora_client::{BlockingClient, Builder};
 use bdk_esplora::EsploraExt;
 
 use std::sync::Arc;
+use bdk::KeychainKind;
 
 pub struct EsploraClient(BlockingClient);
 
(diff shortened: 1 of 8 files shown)


thunderbiscuit:
@reez @notmandatory I ended up adding 2 new variants on the EsploraError enum, `FullScanRequestAlreadyConsumed` and `SyncRequestAlreadyConsumed`. I'm using them like so:

```rust
        let request: BdkSyncRequest = request
            .0
            .lock()
            .unwrap()
            .take()
            .ok_or(EsploraError::SyncRequestAlreadyConsumed)?;
```

Let me know what you think!

notmandatory:
> @reez @notmandatory I ended up adding 2 new variants on the EsploraError enum, `FullScanRequestAlreadyConsumed` and `SyncRequestAlreadyConsumed`. 

⛏️ Only nit I have is to consolidate into one error with a slightly shorter name like `RequestAlreadyUsed`. But if you'd rather keep two errors and the longer names it's fine with me too.

reez:
> > @reez @notmandatory I ended up adding 2 new variants on the EsploraError enum, `FullScanRequestAlreadyConsumed` and `SyncRequestAlreadyConsumed`.
> 
> ⛏️ Only nit I have is to consolidate into one error with a slightly shorter name like `RequestAlreadyUsed`. But if you'd rather keep two errors and the longer names it's fine with me too.

I like this idea of consolidating 👍

reez (changes requested):
I think we need some changes in terms of adding/exposing `commit` in this PR, and then using `apply_update` and `commit` after we create a `Request` and `Update`. 


Since this spans a few different files in a few different places I didn't add it as a comment to each spot, but for example after:

```
val fullScanRequest: FullScanRequest = wallet.startFullScan()
val update = esploraClient.fullScan(fullScanRequest, 10uL, 1uL)
```

... we need to then [apply the update](https://docs.rs/bdk/1.0.0-alpha.7/bdk/wallet/struct.Wallet.html#method.apply_update) and then [commit](https://docs.rs/bdk/1.0.0-alpha.7/bdk/wallet/struct.Wallet.html#method.commit)

Let me know if any of that is unclear, or any thoughts or comments on that.

bdk-ffi/src/esplora.rs
```diff
@@ -17,41 +23,57 @@ impl EsploraClient {
...
         parallel_requests: u64,
     ) -> Result<Arc<Update>, EsploraError> {
-        let wallet = wallet.get_wallet();
+        // using option and take is not ideal but the only way to take full ownership of the request
+        let request: BdkFullScanRequest<KeychainKind> = request
+            .0
+            .lock()
+            .unwrap()
+            .take()
+            .ok_or(EsploraError::RequestAlreadyConsumed)?;
```
notmandatory, line 38: ❤️ good to add a new error for this.

bdk-ffi/src/esplora.rs
```diff
@@ -17,41 +23,57 @@ impl EsploraClient {
...
+    ) -> Result<Arc<Update>, EsploraError> {
+        // using option and take is not ideal but the only way to take full ownership of the request
+        let request: BdkSyncRequest = request
             .0
-            .update_local_chain(previous_tip, missing_heights)
-            .map_err(|e| EsploraError::from(*e))?;
+            .lock()
+            .unwrap()
+            .take()
+            .ok_or(EsploraError::RequestAlreadyConsumed)?;
```
reez, line 64: Can be handled in a follow up PR, but what did ya'll think was the best way to handle this instead of using `unwrap()` which would panic?

Some idiomatic approaches I know of are:
- handle the option explicitly
- use match
- use if let

Handling the option explicitly could be something like:
```
let request = request
    .0
    .lock()
    .ok_or(EsploraError::LockError)?
    .take()
    .ok_or(EsploraError::RequestAlreadyConsumed)?;
```
... but one problem with that is we don't have an `EsploraError::LockError` and I'm not sure that error even fits as an EsploraError necessarily so if it was a different type we'd have to map it to an EsploraError since thats our return type, etc.

Thoughts on what is most idiomatic and what we use in our codebase that feels best?

thunderbiscuit: We have mutexes in a few other places, and at the moment we are using unwraps on all the lock methods. See [psbt.extract_tx()](https://github.com/bitcoindevkit/bdk-ffi/blob/7d951578d0b1b40a3f6f5a23b98166f72ae3845b/bdk-ffi/src/bitcoin.rs#L168-L172) and [DescriptorSecretKey.derive()](https://github.com/bitcoindevkit/bdk-ffi/blob/7d951578d0b1b40a3f6f5a23b98166f72ae3845b/bdk-ffi/src/keys.rs#L92).

It might be good to define a proper error for those, but I do imagine that would be maybe not within the scope of this update.

reez: > We have mutexes in a few other places, and at the moment we are using unwraps on all the lock methods. See [psbt.extract_tx()](https://github.com/bitcoindevkit/bdk-ffi/blob/7d951578d0b1b40a3f6f5a23b98166f72ae3845b/bdk-ffi/src/bitcoin.rs#L168-L172) and [DescriptorSecretKey.derive()](https://github.com/bitcoindevkit/bdk-ffi/blob/7d951578d0b1b40a3f6f5a23b98166f72ae3845b/bdk-ffi/src/keys.rs#L92).
> 
> It might be good to define a proper error for those, but I do imagine that would be maybe not within the scope of this update. Just my 2 cents.

Yeah, I mentioned it can be handled in a follow up PR because I wasn't wanting it to hold up anything about this PR (I'm ready to use all the goodies in this PR, heh), so I'm with ya there for sure. And thanks for calling out those other places for me like `extract_tx()` and `derive`

reez (approved):
ACK e9a76287c8bc6012fe27d1531ffbda3554239a46

I believe the related PR's can also be closed when this is merged:
- #445 
- #461 

🥳

PR #530, opened 2024-05-08
Add memory wallet

thunderbiscuit:
I almost can't believe we get this for free now that the generic on the Wallet type has been removed. 

Don't merge, needs actual testing. My main concern would be that certain methods don't work on the memory wallet, but that should be captured at compile time, so... are we just getting away with it with such simple changes? It's too good to be true.

### Changelog notice

```md
Added
  - New memory wallet [#528]

[#528]: https://github.com/bitcoindevkit/bdk-ffi/pull/528
```

#### All Submissions:

### DIFF
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -271,6 +271,9 @@ interface Wallet {
   [Throws=WalletCreationError]
   constructor(Descriptor descriptor, Descriptor? change_descriptor, string persistence_backend_path, Network network);
 
+  [Name=new_no_persist, Throws=DescriptorError]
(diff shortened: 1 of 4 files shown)


thunderbiscuit:
@reez I have tested those locally with all our JVM tests by turning them into memory wallets and all tests passed.

thunderbiscuit:
@reez this is ready for final review. I rebased #535 onto this one, so if you can review this one first we'll merge that and then look at the electrum client.

reez (approved):
ACK adc1ad8f01d222239de3b4d2294885667ef68091

Good to merge when new Swift test ✅ on CI

PR #531, opened 2024-05-10
Use signet for live tests

thunderbiscuit:
This PR moves the live tests from testnet to signet. Merge after #518.

Note that the tests below do not actually run the live tests. The runs for the live tests triggered manually is [here](https://github.com/bitcoindevkit/bdk-ffi/actions/runs/9034091785). Note: the balance on the wallet was 0. The new run is [here](https://github.com/bitcoindevkit/bdk-ffi/actions/runs/9036407306).


reez (approved):
ACK f27bada9c90baaed610436aed6a5c0f25b5b42bc

PR #532, opened 2024-05-10
Bring back PSBT methods

thunderbiscuit:
This required exposing the whole PSBT Error (32 variants!). Draft for now.

### DIFF
--- a/bdk-ffi/Cargo.toml
+++ b/bdk-ffi/Cargo.toml
@@ -22,11 +22,11 @@ bdk = { version = "1.0.0-alpha.11", features = ["all-keys", "keys-bip39"] }
 bdk_esplora = { version = "0.13.0", default-features = false, features = ["std", "blocking", "blocking-https-rustls"] }
 bdk_electrum = { version = "0.13.0" }
 bdk_file_store = { version = "0.11.0" }
+bdk_bitcoind_rpc = { version = "0.10.0" }
 
 uniffi = { version = "=0.27.1" }
 bitcoin-internals = { version = "0.2.0", features = ["alloc"] }
 thiserror = "1.0.58"
-serde_json = "1.0.116"
 
 [build-dependencies]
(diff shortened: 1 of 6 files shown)


bdk-ffi/Cargo.toml
```diff
@@ -25,6 +25,7 @@ bdk_file_store = { version = "0.10.0" }
 uniffi = { version = "=0.26.1" }
 bitcoin-internals = { version = "0.2.0", features = ["alloc"] }
 thiserror = "1.0.58"
+serde_json = "1.0.116"
```
reez, line 28: I think instead of adding `serde_json` directly here we can add `bdk_bitcoind_rpc = { version = "0.10.0" }` a few lines above.

Then in `bitcoin.rs` we can `use bdk_bitcoind_rpc::bitcoincore_rpc::jsonrpc::serde_json;` and nothing else needs to change (at least from me testing locally).

How does that sound to you?

reez (changes requested):

reez (commented):
Awesome job adding the massive `PsbtError` and structuring everything really well, just one item requesting a change on that I added a comment for-

bdk-ffi/Cargo.toml
thunderbiscuit: Yeah that's what we had before I think. I wasn't sure if it was cleaner to import the dependency directly or add the bitcoind_rpc lib to our dependencies, but I think because we'll be adding the RPC client anyway this will become a dependency in the future, so it makes sense to just grab it from there.

reez (approved):
ACK e5e7aba2088158f36aeec3175efbe4d3406327e9

bdk-ffi/Cargo.toml
reez: Cool, that makes sense mentioning how we will be adding the rpc client too, good additional context/point

PR #533, opened 2024-05-11
Bump rust bdk to alpha 11

reez:
#### All Submissions:

#### New Features:

#### Bugfixes:


thunderbiscuit (commented):
Looks good! My one thought that might need more discussion is around the Amount type being a dictionary instead of an interface. I think if we eventually want to expand it to more (and the from_sat and from_btc method might be the first ones we go after) in order to use the type for what it's pretty good at, we'll need methods on it, and therefore will not be able to have fields.

Another option is to have an actual struct that would hold both as fields, but then no other cool methods could be called on it. Something like
```rust
struct Amount {
    pub sat: u64,
    pub btc: f64,
}
```

But then you'd still need constructors from_sat and from_btc to create it from the bindigns side (for example when providing it to your txbuilder)? I haven't tried that before but I _think_ that might not be possible.

thunderbiscuit:
Oh and the `Amount` type belongs in the `bitcoin` module rather than the `types` module.

reez:
> Oh and the `Amount` type belongs in the `bitcoin` module rather than the `types` module.

my b, fixed!

reez:
@thunderbiscuit ready for review/merge, rebased on the signet PR, test passing

thunderbiscuit (approved):
ACK e5540880670cc106fc8546964991eadf8463ddbb. Good work! A lot of files changes, but nothing too insane in the end.

Would you mind adding text for the changelog? That way we can just copy/paste it when we make the release.

reez:
> ACK [e554088](https://github.com/bitcoindevkit/bdk-ffi/commit/e5540880670cc106fc8546964991eadf8463ddbb). Good work! A lot of files changes, but nothing too insane in the end.
> 
> Would you mind adding text for the changelog? That way we can just copy/paste it when we make the release.

Sure, added, hopefully its what you were looking for-

thunderbiscuit:
Oh well that's not quite what I meant but it works!

PR #540, opened 2024-05-21
Prepare language bindings libraries for alpha 11 release

thunderbiscuit:
This PR:
1. Fixes the Android tests in 2 ways:
  - The tests execute in parallel, so even if we have a teardown method on the persistence file, two tests can use it at the same time and if (as if they currently stand) one is on testnet and one is on signet, the wallet errors out on initialization
  - The live tests didn't update for the new `Amount` type
2. Bumps the (unused) version of the library in the `Cargo.toml` file to alpha 11.
3. Fixes the Swift tests to also print amounts in sat


reez (approved):
ACK c6174199dde377643f6067df88b30265ffb9984d

PR #544, opened 2024-05-28
Update bdk_wallet to alpha 12

thunderbiscuit:
This PR upgrades the (newly named!) bdk_wallet library to the alpha 12 version. 

Also part of this are the upgrade to bdk_esplora (0.14.0), bdk_electrum (0.14.0), and the introduction of the new bdk_sqlite (0.1.0) persistence crate.

Note:
1. The diff is pretty big because all the imports using `bdk` are now switched to `bdk_wallet`.
2. I changed the names of the persistence files in the tests to `.sqlite`, as is common for those types of files (but is not required). This is also the naming used on the Rust bdk side, so I think it's a good fit anyway.
3. The core upgrade here for the sqlite persistence is surprisingly small: lines 41-42 in the wallet module.
4. I added the SqliteError type which is not used yet (should I remove it and wait to add it when we have a need?). The current places where the codebase can throw a sqlite error is on the wallet creation method, but this one can only throw one of the variants of the [BdkSqliteError](https://docs.rs/bdk_sqlite/0.1.0/bdk_sqlite/enum.Error.html), the one where it's a wrapper for the rusqlite error. So I added that as a variant on the WalletCreationError (that's required), but I also added just the normal bdk Sqlite error just so this PR has it and we can use it in the future.

### Changelog notice
```md
Upgrades
  - Update bdk_esplora client to 0.14.0 [#544]
  - Update bdk_electrum client to 0.14.0 [#544]

Added
  - bdk_sqlite persistence [#544]

Removed
  - flat file persistence [#544]

[#544]: https://github.com/bitcoindevkit/bdk-ffi/pull/544
```

#### All Submissions:

### DIFF
--- a/bdk-ffi/src/error.rs
+++ b/bdk-ffi/src/error.rs
@@ -1,28 +1,28 @@
 use crate::bitcoin::OutPoint;
 
-use bdk::bitcoin::address::Error as BdkAddressError;
-use bdk::bitcoin::address::ParseError;
-use bdk::bitcoin::amount::ParseAmountError as BdkParseAmountError;
-use bdk::bitcoin::bip32::Error as BdkBip32Error;
-use bdk::bitcoin::consensus::encode::Error as BdkEncodeError;
-use bdk::bitcoin::psbt::Error as BdkPsbtError;
-use bdk::bitcoin::psbt::ExtractTxError as BdkExtractTxError;
-use bdk::bitcoin::psbt::PsbtParseError as BdkPsbtParseError;
-use bdk::bitcoin::Network;
-use bdk::chain::local_chain::CannotConnectError as BdkCannotConnectError;
-use bdk::chain::tx_graph::CalculateFeeError as BdkCalculateFeeError;
-use bdk::descriptor::DescriptorError as BdkDescriptorError;
-use bdk::keys::bip39::Error as BdkBip39Error;
-use bdk::miniscript::descriptor::DescriptorKeyParseError as BdkDescriptorKeyParseError;
-use bdk::wallet::error::BuildFeeBumpError;
-use bdk::wallet::error::CreateTxError as BdkCreateTxError;
-use bdk::wallet::signer::SignerError as BdkSignerError;
-use bdk::wallet::tx_builder::AddUtxoError;
-use bdk::wallet::NewOrLoadError;
-use bdk::KeychainKind;
 use bdk_electrum::electrum_client::Error as BdkElectrumError;
 use bdk_esplora::esplora_client::{Error as BdkEsploraError, Error};
-use bdk_file_store::FileError as BdkFileError;
+use bdk_sqlite::rusqlite::Error as BdkSqliteError;
+use bdk_wallet::bitcoin::address::Error as BdkAddressError;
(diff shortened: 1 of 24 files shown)


bdk-ffi/src/wallet.rs
```diff
@@ -296,7 +296,7 @@ impl TxBuilder {
         })
     }
 
-    pub(crate) fn fee_absolute(&self, fee_amount: u64) -> Arc<Self> {
+    pub(crate) fn fee_absolute(&self, fee_amount: Arc<Amount>) -> Arc<Self> {
```
reez, line 299: love the new `Amount` type here and other places, v cool

bdk-ffi/src/error.rs
```diff
@@ -606,7 +618,7 @@ pub enum TxidParseError {
     InvalidTxid { txid: String },
 }
 
-// This error combines the Rust bdk::wallet::NewOrLoadError and bdk_file_store::FileError
+// This error combines the Rust bdk::wallet::NewOrLoadError and bdk_wallet::rusqlite::Error
```
reez, line 621: thanks for even remembering to update these spots, you're on it!

bdk-ffi/src/types.rs
```diff
@@ -26,12 +26,8 @@ pub struct CanonicalTx {
     pub chain_position: ChainPosition,
 }
 
-impl From<BdkCanonicalTx<'_, Arc<bdk::bitcoin::Transaction>, ConfirmationTimeHeightAnchor>>
-    for CanonicalTx
-{
-    fn from(
-        tx: BdkCanonicalTx<'_, Arc<bdk::bitcoin::Transaction>, ConfirmationTimeHeightAnchor>,
-    ) -> Self {
+impl From<BdkCanonicalTx<'_, Arc<BdkTransaction>, ConfirmationTimeHeightAnchor>> for CanonicalTx {
```
reez, line 29: thanks for doing things like this in this PR so we can have things like `Arc<BdkTransaction>` instead of `Arc<bdk::bitcoin::Transaction>`, makes it clean and I need to make sure I'm paying attention to this sort of stuff to for any of my new PR's

reez (approved):

reez:
Really good job!

for :

> I added the SqliteError type which is not used yet (should I remove it and wait to add it when we have a need?). The current places where the codebase can throw a sqlite error is on the wallet creation method, but this one can only throw one of the variants of the [BdkSqliteError](https://docs.rs/bdk_sqlite/0.1.0/bdk_sqlite/enum.Error.html), the one where it's a wrapper for the rusqlite error. So I added that as a variant on the WalletCreationError (that's required), but I also added just the normal bdk Sqlite error just so this PR has it and we can use it in the future.

...I'm good with either decision. In testing on the iOS app i did happen to run into the variant on the `WalletCreationError` so that was nice.

I tested the bindings by building them locally and then using them on the BDK iOS app, and made some slight adjustments to make it transition smoothly on the iOS client just in case anyone is interested in taking a peek at that: https://github.com/bitcoindevkit/BDKSwiftExampleWallet/pull/146

PR #546, opened 2024-05-30
Use external uniffi library for rust-bitcoin types

### LINKED ISSUE #556: Publish the rust-bitcoin types in their own crate
The types that reside in the current `bitcoin.rs` file are the ones that come from the rust-bitcoin crate. I have made good progress to pull those out of this repo and allow them to live in a separate crate and simply be imported. This has serious advantages in the long run:
- It allows other bitcoin projects that build language bindings using uniffi to leverage these types without re-writing them and maintaining them themselves
- If other projects also use those types, it brings more eyes and review on the codebase
- We potentially pool the burden of development and exposing of more rust-bitcoin types among more projects

thunderbiscuit:
This PR allows our library to depend on an external [bitcoin-ffi](https://github.com/thunderbiscuit/rust-bitcoin-ffi) library. This is useful because it means we can collaborate with other projects who also need to use types from rust-bitcoin in their uniffi bindings.

I propose we merge _some_ types coming from bitcoin-ffi for the beta 2 release as a good way to introduce the design and workflow slowly but surely. All libraries build and work well locally and pass CI.

Closes #556.


reez:
This is so so great.

Compiled and ran this PR's Swift bindings locally and used in Example Wallet and everything built and ran in the iOS app ✅

You may already be planning on it, but would love to see this squash down to one commit when merged since that will be nice historically to go back to see what all had to be changed to use the external `bitcoin-ffi` ("Use external uniffi library for rust-bitcoin types").

ACK ee195fb3c4cdd7341e4b6fcf2f58238e2a50dea5

reez (approved):

thunderbiscuit:
Ah yes good point, I think a squash makes sense.

thunderbiscuit:
I have run all live tests locally for all four libraries. In addition, here is the CI workflow for the live tests: https://github.com/bitcoindevkit/bdk-ffi/actions/runs/10636432218.

If all passes in both CIs I'll merge later on today.

PR #551, opened 2024-06-04
Use `Display` trait to auto-populate the toString method and `description` fields

thunderbiscuit:
This is one of my favourite little PRs in a long time.

We cannot write an explicit toString() method on our objects because the name clashes with the `toString()` method on the JVM, and alternatively we cannot add a `description` field on objects because it clashes with the field in Swift. This means that for objects that have a requirement for this toString() method, we actually add a method called `asString()`. It works, but it's uglier than it needs to be and feels a little odd (_why is the toString() not implemented on this type?_, a developer might ask).

But as of the 0.26 release of uniffi, if you define a Display trait on a type, it will automatically use that for the toString() method in Kotlin and the `description` field in Swift! This means of course that we can remove the odd-looking `asString()` (which achieves the same thing), but also that this is can be derived automatically for us by Rust! Here's how it looks in the UDL:

```idl
[Traits=(Display)]
interface Address { }
```

Old printout in Kotlin:
```kotlin
println("Address asString: ${addressInfo.address.asString()}")
// Address asString: tb1qzg4mckdh50nwdm9hkzq06528rsu73hjxxzem3e

println("Address toString: ${addressInfo.address.toString()}")
// Address toString: org.bitcoindevkit.Address@ec2cc4
```

New printout:
```kotlin
// Old stuff
println("Address asString: ${addressInfo.address.asString()}")
// Address asString: tb1qzg4mckdh50nwdm9hkzq06528rsu73hjxxzem3e

// Better
println("Address toString: ${addressInfo.address.toString()}")
// Address toString: tb1qzg4mckdh50nwdm9hkzq06528rsu73hjxxzem3e

// Pure sugar!
println("Address with implicit toString: ${addressInfo.address}")
// Address with implicit toString: tb1qzg4mckdh50nwdm9hkzq06528rsu73hjxxzem3e
```
Note that the `toString()` method is also applied automatically in many places, removing the need to actually write it out explicitly.

In Swift it's just as good. The difference is that you would have gotten an error if you tried to use the `description` field before:
```swift
print("Address description: \(addressInfo.address.description)")
// error: value of type 'Address' has no member 'description'

print("Address implicit description: \(addressInfo.address)")
Address implicit description: BitcoinDevKit.Address
```

You now get:
```swift
// Old stuff
print("Address asString: \(addressInfo.address.asString())")
// Address asString: tb1qzg4mckdh50nwdm9hkzq06528rsu73hjxxzem3e

// Pretty good
print("Address description: \(addressInfo.address.description)")
Address description: tb1qzg4mckdh50nwdm9hkzq06528rsu73hjxxzem3e

// Awesome!
print("Address implicit description: \(addressInfo.address)")
Address implicit description: tb1qzg4mckdh50nwdm9hkzq06528rsu73hjxxzem3e
```

The structs currently affected (and made better by this change) are:
- `Address`
- `DescriptorSecretKey`
- `DescriptorPublicKey`
- `Mnemonic`
- `Descriptor`

But I suspect there might be many more that could use a clean `toString()` implementation which we did not have a clean way to provide.

### Changelog notice

TODO

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/LiveTxBuilderTest.kt
+++ b/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/LiveTxBuilderTest.kt
@@ -36,7 +36,7 @@ class LiveTxBuilderTest {
         println("Balance: ${wallet.getBalance().total.toSat()}")
 
         assert(wallet.getBalance().total.toSat() > 0uL) {
-            "Wallet balance must be greater than 0! Please send funds to ${wallet.revealNextAddress(KeychainKind.EXTERNAL).address.asString()} and try again."
+            "Wallet balance must be greater than 0! Please send funds to ${wallet.revealNextAddress(KeychainKind.EXTERNAL).address} and try again."
         }
 
         val recipient: Address = Address("tb1qrnfslnrve9uncz9pzpvf83k3ukz22ljgees989", Network.SIGNET)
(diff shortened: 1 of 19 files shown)


reez (approved):
ACK 153d9d59f90c5bcc11856af3accafbea53d6d86a

I love this too

Issue #552, opened 2024-06-05
Set name of Swift module in uniffi.toml


thunderbiscuit:
You can add the `module_name` field to the uniffi.toml file to name the module correctly from the get go instead of building the default one (`bdk.swift`) and then renaming it in our shell build file. See all config options for Swift [here](https://mozilla.github.io/uniffi-rs/latest/swift/configuration.html).

Note that fixing this requires adding to the uniffi.toml file:
```toml
[bindings.swift]
cdylib_name = "bdkffi"
module_name = "BitcoinDevKit"
```

But also fixing the `build-local-swift.sh` file as well as the CI workflow on the [bdk-swift](https://github.com/bitcoindevkit/bdk-swift) repo.

- [ ] uniffi.toml
- [ ] build-local-swift.sh
- [ ] publish-spm.yaml on bdk-swift repository

thunderbiscuit:
This might need changes in other places too: we have some files that define metadata for the library that use the bdkFFI string (which then becomes BitcoinDevKitFFI), and the new library headers and umbrella.h and module.modulemap all include it. So not sure if the change in that metadata would have a bigger impact than I initially imaged.

thunderbiscuit:
Done!

PR #557, opened 2024-06-12
chore(swift): update build-local-swift.sh to use xcodebuild tool

notmandatory:
### Description

I updated the build script to use the xcodebuild tool which creates the xc.framework so we don't need to manually copy header and `libbdkffi.a` files or manually create the `Info.plist` file.

I don't know if this will fix any current problems, but it's a start to getting things built the Apple™	way.

### Notes to the reviewers

For this update I referenced:

* https://www.strathweb.com/2023/07/calling-rust-code-from-swift/#switching-to-a-framework
* https://developer.apple.com/documentation/xcode/creating-a-multi-platform-binary-framework-bundle#Generate-the-XCFramework-bundle

#### All Submissions:


thunderbiscuit:
Really good to see the build workflow being updated and cleaned up a bit. A few thoughts just so I don't forget:
- The bdk.xcframework directory can now be completely removed from version control I think
- Since the PR in bitcoindevkit/bdk-swift#49 now uses the script, I think a name change for that script might work well

reez:
> * We just need to make sure that this new Info.plist doesn't impact our fix for XCode 15.3

Just did some testing a few different ways to see if it would break in Xcode 15.3+ and Xcode 16 beta, and everything worked as expected, so that's really great, nice Steve!

thunderbiscuit:
Note that this fixes #552.

thunderbiscuit:
I have also built this locally and all tests pass. CI looks good. @reez do I understand that you have tried using this in XCode and the example iOS app?

It's an important change because it rewrites how we build and publish our Swift package, but I don't know that we could test it any more than what we did (let me know if you'd like to still test it in different ways). My take on this is that I'd rather merge it before we get to the beta releases and the true 1.0, so we catch anything and fix if needed earlier than later.

It's an important change so I'll wait for ACKs of all 3 of us before merging.

thunderbiscuit (approved):
Tested ACK 21ac5267b537e6883f12a7fafb4f44eae7d5a4a4.

reez:
> I have also built this locally and all tests pass. CI looks good. @reez do I understand that you have tried using this in XCode and the example iOS app?
> 
> It's an important change because it rewrites how we build and publish our Swift package, but I don't know that we could test it any more than what we did (let me know if you'd like to still test it in different ways). My take on this is that I'd rather merge it before we get to the beta releases and the true 1.0, so we catch anything and fix if needed earlier than later.
> 
> It's an important change so I'll wait for ACKs of all 3 of us before merging.


>  @reez do I understand that you have tried using this in XCode and the example iOS app?
That's correct. 

> I don't know that we could test it any more than what we did
I'm not sure how we could either.

>  My take on this is that I'd rather merge it before we get to the beta releases and the true 1.0
Agree.

Tested ACK 21ac5267b537e6883f12a7fafb4f44eae7d5a4a4.

reez (approved):

PR #559, opened 2024-06-13
Add display trait to descriptor

reez:
### Description

Adds more to #551 , I was working on BDK iOS and just saw that I was using `Descriptor.asString()` so I wanted to make that nicer by using the display trait on `Mnemonic` instead now after seeing #551 .

This also removes the `asString()` method which is nice.

### Notes to the reviewers

We still have the `as_string_private()`, I couldnt think of an immediately good way to change that, let me know if you know of one or I'll keep thinking on it to see if something strikes me; but I dont see any reason for that to hold this up even though the pairing of to_string and as_string_private might be a little awkward, we are already walking the path of using this new to_string from the Display trait so we should keep using it on types like this because its really nice.

Let me know any thoughts on that though-

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/OfflineDescriptorTest.kt
+++ b/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/OfflineDescriptorTest.kt
@@ -15,7 +15,7 @@ class OfflineDescriptorTest {
 
         assertEquals(
             expected = "tr([be1eec8f/86'/1'/0']tpubDCTtszwSxPx3tATqDrsSyqScPNnUChwQAVAkanuDUCJQESGBbkt68nXXKRDifYSDbeMa2Xg2euKbXaU3YphvGWftDE7ozRKPriT6vAo3xsc/0/*)#m7puekcx",
-            actual = descriptor.asString()
+            actual = descriptor.toString()
         )
     }
 }
--- a/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/OfflineWalletTest.kt
+++ b/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/OfflineWalletTest.kt
@@ -29,7 +29,7 @@ class OfflineWalletTest {
         val descriptorSecretKey: DescriptorSecretKey = DescriptorSecretKey(Network.TESTNET, mnemonic, null)
         val descriptor: Descriptor = Descriptor.newBip86(descriptorSecretKey, KeychainKind.EXTERNAL, Network.TESTNET)
 
-        assertTrue(descriptor.asString().startsWith("tr"), "Bip86 Descriptor does not start with 'tr'")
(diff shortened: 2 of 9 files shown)


thunderbiscuit:
Hey thanks for this! The as_string_private() is just another method with a name that's close but nothing else IMO. 

BUT you got me looking into it just to make sure this is what it was called on the Rust side and in fact in Rust that method is called [`to_string_with_secret()`](https://github.com/bitcoindevkit/bdk-ffi/blob/84f1329e8457248633d49451aa292c29344c8bc5/bdk-ffi/src/descriptor.rs#L263-L267). Weird! Why did we not name it the same??? Git blame tells me it was ME who did that but surely I would not (I must have left my computer unattended at the library for 20 minutes and someone committed that).

So really we don't know who did this terrible mistake but I suggest we fix it here and name it `to_string_with_secret()`. What do you think?

reez:
> Hey thanks for this! The as_string_private() is just another method with a name that's close but nothing else IMO.
> 
> BUT you got me looking into it just to make sure this is what it was called on the Rust side and in fact in Rust that method is called [`to_string_with_secret()`](https://github.com/bitcoindevkit/bdk-ffi/blob/84f1329e8457248633d49451aa292c29344c8bc5/bdk-ffi/src/descriptor.rs#L263-L267). Weird! Why did we not name it the same??? Git blame tells me it was ME who did that but surely I would not (I must have left my computer unattended at the library for 20 minutes and someone committed that).
> 
> So really we don't know who did this terrible mistake but I suggest we fix it here and name it `to_string_with_secret()`. What do you think?

Love it, will do that right now, thanks for looking at the rust side and the good suggestion

reez:
> Hey thanks for this! The as_string_private() is just another method with a name that's close but nothing else IMO.
> 
> BUT you got me looking into it just to make sure this is what it was called on the Rust side and in fact in Rust that method is called [`to_string_with_secret()`](https://github.com/bitcoindevkit/bdk-ffi/blob/84f1329e8457248633d49451aa292c29344c8bc5/bdk-ffi/src/descriptor.rs#L263-L267). Weird! Why did we not name it the same??? Git blame tells me it was ME who did that but surely I would not (I must have left my computer unattended at the library for 20 minutes and someone committed that).
> 
> So really we don't know who did this terrible mistake but I suggest we fix it here and name it `to_string_with_secret()`. What do you think?



> Hey thanks for this! The as_string_private() is just another method with a name that's close but nothing else IMO.
> 
> BUT you got me looking into it just to make sure this is what it was called on the Rust side and in fact in Rust that method is called [`to_string_with_secret()`](https://github.com/bitcoindevkit/bdk-ffi/blob/84f1329e8457248633d49451aa292c29344c8bc5/bdk-ffi/src/descriptor.rs#L263-L267). Weird! Why did we not name it the same??? Git blame tells me it was ME who did that but surely I would not (I must have left my computer unattended at the library for 20 minutes and someone committed that).
> 
> So really we don't know who did this terrible mistake but I suggest we fix it here and name it `to_string_with_secret()`. What do you think?

this should be all set now @thunderbiscuit I made the change `to_string_with_secret()` 👍

PR #560, opened 2024-06-14
Bump minimum supported Android version to 24

### LINKED ISSUE #500: Increase minimum supported Android API
The minimum supported API for Android is 21, just over 10 years old. [This website](https://apilevels.com/) shows the percentage of devices still on old versions of Android. 21 is just... really old and not used much. We set it to 21 2 and a half years ago when starting bdk-ffi but I think we should take a look at what API number we really want to set as the minimum. Opening this issue to keep track.

thunderbiscuit:
This PR bumps our minimum supported Android version to 24. The lowest version I was able to find a user of bdk supporting was 24 (a very conservative choice, see [here](https://apilevels.com/). 21 at this point was a bit ludicrous, and will probably not be available in future NDKs, which we will have to update at some point for CI and just overall health of the Android libs. I think this gets ahead of those issues while being extremely conservative (API level 24 came out in 2016).

I have tested this locally on emulators and in the example application.

Fixes #500.

_Note: I'm sorry this PR actually contains a second commit which fixes the live tests (I noticed the failure and fixed it without realizing it was on this branch). But now it's here I'm just going to leave it here anyway. It's a good fix that was missed as part of the Display trait PRs because it's part of the live tests that don't get run on all normal PR CI runs._

### DIFF
--- a/bdk-android/lib/build.gradle.kts
+++ b/bdk-android/lib/build.gradle.kts
@@ -18,7 +18,7 @@ android {
     compileSdk = 34
 
     defaultConfig {
-        minSdk = 21
+        minSdk = 24
         targetSdk = 34
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         consumerProguardFiles("consumer-rules.pro")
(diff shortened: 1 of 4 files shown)


reez (approved):
ACK a2794f25b09c5fd976922ffe611a0a45b236ea49

Yup this makes sense and 24+ is more than reasonable, good change, and thanks for fixing the live test

PR #586, opened 2024-08-30
feat: extract_keys

reez:
### Description

Adds [extract_keys](https://docs.rs/bdk_wallet/1.0.0-beta.2/bdk_wallet/struct.Wallet.html#method.load)

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-jvm/lib/src/test/kotlin/org/bitcoindevkit/OfflinePersistenceTest.kt
+++ b/bdk-jvm/lib/src/test/kotlin/org/bitcoindevkit/OfflinePersistenceTest.kt
@@ -40,4 +40,35 @@ class OfflinePersistenceTest {
             actual = addressInfo.address.toString(),
         )
     }
+
+    @Test
+    fun testPersistenceWithDescriptor() {
+        val connection = Connection(persistenceFilePath)
+
+        val descriptorPub = Descriptor(
+            "wpkh([9122d9e0/84'/1'/0']tpubDCYVtmaSaDzTxcgvoP5AHZNbZKZzrvoNH9KARep88vESc6MxRqAp4LmePc2eeGX6XUxBcdhAmkthWTDqygPz2wLAyHWisD299Lkdrj5egY6/0/*)#zpaanzgu",
+            Network.SIGNET
+        )
+        val changeDescriptorPub = Descriptor(
+            "wpkh([9122d9e0/84'/1'/0']tpubDCYVtmaSaDzTxcgvoP5AHZNbZKZzrvoNH9KARep88vESc6MxRqAp4LmePc2eeGX6XUxBcdhAmkthWTDqygPz2wLAyHWisD299Lkdrj5egY6/1/*)#n4cuwhcy",
+            Network.SIGNET
+        )
+
+        val wallet: Wallet = Wallet.load(
+            descriptorPub,
+            changeDescriptorPub,
+            connection
+        )
+        val addressInfo: AddressInfo = wallet.revealNextAddress(KeychainKind.EXTERNAL)
+        println("Address: $addressInfo")
+
+        assertEquals(
+            expected = 7u,
+            actual = addressInfo.index,
+        )
+        assertEquals(
+            expected = "tb1qan3lldunh37ma6c0afeywgjyjgnyc8uz975zl2",
+            actual = addressInfo.address.toString(),
+        )
+    }
 }
--- a/bdk-ffi/src/wallet.rs
+++ b/bdk-ffi/src/wallet.rs
@@ -68,6 +68,7 @@ impl Wallet {
         let wallet: PersistedWallet<BdkConnection> = BdkWallet::load()
             .descriptor(KeychainKind::External, Some(descriptor))
             .descriptor(KeychainKind::Internal, Some(change_descriptor))
+            .extract_keys()
             .load_wallet(db)?
             .ok_or(LoadWithPersistError::CouldNotLoad)?;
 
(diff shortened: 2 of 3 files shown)


rustaceanrob:
From the miniscript [source code](https://docs.rs/miniscript/latest/src/miniscript/descriptor/mod.rs.html#745-782)
```rust
  fn key_to_string(pk: &DescriptorPublicKey, key_map: &KeyMap) -> Result<String, ()> {
      Ok(match key_map.get(pk) {
          Some(secret) => secret.to_string(),
          None => pk.to_string(),
      })
  }
```

So the lines above just return the public keys if there are no secrets

thunderbiscuit (changes requested):
Lol the simplest change (1 line, 15 characters) with the biggest effect haha. One thing for me is I'd like a test for this with a public descriptor. You could basically call the load method the [way we do it here](https://github.com/bitcoindevkit/bdk-ffi/blob/d74acfae722746a9add8314bb226fc94fd78cf62/bdk-jvm/lib/src/test/kotlin/org/bitcoindevkit/OfflinePersistenceTest.kt#L22C1-L42C6) using the mock persistence db but feeding it the public descriptor [used here](https://github.com/bitcoindevkit/bdk-ffi/blob/d74acfae722746a9add8314bb226fc94fd78cf62/bdk-jvm/lib/src/test/kotlin/org/bitcoindevkit/OfflineDescriptorTest.kt#L8-L18).

I see this current API with the constructors instead of the builders as fairly unstable at the moment: they prevent the granularity that the builders offer (as is, but we could add arguments), and even at the Rust layer those builders and methods are so so fresh that I don't consider them set in stone yet. Just a random thought thrown out there.

bdk-jvm/lib/src/test/kotlin/org/bitcoindevkit/OfflinePersistenceTest.kt
```diff
@@ -40,4 +40,31 @@ class OfflinePersistenceTest {
             actual = addressInfo.address.toString(),
         )
     }
+
+    @Test
+    fun testPersistenceWithDescriptor() {
+        val connection = Connection(persistenceFilePath)
+
+        val descriptorPub = Descriptor(
```
reez, line 48: possible TODO if this is the sort of test we want:
- rename function
- rename `val`
- make `val` private

bdk-jvm/lib/src/test/kotlin/org/bitcoindevkit/OfflinePersistenceTest.kt
```diff
@@ -40,4 +40,31 @@ class OfflinePersistenceTest {
...
+            "wpkh([9122d9e0/84'/1'/0']tpubDCYVtmaSaDzTxcgvoP5AHZNbZKZzrvoNH9KARep88vESc6MxRqAp4LmePc2eeGX6XUxBcdhAmkthWTDqygPz2wLAyHWisD299Lkdrj5egY6/0/*)#zpaanzgu",
```
reez, line 49: I grabbed this value from printing out the descriptor in [testPersistence](https://github.com/bitcoindevkit/bdk-ffi/blob/d74acfae722746a9add8314bb226fc94fd78cf62/bdk-jvm/lib/src/test/kotlin/org/bitcoindevkit/OfflinePersistenceTest.kt#L22C1-L42C6) since the value in [testDescriptorBip86](https://github.com/bitcoindevkit/bdk-ffi/blob/d74acfae722746a9add8314bb226fc94fd78cf62/bdk-jvm/lib/src/test/kotlin/org/bitcoindevkit/OfflineDescriptorTest.kt#L8-L18) didn't work (`84'` vs `86'`)

bdk-jvm/lib/src/test/kotlin/org/bitcoindevkit/OfflinePersistenceTest.kt
```diff
@@ -40,4 +40,31 @@ class OfflinePersistenceTest {
...
+            Network.SIGNET
+        )
+
+        val wallet: Wallet = Wallet.load(
+            descriptorPub,
+            changeDescriptor,
```
thunderbiscuit, line 55: This is one of the quirks of the new API: you can give a private descriptor and a public descriptor, and the extract keys will work (only on the private one) and then you have an odd wallet that can sign but only parts of the transactions if you have UTXOs from both chains, and none if you have UTXOs from your public chain, but also _all_ if you only use UTXOs of the private descriptor chain. Anyway it's not undefined behaviour per se, but it's odd and I'd not recommend it to anyone unless they have a very clear reason for it. I suggest we don't add a test that displays that weird behaviour either.

Let's use 2 public descriptors instead what do you think? Or let me know if you had a reason in mind for using one of each.

thunderbiscuit (changes requested):
My comment applies for both JVM and Swift tests.

bdk-jvm/lib/src/test/kotlin/org/bitcoindevkit/OfflinePersistenceTest.kt
thunderbiscuit: By the way this is another reason why multipath descriptors are awesome.

reez: > Let's use 2 public descriptors instead what do you think? Or let me know if you had a reason in mind for using one of each.

Sounds good, I was just trying it to see what would happen in that case, pushing up changes to the tests to use 2 public descriptors now-

thunderbiscuit (approved):
ACK 647726434e7ad1afb129e017712b6998586c8023.

PR #590, opened 2024-09-16
Small refactors

### LINKED ISSUE #583: Move TxBuilder and associated types in their correct modules
The TxBuilder should go in its own module and the other 2 or 3 smaller types should go in the types module.

thunderbiscuit:
### Description

This PR:
1. Removes our dependency on bitcoin-internals (it was used for one method in the errors in order to convert an array of bytes to a hex string). The library is not recommended for usage outside of the rust-bitcoin org, and indeed it was the wrong dependency for the job. We can simply use the DisplayHex trait implemented in `bdk_wallet::bitcoin::hex::DisplayHex`.
2. Creates a tx_builder module to mirror the BDK side of things, and moves the TxBuilder types there.
3. Moves two smaller types that were forgotten in the wallet module to their correct `types` module.

### Notes to the reviewers
These are simply refactors and do not impact the public API.

### Changelog notice
No changelog notice.

Fixes #583.

### DIFF
--- a/bdk-ffi/Cargo.lock
+++ b/bdk-ffi/Cargo.lock
@@ -140,7 +140,7 @@ version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2c8d66485a3a2ea485c1913c4572ce0256067a5377ac8c75c4960e1cda98605f"
 dependencies = [
- "bitcoin-internals 0.3.0",
+ "bitcoin-internals",
  "bitcoin_hashes 0.14.0",
 ]
 
@@ -182,7 +182,6 @@ dependencies = [
  "bdk_esplora",
  "bdk_wallet",
  "bitcoin-ffi",
- "bitcoin-internals 0.2.0",
  "thiserror",
  "uniffi",
 ]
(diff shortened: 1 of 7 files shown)


reez:
> ### Description
> This PR:
> 
> 1. Removes our dependency on bitcoin-internals (it was used for one method in the errors in order to convert an array of bytes to a hex string). The library is not recommended for usage outside of the rust-bitcoin org, and indeed it was the wrong dependency for the job. We can simply use the DisplayHex trait implemented in `bdk_wallet::bitcoin::hex::DisplayHex`.
> 2. Creates a tx_builder module to mirror the BDK side of things, and moves the TxBuilder types there.
> 3. Moves two smaller types that were forgotten in the wallet module to their correct `types` module.
> 
> ### Notes to the reviewers
> These are simply refactors and do not impact the public API.
> 
> ### Changelog notice
> No changelog notice.
> 
> Fixes #583.

Absolutely love all of these changes, makes the codebase even better and crisp.

reez (approved):
ACK 5ad8de2b5768b6467f7e6095d82600837b9a4d49

PR #584, opened 2024-08-16
Use shell scripts to build bdk-jvm and bdk-android

thunderbiscuit:
This PR removes the Gradle plugins form the JVM and Android libraries and replaces them with simpler to read, simpler to maintain shell scripts.

## TODO
- [x] Fix test CI workflow (JVM)
- [x] Fix publishing CI workflow (JVM)
- [x] Fix justfile (JVM)
- [x] Fix live tests workflow
- [x] Fix test CI workflow (Android)
- [x] Fix publishing CI workflow (Android)
- [x] Fix justfile (Android)

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #598, opened 2024-09-26
feat: esplora client get_tx

reez:
### Description

Adds [get_tx](https://docs.rs/bdk_wallet/latest/bdk_wallet/struct.Wallet.html#method.get_tx) to Esplora Client

### Notes to the reviewers

Jurvis just added this method to pre-1.0 (thank you!), so I'm adding to 1.0; while similar to pre-1.0 it is not exactly the same but I tried to match the pre-1.0 functions naming conventions.

I was also going to add this method for Electrum but since that is slightly more different I decided to spin it out into a separate upcoming PR.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/esplora.rs
+++ b/bdk-ffi/src/esplora.rs
@@ -6,6 +6,7 @@ use crate::types::{FullScanRequest, SyncRequest};
 use bdk_esplora::esplora_client::{BlockingClient, Builder};
 use bdk_esplora::EsploraExt;
 use bdk_wallet::bitcoin::Transaction as BdkTransaction;
+use bdk_wallet::bitcoin::Txid;
 use bdk_wallet::chain::spk_client::FullScanRequest as BdkFullScanRequest;
 use bdk_wallet::chain::spk_client::FullScanResult as BdkFullScanResult;
 use bdk_wallet::chain::spk_client::SyncRequest as BdkSyncRequest;
@@ -14,6 +15,7 @@ use bdk_wallet::KeychainKind;
 use bdk_wallet::Update as BdkUpdate;
 
 use std::collections::BTreeMap;
+use std::str::FromStr;
 use std::sync::Arc;
 
 pub struct EsploraClient(BlockingClient);
@@ -80,5 +82,11 @@ impl EsploraClient {
         self.0
             .broadcast(&bdk_transaction)
             .map_err(EsploraError::from)
+    }
+
+    pub fn get_tx(&self, txid: String) -> Result<Option<Arc<Transaction>>, EsploraError> {
+        let txid = Txid::from_str(&txid)?;
+        let tx_opt = self.0.get_tx(&txid)?;
+        Ok(tx_opt.map(|inner| Arc::new(Transaction::from(inner))))
     }
 }
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -580,6 +580,9 @@ interface EsploraClient {
 
   [Throws=EsploraError]
   void broadcast([ByRef] Transaction transaction);
+
+  [Throws=EsploraError]
+  Transaction? get_tx(string txid);
 };
 
 // ------------------------------------------------------------------------
--- a/bdk-ffi/src/error.rs
+++ b/bdk-ffi/src/error.rs
@@ -7,6 +7,7 @@ use bdk_wallet::bitcoin::address::FromScriptError as BdkFromScriptError;
 use bdk_wallet::bitcoin::address::ParseError as BdkParseError;
 use bdk_wallet::bitcoin::bip32::Error as BdkBip32Error;
 use bdk_wallet::bitcoin::consensus::encode::Error as BdkEncodeError;
+use bdk_wallet::bitcoin::hashes::hex::HexToArrayError as BdkHexToArrayError;
 use bdk_wallet::bitcoin::hex::DisplayHex;
 use bdk_wallet::bitcoin::psbt::Error as BdkPsbtError;
 use bdk_wallet::bitcoin::psbt::ExtractTxError as BdkExtractTxError;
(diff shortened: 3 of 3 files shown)


bdk-ffi/src/esplora.rs
```diff
@@ -81,4 +84,14 @@ impl EsploraClient {
             .broadcast(&bdk_transaction)
             .map_err(EsploraError::from)
     }
+
+    pub fn get_tx(&self, txid: String) -> Result<Option<Arc<Transaction>>, EsploraError> {
+        let txid = Txid::from_str(txid.as_str()).map_err(|e| EsploraError::Parsing {
+            error_message: format!("Invalid txid: {}", e),
+        })?;
```
thunderbiscuit, line 91: Do I understand correctly that this is sort of remapping the error you might get from the `Txid::from_str()` into an `EsploraError::Parsing`?

I do think it probably makes sense here because otherwise you'd need to combine both the Txid parsing error and the Esplora error into one, which gets messy when there are multiple of those combinations across the codebase.

thunderbiscuit (commented):
Just one small question to make sure I understand, but this looks good to go.

bdk-ffi/src/esplora.rs
thunderbiscuit: Other observation: with the introduction of the bitcoin-ffi type `Txid` and the `impl_string_custom_typedef` macro, can that translation into the Rust Txid type happen automatically? It's a String on the target language side but becomes a true Rust `Txid` when you feed it to a Rust function? It's my first time using the TypeDef in practice so not 100% sure... but maybe?

reez: > Do I understand correctly that this is sort of remapping the error you might get from the `Txid::from_str()` into an `EsploraError::Parsing`?
> 
> I do think it probably makes sense here because otherwise you'd need to combine both the Txid parsing error and the Esplora error into one, which gets messy when there are multiple of those combinations across the codebase.

yup that's correct.

Needing to remap the error that we get back from parsing which is different than the `EsploraError` type that we return from the function.

Looking at the Txid docs for `from_str` it seemed like the `EsploraError::Parsing` would make the most sense.

But yeah I played around with a few things, and what you mention about combining seemed a bit messy to me too.

The one thing I did think about doing was creating a conversion for it in `error.rs` like:

```
impl From<BdkHexToArrayError> for EsploraError {
    fn from(error: BdkHexToArrayError) -> Self {
        EsploraError::Parsing {
            error_message: format!("Invalid txid: {}", error),
        }
    }
}
```

but `bitcoin::hashes::hex::HexToArrayError` would need to be added as a dependency so I didn't think it was worth adding a new dependency based on one error to be able to clean up 

```
     let txid = Txid::from_str(txid.as_str()).map_err(|e| EsploraError::Parsing {
            error_message: format!("Invalid txid: {}", e),
        })?;
```

to something like `let txid = Txid::from_str(&txid)?;`.

but as I'm writing this it's dawning on me that bdk_wallet probably exports that dependency so I dont need to explicitly add it which solves my initial hesitation to add the conversion of the error in `error.rs`.... so look out for a push with that change right now to make things look even nicer.

reez: > Other observation: with the introduction of the bitcoin-ffi type `Txid` and the `impl_string_custom_typedef` macro, can that translation into the Rust Txid type happen automatically? It's a String on the target language side but becomes a true Rust `Txid` when you feed it to a Rust function? It's my first time using the TypeDef in practice so not 100% sure... but maybe?

From my understanding that `impl_string_custom_typedef` macro only helps with conversions at the FFI boundary not within Rust functions.

thunderbiscuit:
This might solve both your issues in one go. I tried it locally and everything compiles well. I think the neat part is that because you pass in a true Txid type and not a String, your parsing error will happen prior to you passing the Txid to the method (on the construction of the Txid type in theory, but now that I think of it we don't have a constructor for it... so where does the error get thrown exactly if you provide a malformed txid string?).

Anyway will test a bit more, but your `Result` is cleaner because it doesn't need to wrap two errors.

```rust
pub fn get_tx(&self, txid: Txid) -> Result<Option<Arc<Transaction>>, EsploraError> {
    let tx_opt = self.0.get_tx(&txid)?;
    Ok(tx_opt.map(|inner| Arc::new(Transaction::from(inner))))
}
```

reez:
> This might solve both your issues in one go. I tried it locally and everything compiles well. I think the neat part is that because you pass in a true Txid type and not a String, your parsing error will happen prior to you passing the Txid to the method (on the construction of the Txid type in theory, but now that I think of it we don't have a constructor for it... so where does the error get thrown exactly if you provide a malformed txid string?).
> 
> Anyway will test a bit more, but your `Result` is cleaner because it doesn't need to wrap two errors.
> 
> ```rust
> pub fn get_tx(&self, txid: Txid) -> Result<Option<Arc<Transaction>>, EsploraError> {
>     let tx_opt = self.0.get_tx(&txid)?;
>     Ok(tx_opt.map(|inner| Arc::new(Transaction::from(inner))))
> }
> ```

Definitely tried changing the function signature to that exact thing while messing around with the best shape of this, so happy to expand this convo to potentially changing the function signature, but I opted to not at the time because it felt like changing to using a true Txid type instead of a String was something that might have other implications (you mention where error might get thrown if malformed txid string) and/or other spots in the codebase we might want to make the same change (`BumpFeeTxBuilder`?). So I kind of stopped myself moving forward in that direction, but I'm cool with either direction you want to take this, and am glad you are thinking about some similar things (and also things I may not have thought of)!

thunderbiscuit:
Ok so I opened an issue relating our findings here: bitcoindevkit/bitcoin-ffi#20

bdk-ffi/src/esplora.rs
```diff
@@ -13,7 +13,10 @@ use bdk_wallet::chain::spk_client::SyncResult as BdkSyncResult;
 use bdk_wallet::KeychainKind;
 use bdk_wallet::Update as BdkUpdate;
 
+use bitcoin_ffi::Txid;
```
thunderbiscuit, line 16: I feel like you don't need this bitcoin_ffi::Txid type imported since you're not using the bitcoin_ffi type really. A cleaner import would be `use bdk_wallet::bitcoin::Txid;`

thunderbiscuit (changes requested):
Small import change but otherwise looks good.

bdk-ffi/src/esplora.rs
reez: great call, updated-

thunderbiscuit (approved):
ACK 87fd113c9f2f8c9f694434ef1898289b29fc657f. Feel free to merge, I'll rebase #584.

Issue #596, opened 2024-09-26
API surface of the `Wallet` type


thunderbiscuit:
After taking some time to look at what parts of the `Wallet` API were available/not available/useful for the bindings, many methods stood out. This issue will track them. Note that not all of those are top priority, but most of them we figured would be useful and not too costly to include either in the first version of the 1.0 or in a future feature release (1.1, 1.2, etc.)

- [x] Wallet::peek_address
- [x] Wallet::cancel_tx
- [x] Wallet::next_derivation_index
- [x] Wallet::derivation_of_spk
- [x] Wallet::descriptor_checksum
- [x] Wallet::finalize_psbt
- [x] Wallet::get_utxo
- [x] Wallet::list_unused_addresses
- [x] Wallet::mark_used
- [x] Wallet::next_unused_address
- [x] Wallet::policies
- [x] Wallet::public_descriptor
- [x] Wallet::reveal_addresses_to
- [x] ~Wallet::unbroadcast_transactions~

rustaceanrob:
What I would consider high priority:`finalize_psbt`, `cancel_tx`, `public_descriptor`, `descriptor_checksum`. I can hack at those if others agree they should be included in `1.0`

thunderbiscuit:
The `Wallet::unbroadcast_transactions` method requires exposing the [`TxNode`](https://docs.rs/bdk_chain/0.20.0/bdk_chain/tx_graph/struct.TxNode.html) type, which is a bit more than I want to do at the moment for a method no one has asked directly. 

I'd be great to define the exact use case for this (I think its about creating unbroadcast transactions so as to lock the UTXOs for transaction builder, and eventually to take stock maybe of those tx we've created and maybe want to cancel using `cancel_tx()`? Need to dig a bit more.

thunderbiscuit:
With the merge of #626 and #625 this list will be complete.

Issue #597, opened 2024-09-26
API surface of the `TxBuilder` type


thunderbiscuit:
After taking some time to look at what parts of the TxBuilder API were available/not available/useful for the bindings, many methods stood out. This issue will track them. Note that not all of those are top priority, but most of them we figured would be useful and not too costly to include either in the first version of the 1.0 or in a future feature release (1.1, 1.2, etc.)

- [ ] TxBuilder::add_data
- [ ] TxBuilder::coin_selection
- [ ] TxBuilder::current_height
- [ ] TxBuilder::nlocktime
- [ ] TxBuilder::ordering
- [ ] TxBuilder::sighash
- [ ] TxBuilder::allow_dust
- [ ] TxBuilder::policy_path
- [ ] TxBuilder::version

thunderbiscuit:
Some notes related to the APIs I looked into:

1. [TxBuilder::ordering](https://docs.rs/bdk_wallet/1.0.0-beta.5/bdk_wallet/struct.TxBuilder.html#method.ordering) requires adding the [TxOrdering](https://docs.rs/bdk_wallet/1.0.0-beta.5/bdk_wallet/tx_builder/enum.TxOrdering.html) type, which has the `Custom` variant:
```rust
Custom {
    input_sort: Arc<dyn Fn(&TxIn, &TxIn) -> Ordering>,
    output_sort: Arc<dyn Fn(&TxOut, &TxOut) -> Ordering>,
}
```

PR #601, opened 2024-10-07
Add wallet apis

thunderbiscuit:
This PR brings in more `Wallet` APIs listed in #596. Draft for now, I'll see about adding as much as I can today/tomorrow.

### Changelog notice

```md
Added:
  - Wallet::cancel_tx method [#601]
  - Wallet::get_utxo [#601]
  - Wallet::derivation_of_spk [#601]

[#601]: https://github.com/bitcoindevkit/bdk-ffi/pull/601
```

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -376,11 +376,13 @@ interface Wallet {
   constructor(Descriptor descriptor, Descriptor change_descriptor, Connection connection);
 
   void cancel_tx([ByRef] Transaction tx);
-  
+
   u32? derivation_index(KeychainKind keychain);
 
   LocalOutput? get_utxo(OutPoint op);
-  
+
+  KeychainAndIndex? derivation_of_spk(Script spk);
+
   AddressInfo reveal_next_address(KeychainKind keychain);
 
   AddressInfo peek_address(KeychainKind keychain, u32 index);
@@ -631,6 +633,11 @@ dictionary ScriptAmount {
 dictionary SentAndReceivedValues {
     Amount sent;
     Amount received;
+};
+
+dictionary KeychainAndIndex {
+    KeychainKind keychain;
+    u32 index;
 };
 
 // ------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


thunderbiscuit:
The two remaining methods we talked about exposing (Wallet::policies and Wallet::public_descriptor) require returning a simplified version of the Rust return type or adding new types we don't currently have.

I think both of those might be fine just returning strings (a policy string/json and a descriptor) but at the same time if we change our mind later we'll create a breaking change by switching the return type of the methods. 

I'm thinking we can merge this PR as is and punt on those 2 methods which are not super requested and simply add them in a 1.1 release (or one of the later betas if we go to 6 and 7), potentially with their full types or simply strings after more discussion with the team. This way these methods can make it into the beta 5 release and we can move on to the TxBuilder methods we want to add (see #597).

What do you think @reez @notmandatory @rustaceanrob?

rustaceanrob:
My main fixation on `public_descriptor` is to allow `1.0` users to export their wallet somehow as of the first stable release. I agree that it seems to warrant a new type. I could be persuaded either way, but I felt that some wallet export support should be available from day one.

rustaceanrob:
Pull request LGTM

thunderbiscuit:
@rustaceanrob yeah true that. The reason I think we _could_ get away with it is that right now the `Wallet` constructor requires a `Descriptor` (one of our sort of combined/merge type for ffi), which has a `to_string()`/`Display` method on it, which prints the same public descriptor that will come out of the `Wallet::public_descriptor` method.

Mind you we could just decide to return the public descriptor string too. We have a bit of time to think about it before the final 1.0 on the Rust side I think. Eventually I do think that a more complete exposition of the Descriptor types and their inner parts might be nice to have for us, because right now at the ffi layer you get much less than you do in Rust for those types, including things like rust-bitcoin keys stuff (PublicKey, ExtendedKey, etc.). One of the hiccups for them is that they're absolutely chuck-full of generics, so they need a bit of massaging. Maybe a job for bitcoin-ffi?

reez:
ACK c8a8983ee401894c3247869df5f57818f8702ead

Can merge, and I think those final two things still to expose I'm interested in discussing more per https://github.com/bitcoindevkit/bdk-ffi/issues/605 and https://github.com/bitcoindevkit/BDKSwiftExampleWallet/issues/230

reez (approved):

PR #611, opened 2024-10-22
Add TxBuilder APIs

thunderbiscuit:
This PR brings in some of the missing APIs on the TxBuilder outlined in #597.

- [x] add_data

### Changelog notice

```md
Added:
  - TxBuilder::add_data [#611]

[#611]: 
```

#### All Submissions:

#### New Features:


reez (approved):
Watching Steve's tabconf talk last week he used this functionality and I actually jotted down a note about it wanting to try it out in bdk-ffi/iOS

ACK e06f2a9d8b699ea055b14d14e3192ab6eac249fe

Side note: the two checkboxes `I've added tests for the new feature` + `I've added docs for the new feature` are checked, are you adding a test as a follow up and/or adding docs as a follow up to whatever PR we do adding docstrings, or were those just checked accidentally?

rustaceanrob:
With the discussion on bitcoindevkit/bdk#1676 I think we should add a `set_version` or similar while we are at it. Transaction version one is a fingerprint, and users might be clever enough to use transaction version three

thunderbiscuit:
This is finally ready for review. I recommend reviewing 1 commit at a time, as they are individual independent pieces. There are 3 methods left on the TxBuilder identified in #597, but all 3 require a bit more work/thought behind them, and would not be creating breaking changes when we add them.

I think merging the 5 methods proposed here makes sense for the 1.0 release.

bdk-ffi/src/tx_builder.rs
```diff
@@ -237,6 +248,10 @@ impl TxBuilder {
         if let Some(sequence) = self.sequence {
             tx_builder.set_exact_sequence(Sequence(sequence));
         }
+        if !&self.data.is_empty() {
+            let push_bytes = PushBytesBuf::try_from(self.data.clone())?;
+            tx_builder.add_data(&push_bytes);
+        }
```
reez, line 291: I'm reviewing these commit by commit as recommended, and for this `add_data` I was wondering if we needed to manually validate the size of the data here based on Bitcoin's standard transaction relay policy. I'm wondering this because that thought popped into my head so I started writing some tests around it like:
- `test_add_data`
- `test_add_data_within_limit`
- `test_add_data_exceeding_limit`
- `test_add_empty_data`

and I couldn't get `fn test_add_data_exceeding_limit` test to work without adding something like 
```
            if self.data.len() > 80 {
                return Err(CreateTxError::PushBytesError);
            }
```

in the `finish` method.

Thoughts? I could totally be missing something though.

Also let me know if adding the specific implementation of my tests is helpful too.

bdk-ffi/src/tx_builder.rs
```diff
@@ -204,6 +206,13 @@ impl TxBuilder {
         })
     }
 
+    pub(crate) fn current_height(&self, height: u32) -> Arc<Self> {
+        Arc::new(TxBuilder {
+            current_height: Some(height),
+            ..self.clone()
+        })
+    }
+
```
reez, line 222: reviewing commit by commit, this looks good, I wrote a test for it and it passed ✅

bdk-ffi/src/tx_builder.rs
```diff
@@ -213,6 +216,13 @@ impl TxBuilder {
         })
     }
 
+    pub(crate) fn nlocktime(&self, locktime: LockTime) -> Arc<Self> {
+        Arc::new(TxBuilder {
+            locktime: Some(locktime),
+            ..self.clone()
+        })
+    }
+
```
reez, line 229: looks great, wrote a test and passed ✅

bdk-ffi/src/tx_builder.rs
```diff
@@ -223,6 +225,13 @@ impl TxBuilder {
         })
     }
 
+    pub(crate) fn allow_dust(&self, allow_dust: bool) -> Arc<Self> {
+        Arc::new(TxBuilder {
+            allow_dust,
+            ..self.clone()
+        })
+    }
+
```
reez, line 236: looks great as well, wrote a test and passed ✅

bdk-ffi/src/tx_builder.rs
```diff
@@ -232,6 +234,13 @@ impl TxBuilder {
         })
     }
 
+    pub(crate) fn version(&self, version: i32) -> Arc<Self> {
+        Arc::new(TxBuilder {
+            version: Some(version),
+            ..self.clone()
+        })
+    }
+
```
reez, line 243: Was able to write two tests (one for version 1 and one for version 2) and both passed, looks good!

reez (changes requested):
did a review, left comments for each function I tested. the only one I have outstanding questions on is `add_data`, so just requesting changes atm but I could be wrong on my side

good job with these tho!

bdk-ffi/src/tx_builder.rs
rustaceanrob: Personally, I don't think `add_data` should land for `1.0.0`. `PushBytesBuf` doesn't have a clear API and `is_standard_op_return` isn't even a method yet in `rust-bitcoin` (I am [patching](https://github.com/rust-bitcoin/rust-bitcoin/pull/3643) the method right now). Ideally everything about OP_RETURN is left to the Rust-side and FFI has no logic for handling this

reez: > Ideally everything about OP_RETURN is left to the Rust-side and FFI has no logic for handling this

Def agree with this 100%. I just happened to be running into an issue with it when I was coming up with test to test what happened when I gave it data exceeding that limit. But maybe I wrote my test incorrectly, will double back and look at it again

rustaceanrob: There is no check on the Rust side and currently it will allow for invalid OP_RETURN data

reez: Looking back at my test the other thing that played into the way I wrote it was how I approached non-standard transactions.

reez: > There is no check on the Rust side and currently it will allow for invalid OP_RETURN data

Thanks for this! I didn't see it before I wrote:

> Looking back at my test the other thing that played into the way I wrote it was how I approached non-standard transactions.

I'm actually really interested in where we land on this conversation even if for the time being overall we end up at this place you mentioned rob:

> I don't think add_data should land for 1.0.0

thunderbiscuit:
@reez docstrings are absolutely required at this point thanks for reminding me and feel free put your foot down a little harder when I forget and use the full might of the full stop character just kidding that would hurt my feelings.

But jokes aside I do expect PRs going forward to have docstrings so please hold me to it. We've gotten out of the habit but it's not an excuse.

reez:
> * I've added docs for the new feature

Haha, no problem, sounds good, my plan is to just look for this checkbox to be ✅ on PR's with new methods from now on

Issue #605, opened 2024-10-10
Spending Policy Required: External error on tx builder finish


BitcoinZavior:
**Describe the bug**

Encountering a `Spending policy required: External` error when attempting to build a transaction using a multi-sig descriptor. Currently, bdk-ffi does not provide a `policy_path()` in `txBuilder`, which is necessary for specifying the correct policy path during transaction creation.

Proposed Solution:
- Expose `policies()` from the wallet to retrieve available spending policies.
- Add support for `policy_path()` in `txBuilder` to allow specifying the correct policy path when building the transaction.

These changes will ensure that the correct branch of a script is selected when creating transactions, preventing ambiguity issues.

Use Case:
An example where the `policy_path` is needed is with the following descriptor:

`wsh(thresh(2,pk(A),sj:and_v(v:pk(B),n:older(6)),snj:and_v(v:pk(C),after(630000))))`,
which is derived from the miniscript policy:

`thresh(2,pk(A),and(pk(B),older(6))`,and`(pk(C),after(630000)))`.
In this case, the script can be satisfied in multiple ways. Without specifying the `policy_path`, it becomes ambiguous which spending path should be used for the transaction, leading to the error.

BitcoinZavior:
@afilini issue raised for https://github.com/LtbLightning/bdk-flutter/issues/144

reez:
Closing due to #629 being merged now-

PR #612, opened 2024-10-25
Add Spending Policy API

BitcoinZavior:
### Description

Fixes the follwoing bug:

Encountering a Spending policy required: External error when attempting to build a transaction using a multi-sig descriptor. Currently, bdk-ffi does not provide a policy_path() in txBuilder, which is necessary for specifying the correct policy path during transaction creation.

The PR fix has the following changes:

- Expose policies() from the wallet to retrieve available spending policies.
- Add support for policy_path() in txBuilder to allow specifying the correct policy path when building the transaction.

#### All Submissions:

#### New Features:

#### Bugfixes:

Fixes https://github.com/bitcoindevkit/bdk-ffi/issues/605

### DIFF
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -444,7 +444,9 @@ interface Wallet {
 
 interface Update {};
 
-interface Policy {};
+interface Policy {
+  string id();
+};
 
 interface TxBuilder {
   constructor();
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -461,6 +461,8 @@ interface TxBuilder {
 
   TxBuilder add_utxo(OutPoint outpoint);
 
+  TxBuilder policy_path(record<string, sequence<u64>> policy_path, KeychainKind keychain);
+
   TxBuilder change_policy(ChangeSpendPolicy change_policy);
 
   TxBuilder do_not_spend_change();
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -401,6 +401,9 @@ interface Wallet {
 
   string descriptor_checksum(KeychainKind keychain);
 
+  [Throws=DescriptorError]
+  Policy? policies(KeychainKind keychain);
+
   Balance balance();
 
   [Throws=CannotConnectError]
@@ -440,6 +443,8 @@ interface Wallet {
 };
 
 interface Update {};
+
+interface Policy {};
 
 interface TxBuilder {
   constructor();
--- a/bdk-ffi/src/types.rs
+++ b/bdk-ffi/src/types.rs
@@ -251,3 +251,8 @@ impl From<Policy> for BdkPolicy {
         value.0
     }
 }
+impl Policy {
+    pub fn id(&self) -> String {
+        self.0.id.clone()
+    }
+}
--- a/bdk-ffi/src/types.rs
+++ b/bdk-ffi/src/types.rs
@@ -15,12 +15,12 @@ use bdk_wallet::chain::tx_graph::CanonicalTx as BdkCanonicalTx;
 use bdk_wallet::chain::{
     ChainPosition as BdkChainPosition, ConfirmationBlockTime as BdkConfirmationBlockTime,
 };
+use bdk_wallet::descriptor::Policy as BdkPolicy;
 use bdk_wallet::AddressInfo as BdkAddressInfo;
 use bdk_wallet::Balance as BdkBalance;
 use bdk_wallet::KeychainKind;
 use bdk_wallet::LocalOutput as BdkLocalOutput;
 use bdk_wallet::Update as BdkUpdate;
-
 use std::sync::{Arc, Mutex};
 
 #[derive(Debug)]
@@ -236,4 +236,18 @@ pub struct SentAndReceivedValues {
 pub struct KeychainAndIndex {
     pub keychain: KeychainKind,
     pub index: u32,
+}
+
+/// Descriptor spending policy
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct Policy(BdkPolicy);
+impl From<BdkPolicy> for Policy {
+    fn from(value: BdkPolicy) -> Self {
+        Policy(value)
+    }
+}
+impl From<Policy> for BdkPolicy {
+    fn from(value: Policy) -> Self {
+        value.0
+    }
 }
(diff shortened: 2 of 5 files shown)


bdk-ffi/src/bdk.udl
```diff
@@ -441,6 +444,10 @@ interface Wallet {
 
 interface Update {};
 
+interface Policy {
+  string id();
```
reez, line 448: Is this the only field you need/want for your use case? Just curious really, since we can expose other fields later if needed obv.

Regardless, looks good https://docs.rs/bdk_wallet/latest/bdk_wallet/descriptor/policy/struct.Policy.html

bdk-ffi/src/bdk.udl
```diff
@@ -456,6 +463,8 @@ interface TxBuilder {
 
   TxBuilder add_utxo(OutPoint outpoint);
 
+  TxBuilder policy_path(record<string, sequence<u64>> policy_path, KeychainKind keychain);
```
reez, line 466: I think this is our first use of `record` https://mozilla.github.io/uniffi-rs/0.27/udl/builtin_types.html in the UDL 

Looks good https://docs.rs/bdk_wallet/latest/bdk_wallet/tx_builder/struct.TxBuilder.html#method.policy_path

bdk-ffi/src/types.rs
```diff
@@ -15,12 +15,12 @@ use bdk_wallet::chain::tx_graph::CanonicalTx as BdkCanonicalTx;
 use bdk_wallet::chain::{
     ChainPosition as BdkChainPosition, ConfirmationBlockTime as BdkConfirmationBlockTime,
 };
+use bdk_wallet::descriptor::Policy as BdkPolicy;
 use bdk_wallet::AddressInfo as BdkAddressInfo;
 use bdk_wallet::Balance as BdkBalance;
 use bdk_wallet::KeychainKind;
 use bdk_wallet::LocalOutput as BdkLocalOutput;
 use bdk_wallet::Update as BdkUpdate;
-
```
reez, line 23: nit: I think we can keep this blank line since we seem to break these `use` in groups

bdk-ffi/src/wallet.rs
```diff
@@ -154,8 +161,7 @@ impl Wallet {
 
     pub(crate) fn sign(
         &self,
-        psbt: Arc<Psbt>,
-        // sign_options: Option<SignOptions>,
```
reez, line 158: nit: might be nice to keep this on a separate line, but at the same time not any sort of blocker to approving this PR (and we might just remove the whole commented line before 1.0 beta anyway on bdk-ffi)

reez (approved):
Nice! Thanks!!

A couple comments that are just notes and don't require any action, and then one question and two nits, but nothing that blocks me approving this or us merging as is:

ACK 4925a8d7484cac05ef0b0d44cf551360e2771bac

nits:
- comment on `wallet.rs`
- comment on `types.rs`

question:
- I see you checked `I've added tests for the new feature` in the PR template does that mean you are planning on adding a test (I didn't see one in current code)? Not even saying it's needed (and I'm assuming you tested on an external application since you opened #605 ). _Note: I didn't test on an external app during this PR review._

thunderbiscuit:
I'm trying to understand this `Policy` type better. Reading the example I imagine this policy to be a tree of policy nodes, each with an ID. But the `Wallet::policies()` method returns a single Policy that has a single id?

How do I know which path (u32) to choose if I'm not given the whole tree and options to choose from? I see that the `Policy` type potentially contains a field of type [`SatisfiableItem`](https://docs.rs/bdk_wallet/latest/bdk_wallet/descriptor/policy/enum.SatisfiableItem.html), which itself could be a 
```rust
Thresh {
    items: Vec<Policy>,
    threshold: usize,
},
```

Should we expose this field too so that users can query the satifiable item(s)?

I'm just not clear on this API and I want to make sure we get you what you need.

bdk-ffi/src/bdk.udl
BitcoinZavior: Other fields have been exposed now.

bdk-ffi/src/bdk.udl
BitcoinZavior: Yes, thanks!

bdk-ffi/src/types.rs
BitcoinZavior: Sure

BitcoinZavior:
Thanks @thunderbiscuit I have exposed other fields as well now.

bdk-ffi/src/wallet.rs
BitcoinZavior: Done!

BitcoinZavior:
> 
> Should we expose this field too so that users can query the satifiable item(s)?
> 
I have now exposed this field

BitcoinZavior:
> I'm trying to understand this `Policy` type better. Reading the [`Wallet::policies` example](https://docs.rs/bdk_wallet/latest/bdk_wallet/struct.Wallet.html#method.policies) I imagine this policy to be a tree/list of policy nodes, each with an ID. But the `Wallet::policies()` method returns a single Policy that has a single id?
> 
> How do I know which path (u32) to choose if I'm not given the whole tree and options to choose from? I see that the `Policy` type potentially contains a field of type [`SatisfiableItem`](https://docs.rs/bdk_wallet/latest/bdk_wallet/descriptor/policy/enum.SatisfiableItem.html), which itself could be of the variant
> 
> ```rust
> Thresh {
>     items: Vec<Policy>,
>     threshold: usize,
> },
> ```
> 
> Should we expose this field too so that users can query the satifiable item(s)?
> 
> I'm just not clear on this API and I want to make sure we get you what you need.
> 
@thunderbiscuit 
I have now exposed that field too.

BitcoinZavior:
> 
> I'm just not clear on this API and I want to make sure we get you what you need.
> 
@thunderbiscuit Yes this is what we need. An integration test has been added here for the use case using the new exposed API:
https://github.com/LtbLightning/bdk-flutter/pull/147/files#diff-7c1bbba299e1163f3f58b2261196dbc0d886e22f40451fdb463ca1409d2b9ef6

thunderbiscuit:
Thanks for the work @BitcoinZavior! I squashed and rebased. Merging now.

thunderbiscuit:
For some reason I can't force push to your branch. But I added your commit in #626.

thunderbiscuit:
Merged in #629. Thanks for the work @BitcoinZavior!

reez:
Awesome work @BitcoinZavior !!!

PR #622, opened 2024-11-05
Bump uniffi-rs to 0.28.0

jurvis:
### Description

Bumps uniffi-rs to 0.28.0

#### All Submissions:


reez:
Thanks Jurvis!!

It looks like the CI failure is due to `clap_derive` increasing its Minimum Supported Rust Version (MSRV) to 1.74 (and some of these are failing because they use 1.71 or 1.67 etc). The bdk-ffi master branch looks like its set to Rust 1.77.1 for these various CI configurations so I think bumping the Rust version to 1.74 or higher in any of these failures should be perfectly fine for this PR, but open to if anyone thinks any differently or if I'm missing something else?

notmandatory:
I created #624 which includes the commit in this PR and adds fixes to CI so tests pass.

PR #580, opened 2024-07-31
Add finalize to Psbt

andreasgriffin:
### Description

Add serialize() to finalize.  See https://github.com/bitcoindevkit/bdk-ffi/issues/469

#### All Submissions:

* [ ] I've signed all my commits
* [ ] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [ ] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:

* [ ] I've added tests for the new feature
* [ ] I've added docs for the new feature

#### Bugfixes:

* [ ] This pull request breaks the existing API
* [ ] I've added tests to reproduce the issue which are now passing
* [ ] I'm linking the issue being fixed by this PR

Issue #469, opened 2024-03-14
Question: Finalizing a fully signed PSBT


andreasgriffin:
I have the following  fully signed (regtest) PSBT
```
[...]

PR #630, opened 2024-11-16
Add `finalize` to `Psbt`

rustaceanrob:
### Description

#580 does the initial work to adding `finalize` to `Psbt`. I would like to get this in for @andreasgriffin and revisit this later in `bitcoin-ffi`. I think #580 has the right idea, but I would like a more detailed system of _why_ finalizing might have failed. Unfortunately, the `finalize` function using a `Vec<E>` of errors, which is not expressible by UniFFI types, so I created a new `FinalizedPsbtResult` type that contains either the original `Psbt` or the finalized `Psbt`, and any errors that may be associated with the `finalize` step.

I have thought about this for a while and I think this is the best approach, but open to any ideas.

### Notes to the reviewers

Please confirm this works with your example `Psbt` @andreasgriffin 

### Changelog notice

- Add `finalize` method to `Psbt`

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -259,6 +259,13 @@ interface PsbtParseError {
 };
 
 [Error]
+interface PsbtFinalizeError {
+  InputError(string reason, u32 index);
+  WrongInputCount(u32 in_tx, u32 in_map);
+  InputIdxOutofBounds(u32 psbt_inp, u32 requested);
+};
+
+[Error]
 interface RequestBuilderError {
   RequestAlreadyConsumed();
 };
@@ -958,7 +965,15 @@ interface Psbt {
   [Throws=PsbtError]
   Psbt combine(Psbt other);
 
+  FinalizedPsbtResult finalize();
+
   string json_serialize();
+};
+
+dictionary FinalizedPsbtResult {
+  Psbt psbt;
+  boolean could_finalize;
+  sequence<PsbtFinalizeError>? errors;
 };
 
 dictionary TxIn {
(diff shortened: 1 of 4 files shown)


andreasgriffin:
Thanks!  Yes I can confirm that it finalizes the PSBT in  https://github.com/bitcoindevkit/bdk-ffi/issues/469  correctly.

reez:
> but I would like a more detailed system of why finalizing might have failed

Same, more details are great for things like this

> Unfortunately, the finalize function using a Vec<E> of errors, which is not expressible by UniFFI types, so I created a new FinalizedPsbtResult type that contains either the original Psbt or the finalized Psbt, and any errors that may be associated with the finalize step.

This makes sense to me.

Concept ACK c432e2d975c6d5b9fbca521c36eff37d0790540b

Built and tested the swift bindings locally but didn't get to finish putting it thru a few different scenarios to test it out more (not a blocker on the PR, just ran out of time while taking a quick look at it tonight), did get this as expected when testing out a scenario where I expected `couldFinalize` to be false 👍:

```
finalized: FinalizedPsbtResult(psbt: BitcoinDevKit.Psbt, couldFinalize: false, errors: Optional([BitcoinDevKit.PsbtFinalizeError.InputError(reason: "Could not satisfy Tr descriptor", index: 0)]))
```

thunderbiscuit (approved):

thunderbiscuit:
One thought I just had is that if this is not really agreed upon by other users of bitcoin-ffi, we'll end up with a breaking change if we migrate our Psbt type to the bitcoin-ffi type, or we'll have to just keep supporting our custom Psbt type.

PR #632, opened 2024-11-27
Build Kotlin API docs in CI

thunderbiscuit:
This PR is WIP.

It should allow us to trigger the build for the API docs for JVM and Android in the CI, and the resulting website could be downloaded as an artifact (to then be PR'd on the website).


thunderbiscuit:
The [run here is an example of the workflow that builds and exports the website as a downloadable artifact](https://github.com/thunderbiscuit/bdk-ffi/actions/runs/12053377504).

PR #637, opened 2024-12-12
deps: bump bdk to beta_6

reez:
### Description

This pull request updates the bdk-ffi crate to use the latest beta version of bdk, specifically version [1.0.0-beta.6](https://github.com/bitcoindevkit/bdk/releases/tag/v1.0.0-beta.6). 

The main changes include:
- Updating the Cargo.toml file.
- Add `InvalidResponse` to Esplora [Error](https://docs.rs/esplora-client/0.11.0/esplora_client/enum.Error.html) 
- Adding `transitively` to [ChainPosition](https://docs.rs/bdk_chain/0.21.0/bdk_chain/enum.ChainPosition.html)
- `FullScanResult` -> `FullScanResponse`
- `SyncResult` -> `SyncResponse`
- made a small change, unrelated to beta 6, to `error.rs` to use `BdkEsploraError` in places where it used `Error`, both work but it keeps a consistent convention (and allowed us to remove `Error` from `use bdk_esplora::esplora_client::{Error as BdkEsploraError, Error}`)... kinda hate throwing little things like this in a PR like this but since I was already updating that error I felt like it was worth it and not too confusing for reviewers hopefully

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/Cargo.lock
+++ b/bdk-ffi/Cargo.lock
@@ -4,12 +4,6 @@ version = 3
 
 [[package]]
 name = "ahash"
-version = "0.4.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0453232ace82dee0dd0b4c87a59bd90f7b53b314f3e0f61fe2ee7c8a16482289"
-
-[[package]]
-name = "ahash"
 version = "0.8.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
@@ -181,34 +175,33 @@ dependencies = [
 
 [[package]]
 name = "bdk_chain"
-version = "0.20.0"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3bee1fe68ec0015bce2e4c1754ebbf18d70750a1f0103e3785d34e8959fe8fd7"
+checksum = "65db19f8952e6563c7c0ce38190b930ac57c02bb9f0da4f761e542639f99ed72"
 dependencies = [
  "bdk_core",
  "bitcoin",
  "miniscript",
  "rusqlite",
  "serde",
- "serde_json",
 ]
 
 [[package]]
 name = "bdk_core"
-version = "0.3.0"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "be5e187ee33d5f99f1997a700cc1dfa0524fd1de31e6414c612c9e89ccdaa133"
+checksum = "070ff23e275c978ca19d137a2ae3516262400868a04228394710b270d958afc6"
 dependencies = [
  "bitcoin",
- "hashbrown 0.9.1",
+ "hashbrown",
  "serde",
 ]
 
 [[package]]
 name = "bdk_electrum"
-version = "0.19.0"
+version = "0.20.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fee5560d6758855e5e0778bc3bc25648658684afac667321ddc0cba08aef933a"
+checksum = "8cda5b8cd1175044dc6cc5380b34954f0269ba81719fb238e15fa515b98f200b"
 dependencies = [
  "bdk_core",
  "electrum-client",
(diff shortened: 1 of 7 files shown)


rustaceanrob:
LGTM. Would ACK once out of draft.

reez:
Tests pass, next step is to build bindings locally and test out on BDK iOS, then I'll un-draft PR.

bdk-ffi/src/lib.rs
```diff
@@ -79,6 +79,7 @@ use bitcoin_ffi::FeeRate;
 use bitcoin_ffi::Network;
 use bitcoin_ffi::OutPoint;
 use bitcoin_ffi::Script;
+use bitcoin_ffi::Txid;
```
thunderbiscuit, line 82: I don't particularly like the `Txid` type from bitcoin-ffi and think we should not use it. See https://github.com/bitcoindevkit/bitcoin-ffi/issues/20 for details. In this case here it's a return value from the Rust side so we know it will be correct, but even so, I think using the type in its current form is not a good idea. Open to hear other people's thoughts on this, but IMO we should stick to `String` for txids for now.

reez:
> Tests pass, next step is to build bindings locally and test out on BDK iOS, then I'll un-draft PR.

Build swift bindings locally and used in BDK iOS which still compiled https://github.com/bitcoindevkit/BDKSwiftExampleWallet/pull/264

thunderbiscuit (commented):
I'm still working on my understanding of transitively anchored transactions and when exactly those can occur in your TxGraph, so I won't ACK just yet but this is looking good to me.

One small comment about our use of `Txid`.

bdk-ffi/src/lib.rs
rustaceanrob: Using the type allows for a more robust API in terms of traits. I directly export `NodeEventHandler` straight from `bdk_kyoto` in my PR which uses the actual `Txid`, but it gets coerced to a string for UniFFI users. I find that particularly nice, because Rust design choices have no effect on bindings in that scenario. 

`ldk-node` basically uses string type aliases for almost every struct and event (i.e. payment invoice, preimage, channel ID). I think the correct approach is to use the abstraction if we are confident it is correct.

thunderbiscuit: My beef with it is the uncatchable, (runtime!) panic of your whole app if you or your users attempt for whatever reason to build one of those types incorrectly (a txid missing a letter at the end? The whole app collapses and you can't prevent it!).

@tnull how do you see this? Do you mitigate for this in any way? It just feels like a super dangerous type to have for bindings, and is the opposite of type-safety, because the type does _not_ behave the way you'd want/expect it to, but the fact that it exists implies type-safety, whereas if you have a String it's clear that it's not type-safe, and you have at least that small setting of expectations.

To be clear I'm not against a Txid type at all, I just would like something that can throw an error if you try to build an incorrect one.

reez: FYI: 

Moving to `string` for this PR. Then when we decide *if* we want to go with `Txid` we can have a follow up PR to change all instances of txid `string` to `Txid` type across the codebase.

Discussion can definitely continue here.

PR is ready to go for merge though.

rustaceanrob: > you or your users attempt for whatever reason to build one of those types incorrectly

There is a big different here between using `Txid` as a return value and an argument. I could see the argument is strong for user-supplied values from potentially untrusted sources. Server provides "" for a response? Unlucky. However for return values that `bdk-ffi` constructs, are we sure the user can even mutate these? It seems abundantly safe to me to return `Txid` as opposed to `string` if we expect the user will treat it as a `Txid`, _especially_ if they aren't allow to mutate `Txid`

thunderbiscuit: I agree with the fact that since it is returned, those specific txids it will not cause problems for the users. The problem (as I imagine it at least) is that the type now exists freely at the library level and users can create objects of that type themselves; it's implied that we believe in and support this type as representative of a `Txid` (maybe with the added complication that users familiar with the rust-bitcoin library will think it's a great, well thought-out type when really it's just a string but I'm willing to disregard this for a moment).

Devs wanting to write typesafe software are likely to just start using it in other places (maybe in data structures they fully own), and the type will work more-or-less for them (they can build a txid `abcd` and nothing will get thrown). This is _a little bad_ (because they think they're being typesafe but there are no checks for this Txid being valid in any way), but not _terrible_ (as long as they don't try to use it in a bdk method their apps will not crash and burn). Over time however, if we widen our use of the type and start using it as an argument in methods/constructors, then... kaboom.

thunderbiscuit (approved):
ACK 10c4d25486cfcde74c77d00c6f677f453411e979.

PR #641, opened 2024-12-18
Add `ElectrumClient::server_features` method

thunderbiscuit:
This PR adds two new methods on the electrum client:
1. [`server_features`](https://docs.rs/electrum-client/0.22.0/electrum_client/trait.ElectrumApi.html#tymethod.server_features). Also exposed is the [`ServerFeaturesRes`](https://docs.rs/electrum-client/0.22.0/electrum_client/struct.ServerFeaturesRes.html) struct.
2. [`estimate_fee`](https://docs.rs/electrum-client/0.22.0/electrum_client/trait.ElectrumApi.html#tymethod.estimate_fee)

TODO:
- [x] Changelog notice
- [x] Swift test
- [x] API docs

### Notes to the reviewers

I added a live test with a testnet electrum server. Note that my test with our standard mempool.space signet server returned an error (with a nice error message!) saying the server did not support the `server_features` method.

Note that those 2 tests don't run in the CI here because they are live tests, but I have run them both locally and they work well.

You can run these APIs locally in a Kotlin script:
```shell
cd bdk-jvm
just build-macos-aarch64
just publish-local

# cd ~/temp-or-whatever/
touch fees.main.kts
kotlin fees.main.kts
```

**fees.main.kts**
```kotlin
// fees.main.kts

@file:Repository("file:///~/.m2/repository/")
@file:DependsOn("org.bitcoindevkit:bdk-jvm:1.0.0-beta.6-SNAPSHOT")

import org.bitcoindevkit.ServerFeaturesRes
import org.bitcoindevkit.ElectrumClient

val electrumClient: ElectrumClient = ElectrumClient("ssl://electrum.blockstream.info:50002")
val features: ServerFeaturesRes = electrumClient.serverFeatures()
println("Server Features:\n$features")

val fees6 = electrumClient.estimateFee(6uL)
println("Fees to be in the next block: $fees6")
```

### Changelog notice

```md
Added:
  - ElectrumClient::server_features [#641]
  - ServerFeaturesRes struct [#641]
  - ElectrumClient::estimate_fee [#641]

[#641]: https://github.com/bitcoindevkit/bdk-ffi/pull/641
```

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -1104,6 +1104,10 @@ interface ElectrumClient {
   /// Returns the capabilities of the server.
   [Throws=ElectrumError]
   ServerFeaturesRes server_features();
+
+  /// Estimates the fee required in bitcoin per kilobyte to confirm a transaction in `number` blocks.
+  [Throws=ElectrumError]
+  f64 estimate_fee(u64 number);
 };
 
 /// Response to an ElectrumClient.server_features request.
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -1054,29 +1054,76 @@ interface EsploraClient {
 // bdk_electrum crate
 // ------------------------------------------------------------------------
 
+/// Wrapper around an electrum_client::ElectrumApi which includes an internal in-memory transaction
+/// cache to avoid re-fetching already downloaded transactions.
 interface ElectrumClient {
+  /// Creates a new bdk client from a electrum_client::ElectrumApi
   [Throws=ElectrumError]
   constructor(string url);
 
+  /// Full scan the keychain scripts specified with the blockchain (via an Electrum client) and
+  /// returns updates for bdk_chain data structures.
+  ///
+  /// - `full_scan_request`: struct with data required to perform a spk-based blockchain client
+  ///   full scan, see `FullScanRequest`.
+  /// - `stop_gap`: the full scan for each keychain stops after a gap of script pubkeys with no
+  ///   associated transactions.
+  /// - `batch_size`: specifies the max number of script pubkeys to request for in a single batch
+  ///   request.
+  /// - `fetch_prev_txouts`: specifies whether we want previous `TxOuts` for fee calculation. Note
+  ///   that this requires additional calls to the Electrum server, but is necessary for
+  ///   calculating the fee on a transaction if your wallet does not own the inputs. Methods like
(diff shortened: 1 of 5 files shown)


reez:
Starting to review this! Will add comments and have this review finished in the next day-

reez:
Thanks for the helpful guide to test this locally like this, I ran it, and one comment just to double check I'm on the same page with how this works is `println("Fees to be in the next block: $fees6")` should be something more like `println("Estimated fee rate for confirmation within 6 blocks: $fees6`, since `fees6` is requesting an estimate for 6 blocks instead of for the next block. No code changes or anything necessary obviously since its not code in the PR but just for local testing, but more bringing it up in case I'm misunderstanding something.

> ```kotlin
> // fees.main.kts
> 
> @file:Repository("file:///~/.m2/repository/")
> @file:DependsOn("org.bitcoindevkit:bdk-jvm:1.0.0-beta.6-SNAPSHOT")
> 
> import org.bitcoindevkit.ServerFeaturesRes
> import org.bitcoindevkit.ElectrumClient
> 
> val electrumClient: ElectrumClient = ElectrumClient("ssl://electrum.blockstream.info:50002")
> val features: ServerFeaturesRes = electrumClient.serverFeatures()
> println("Server Features:\n$features")
> 
> val fees6 = electrumClient.estimateFee(6uL)
> println("Fees to be in the next block: $fees6")
> ```

reez (approved):
ACK 01a64d16123e07f6390b22f4792d3e4dd11b1fb8

A few comments/thoughts

- I ran the local tests and the live tests for bdk-jvm (thanks for also adding the companion swift live test), was kind of fun to run those and see the output

- Thanks for adding the code comments in the UDL for all of `ElectrumClient`

- The companion to the `estimate_fee` method but for `EsploraClient` would be [get_fee_estimates](https://docs.rs/esplora-client/latest/esplora_client/blocking/struct.BlockingClient.html#method.get_fee_estimates)? (If so, that something I can look into adding at some point, the only reason I was wondering/asking is just because I got curious what the similar method would be for EsploraClient when I was reviewing this)

thunderbiscuit:
Yeah it's too bad the Esplora equivalent is not just a 1-for-1 match. Previously the clients were all adhering to a trait in the bdk crate so they all had that method on them (the inner implementation was a little different but I don't think we needed to do anything on our end if I remember correctly). In this case we'll have to find exactly how to do it for Esplora to offer it there (and you're right I don't think it's the same method name!).

PR #647, opened 2025-01-08
deps: bump bdk to 1.0.0

reez:
#### All Submissions:

#### New Features:

#### Bugfixes:


bdk-ffi/Cargo.toml
```diff
@@ -18,10 +18,10 @@ path = "uniffi-bindgen.rs"
 default = ["uniffi/cli"]
 
 [dependencies]
-bdk_wallet = { version = "=1.0.0-beta.6", features = ["all-keys", "keys-bip39", "rusqlite"] }
-bdk_core = { version = "0.4.0" }
-bdk_esplora = { version = "0.20.0", default-features = false, features = ["std", "blocking", "blocking-https-rustls"] }
-bdk_electrum = { version = "0.20.0", default-features = false, features = ["use-rustls-ring"] }
+bdk_wallet = { version = "=1.0.0", features = ["all-keys", "keys-bip39", "rusqlite"] }
```
thunderbiscuit, line 21: I think we should be able to remove the `=` sign here and let cargo do its thing since new feature versions should not introduce breaking changes.

bdk-ffi/Cargo.toml
```diff
@@ -18,10 +18,10 @@ path = "uniffi-bindgen.rs"
...
+bdk_core = { version = "0.4.1" }
+bdk_esplora = { version = "0.20.1", default-features = false, features = ["std", "blocking", "blocking-https-rustls"] }
+bdk_electrum = { version = "0.20.1", default-features = false, features = ["use-rustls-ring"] }
 bitcoin-ffi = { git = "https://github.com/bitcoindevkit/bitcoin-ffi", tag = "v0.1.2" }
 
 uniffi = { version = "=0.28.0" }
```
thunderbiscuit, line 27: Don't do that for uniffi though!

thunderbiscuit (approved):
ACK 93edca7cb57c2620ffcd301b7c3348e2dfae2f7d.

Maybe one little change if you agree with it, otherwise it's good to go either way.

bdk-ffi/Cargo.toml
reez: updated to remove the `=`

PR #664, opened 2025-02-05
feat: expose ElectrumClient::block_headers_subscribe method

thunderbiscuit:
This is a bit of a cheat: the method is not intended for a 1-time use, but rather is meant to open a subscription. In any case the Electrum client does return what I think our user wants (LDK-node please confirm!).

I'm not ready to merge this one just yet because I actually didn't expose the [`Header`](https://docs.rs/bitcoin/0.32.5/bitcoin/blockdata/block/struct.Header.html) type fully (which would require 4 new types: `Version`, `BlockHash`, `TxMerkleNode`, and `CompactTarget`). If we do want to expose the [`block_header()`](https://docs.rs/electrum-client/latest/electrum_client/trait.ElectrumApi.html#method.block_header) method, then it might actually be worth taking the time exposing the whole `Header` type from rust-bitcoin.

Let me know if you have an opinion either way, because this would be a breaking change if we expose a simplified version and eventually want to return the full Header.

### Changelog

```md
Added:
    - Expose ElectrumClient::block_headers_subscribe method [#664]

[#664]: https://github.com/bitcoindevkit/bdk-ffi/pull/664


reez:
> I'm not ready to merge this one just yet because I actually didn't expose the [Header](https://docs.rs/bitcoin/0.32.5/bitcoin/blockdata/block/struct.Header.html) type fully (which would require 4 new types: Version, BlockHash, TxMerkleNode, and CompactTarget). If we do want to expose the [block_header()](https://docs.rs/electrum-client/latest/electrum_client/trait.ElectrumApi.html#method.block_header) method, then it might actually be worth taking the time exposing the whole Header type from rust-bitcoin.

> Let me know if you have an opinion either way, because this would be a breaking change if we expose a simplified version and eventually want to return the full Header.

Thanks for getting this PR going!

Yeah this solves the specific use case by providing the information needed/asked. It would be great to expose `Header` fully, but would also definitely be nice to have other specific use cases to help clarify how important it is to expose `Header` fully, I'm personally not using Electrum nearly as much as Esplora so I don't personally have a strong feeling (balancing this is what known user needs vs default of opening up types fully + not having future breaking changes)

thunderbiscuit:
We should also note that the Electrum client is not part of our "semantic 1.0", so if we do end up with small breaking changes there it's ok. Doesn't mean we shouldn't try to get it right on the first shot, but I just thought I'd mention it.

reez:
> I looked at the Header type and in the end it wasn't actually too bad to expose. Most fields on it are typesafe things that simply wrap a u32 or i32, so it was actually easy to just pull those out.
> 
> I think this strikes a good balance of availability/complexity. Let me know what you think @reez @ConorOkus. It think we could add this to the release candidate 1 coming up.

agree this strikes a good balance of those, nice job!

reez (approved):
ACK 4bcc17c3adcb1ec5b247d6853658a4d673addc98

rustaceanrob:
My question here would be what happens when a client subscribes to headers twice by calling this method twice. Would the server get mad? I think the test could be modified to call the method twice and see if that complicates anything?

PR #666, opened 2025-02-06
feat: add get_tx_status to esplora

reez:
#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -1120,6 +1120,10 @@ interface EsploraClient {
   /// blocks) and the value is the estimated feerate (in sat/vB).
   [Throws=EsploraError]
   record<u16, f64> get_fee_estimates();
+
+  /// Get the status of a [`Transaction`] given its [`Txid`].
+  [Throws=EsploraError]
+  TxStatus get_tx_status(string txid);
 };
 
 // ------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


reez:
rebased so ready when ack'd

thunderbiscuit (approved):
Easy win! ACK 623525a75128e9087e2ac6eb875d4c244252b5f7.

I'm a bit surprised by the `Tx` type in the Esplora client (why not use the `Transaction` type from rust-bitcoin?) but that's a question for the Esplora client not for this PR.

It opens up a question we'll have to answer at some point as we keep opening up more and more types: what do we do with very similar types (particularly if their names are the same!)? One of the weaknesses of uniffi is its lack of namespacing. In theory I'd love something like

```kotlin
import org.rustbitcoin.Transaction
import org.bitcoindevkit.esplora.Transaction
```

So I could differentiate between them cleanly, but right now that's not possible, so at the root level of bdk-android/jvm/whatever you have both `Tx` and `Transaction`. This would typically not be good practice in library development.

reez:
> Easy win! ACK [623525a](https://github.com/bitcoindevkit/bdk-ffi/commit/623525a75128e9087e2ac6eb875d4c244252b5f7).
> 
> I'm a bit surprised by the `Tx` type in the Esplora client (why not use the `Transaction` type from rust-bitcoin?) but that's a question for the Esplora client not for this PR.
> 
> It opens up a question we'll have to answer at some point as we keep opening up more and more types: what do we do with very similar types (particularly if their names are the same!)? One of the weaknesses of uniffi is its lack of namespacing. In theory I'd love something like
> 
> ```kotlin
> import org.rustbitcoin.Transaction
> import org.bitcoindevkit.esplora.Transaction
> ```
> 
> So I could differentiate between them cleanly, but right now that's not possible, so at the root level of bdk-android/jvm/whatever you have both `Tx` and `Transaction`. This would typically not be good practice in library development.

These are great thoughts though so I'm definitely glad you added them here! Made me think a bit more about this as well.

PR #671, opened 2025-02-12
feat: add to_address_data to address type

### LINKED ISSUE #669: Add Address.payload() method
**Describe the enhancement**  

Bring back `Address.payload`, item from #522 

**Use case**  
 

**Additional context**

reez:
### Description

Resolve #669 with [to_address_data](https://docs.rs/bitcoin/latest/bitcoin/struct.Address.html#method.to_address_data)

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -6,13 +6,14 @@ use crate::error::{
 use bitcoin_ffi::OutPoint;
 use bitcoin_ffi::Script;
 
-use bdk_wallet::bitcoin::address::{NetworkChecked, NetworkUnchecked};
+use bdk_wallet::bitcoin::address::NetworkChecked;
+use bdk_wallet::bitcoin::address::NetworkUnchecked;
+use bdk_wallet::bitcoin::address::{Address as BdkAddress, AddressData as BdkAddressData};
 use bdk_wallet::bitcoin::consensus::encode::serialize;
 use bdk_wallet::bitcoin::consensus::Decodable;
 use bdk_wallet::bitcoin::io::Cursor;
 use bdk_wallet::bitcoin::psbt::ExtractTxError;
 use bdk_wallet::bitcoin::secp256k1::Secp256k1;
-use bdk_wallet::bitcoin::Address as BdkAddress;
 use bdk_wallet::bitcoin::Network;
 use bdk_wallet::bitcoin::Psbt as BdkPsbt;
 use bdk_wallet::bitcoin::Transaction as BdkTransaction;
@@ -25,6 +26,19 @@ use std::fmt::Display;
 use std::ops::Deref;
 use std::str::FromStr;
 use std::sync::{Arc, Mutex};
+
+#[derive(Debug)]
+pub enum AddressData {
+    P2pkh { pubkey_hash: String },
+    P2sh { script_hash: String },
+    Segwit { witness_program: WitnessProgram },
+}
+
+#[derive(Debug)]
+pub struct WitnessProgram {
+    pub version: u8,
+    pub program: Vec<u8>,
+}
 
 #[derive(Debug, PartialEq, Eq)]
 pub struct Address(BdkAddress<NetworkChecked>);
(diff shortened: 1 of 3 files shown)


reez:
@thunderbiscuit 
- thoughts or better options for this line https://github.com/bitcoindevkit/bdk-ffi/pull/671/files#diff-c73020e295d654c61c706f3634d67285896d23813c1fb4f5c30ff57c220e154dR94 ?
- I tested this and can push up the test/tests I ran for this if helpful too

bdk-ffi/src/bitcoin.rs
```diff
@@ -26,6 +29,19 @@ use std::ops::Deref;
 use std::str::FromStr;
 use std::sync::{Arc, Mutex};
 
+#[derive(Debug)]
+pub enum AddressData {
+    P2pkh { pubkey_hash: Vec<u8> },
```
thunderbiscuit, line 34: I was thinking that maybe what people want here is the hash in string format? Really it's not a problem because I can transform the bytes into a hex locally but I think the `Hash` type maybe handles some of that (big-endianness and little-endianness) when converting to strings? We also return a string rather than the bytes when giving out the Txid (which is also a `Hash`).

Let me know what you think.

thunderbiscuit (commented):
Looks great!

Only one small comment/idea.

bdk-ffi/src/bitcoin.rs
reez: Yeah I agree with what you're saying but I was also trying to balance it with how Conor was using it (which was the initial reason for exposing this method) which he mentioned was this:

```
"Need it to to generate a valid shutdown script for closing Lightning channels"

override fun get_shutdown_scriptpubkey(): Result_ShutdownScriptNoneZ {
        val address = ldkkeysManager!!.wallet.getLastUnusedAddress().address

        return when (val payload: Payload = address.payload()) {
            is Payload.WitnessProgram -> {
                val ver = when (payload.version.name) {
                    in "V0".."V16" -> payload.version.name.substring(1).toIntOrNull() ?: 0
                    else -> 0 // Default to 0 if it doesn't match any "V0" to "V16"
                }

                val result = ShutdownScript.new_witness_program(
                    WitnessProgram(
                        payload.program.toUByteArray().toByteArray(),
                        WitnessVersion(ver.toByte())
                    )
                )
                Result_ShutdownScriptNoneZ.ok((result as Result_ShutdownScriptInvalidShutdownScriptZ.Result_ShutdownScriptInvalidShutdownScriptZ_OK).res)
            }

            else -> {
                Result_ShutdownScriptNoneZ.err()
            }
        }
    }

```

What are your thoughts with that added context?

thunderbiscuit: Yeah the witness program I'd keep as bytes. I was more thinking of the `pubkey_hash` and `script_hash` as hex strings the way we often see/think of hashes.

thunderbiscuit: But honestly I'm ok with Byte arrays! Just a thought.

reez:
> Looks great!
> 
> Only one small comment/idea.
> 
> As for the non-exhaustive, if you add that qualifier to the UDL does it let you do away with the `_ => unimplemented` branch?

Did you happen to get this to compile if you tried this, the only reason I'm asking is because I somehow could not get it to compile (error `NonExhaustive not supported for interface definition`) if I tried this even though to my understanding it actually should compile based on https://mozilla.github.io/uniffi-rs/0.28/udl/enumerations.html#remote-non-exhaustive-enums

bdk-ffi/src/bitcoin.rs
reez: Ah! yeah no what you're saying makes perfect sense to me now. `pubkey_hash` and `script_hash` have been updated/changed to `string`s

thunderbiscuit:
@reez ah yes you're right. This is only possible with simple enums, not interface-based enums (which turn into Sealed Classes in Kotlin, can't remember what in Swift).

Ok this is good to go IMO!

thunderbiscuit (approved):
ACK 889b70ab8684aef5e33fda01b1bd14d8c2def938.

But yes if you have maybe a test showing all 3 variants of the enum (even if it's just in Swift) that'd be great I think!

reez:
> ACK [889b70a](https://github.com/bitcoindevkit/bdk-ffi/commit/889b70ab8684aef5e33fda01b1bd14d8c2def938).
> 
> But yes if you have maybe a test showing all 3 variants of the enum (even if it's just in Swift) that'd be great I think!

(The test I had was a rust test in the `bitcoin.rs` file so I pushed that up and will merge now)

reez:
@ConorOkus this is merged and should be in our next release

PR #685, opened 2025-02-25
Make justfiles awesome

thunderbiscuit:
This PR brings in significant improvements to the justfiles.

See for yourself by running the `just` command in all 4 main directories, which will list the available commands well grouped and documented. Of note are the cleaning up of the ever-tiresome `build` commands which had to be qualified per architecture. The command now accepts an argument you can pass for the architecture you want to build on, but also has a default argument of `macos-aarch64`, which is I think what the devs on this team use the most. You can therefore call `just build` if you are on a mac with silicon chip and it will work.

### DIFF
--- a/.github/workflows/test-swift.yaml
+++ b/.github/workflows/test-swift.yaml
@@ -24,4 +24,4 @@ jobs:
 
       - name: "Run Swift tests"
         working-directory: bdk-swift
-        run: swift test --skip LiveElectrumClientTests --skip LiveMemoryWalletTests --skip LiveTransactionTests --skip LiveTxBuilderTests --skip LiveWalletTests
+        run: swift test --filter Offline
--- a/bdk-android/justfile
+++ b/bdk-android/justfile
@@ -1,29 +1,45 @@
-default:
-  just --list
+[group("Repo")]
+[doc("Default command; list all available commands.")]
+@list:
+  just --list --unsorted
 
-build-linux:
-  bash ./scripts/build-linux-x86_64.sh
+[group("Repo")]
+[doc("Open repo on GitHub in your default browser.")]
+repo:
+  open https://github.com/bitcoindevkit/bdk-ffi
 
-build-macos:
-  bash ./scripts/build-macos-aarch64.sh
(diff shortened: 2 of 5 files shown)


reez:
very cool and clean+organized.

is there any way to get a list of architectures that are available to pass to `just build ARCH=xyz` (without having to jump out of the `just` world, because obv its possible to just look at the `scripts` folder or run `ls ./scripts/` or something but that steps out of the `just` world for a moment)

the most basic way would be something like this maybe (I'm sure there's better+less-hardcoded ways) :

```
[group("Build")]
[doc("List available architectures for the build command.")]
list-architectures:
    echo "Available architectures:"
    echo " - linux-x86_64"
    echo " - macos-aarch64"
    echo " - macos-x86_64"
    echo " - windows-x86_64"
```

not as needed for `swift`, but nice for `jvm` and such

thunderbiscuit:
Great idea! I cleaned up a bit what you had just so it would print nicely, but yes that's totally needed.

rustaceanrob:
Something to note for swift is you can pass the --filter option to filter test names on an argument. Might be useful to emulate test-specific like in the Kotlin case

thunderbiscuit:
After all this time... how did I just learn about this today hahahaha. This is great and will allow us to use the simpler 

```shell
swift test --filter Offline
```

line in our CI, also making it easy to add live tests that won't have to be added manually to the line in the yaml file.

You also made me realize that I can use the `test` command with an optional argument of TEST, running a specific test if provided but otherwise running all of them. This allows us to remove the `test-specific` command.

I just pushed the following new workflows:

```shell
# Swift
just test
just test OfflinePersistenceTests
just test Live

# JVM
just test
just test OfflineWalletTest
```

rustaceanrob:
Without knowing much beyond the basics of `just`, looks like an improvement to me.

ACK 8f6eedde7617b136ca50cd6c92565f2ad343ad80

reez (approved):
ACK 8f6eedde7617b136ca50cd6c92565f2ad343ad80

I dig it!

Issue #687, opened 2025-03-02
Beef up the Descriptor type


thunderbiscuit:
The `Descriptor` type is one of the most complex types coming from BDK/Miniscript, and we expose a simplified version of what's available in Rust, partly because it's a whole lot of generics on top of generics.

 I still need to play around with the Rust code to figure out a few things we could add to make the type more powerful in the bindings, but the first thing that I realized only today is that given a descriptor, there is currently not a lot of introspection available into it.

For example, at any point your `Descriptor` probably came from either the domain layer (you created it yourself from entropy or a recovery phrase) and used our descriptor templates, in which case you know the script type (P2TR, P2WPKH, etc.), _or_ it was given to you by a user in the form of a string. In _this_ case, you can create the descriptor, but cannot query it directly for metadata (script type, checksum, etc.). The only way to do this currently is to manually parse the string, but surely we can do better because those methods are available in some form in Rust.

thunderbiscuit:
See #688 for an example of what I have in mind.

thunderbiscuit:
Related to #357.

cc @andreasgriffin

andreasgriffin:
Really great would be:
```
DescriptorPublicKey(...).xpub
DescriptorPublicKey(...).key_origin
DescriptorPublicKey(...).derivation_path
 
```
master fingerprint was already exposed in [master_fingerprint](https://docs.rs/bdk_wallet/latest/bdk_wallet/descriptor/enum.DescriptorPublicKey.html#method.master_fingerprint)

thunderbiscuit:
Closing this for now. If we need more we should open up specific issues on the APIs we need!

PR #688, opened 2025-03-02
Add `Descriptor::desc_type` method

thunderbiscuit:
Proof of Concept. Will keep iterating. The parsing of some miniscripts that are according to the BIP valid also fails, and I don't know exactly why but assume it's because those are not supported by the miniscript crate? Not sure yet (you can see some of the tests I commented out).

Descriptors taken from the BIPs here: https://github.com/bitcoin/bips/blob/master/bip-0380.mediawiki

See #687.


thunderbiscuit:
This is finally ready for review/merge. 

There is much more to do with the `Descriptor` type, but I don't think it should block this one from moving forward. Non-breaking change.

PR #690, opened 2025-03-08
Switch to proc macros module Kyoto

rustaceanrob:
### Description

Migrates all `kyoto` related types to proc macros. The remote enum `NodeState` remains exported by the UDL as it would require a repeated definition in the `kyoto` module.

### Changelog notice

- Migrate `kyoto` module to proc macros

#### All Submissions:

#### New Features:

#### Bugfixes:


thunderbiscuit:
I was hoping we'd be able to do a few smaller ones first to get everyone on board and familiar with that transition, but it turns out that it's hard to do it one little piece at a time, because the UDL is connected to it all and you can't have a UDL that doesn't parse on its own, and you can't declare both a proc macro and a UDL. So for example you can't just pull out the `FeeRate` type as a single migration, because it's being returned and used in other types.

It looks like the clients are maybe some of the easier parts of the codebase to migrate first... thanks for doing this.

rustaceanrob:
> So for example you can't just pull out the FeeRate type as a single migration, because it's being returned and used in other types.

Interesting. If you want to change a type to proc macros, all instances where it is used and returned must also be proc macros? If that is the case then a module-by-module refactor will probably have to go down.

I was just doing this since the earlier I nipped it in the butt for Kyoto the easier it would be to add new types/methods without horrendous code churn.

Btw, I looked in IntelliJ at the JVM build and the doc strings are used when you define them above the types/methods in this way.

thunderbiscuit (approved):
ACK f46f743141758f346364be516293d9ed1aaa4350. The live Kyoto tests pass locally.

PR #692, opened 2025-03-11
Small post-release fixes

thunderbiscuit:
Our wallet was over 500 indices deep, which made tests slower to complete. 

Note that the descriptors associated with the persistence tests are not changed, as they would require creating and committing new sqlite files but the old ones are fine and the tests are not live anyway so don't require any sync.

Ok this is turning into a bundle of tiny fixes we just need and I don't feel like PRing individually.
- The first commit was something we addressed on the release/1.1 branch but I forgot to cherry pick back on master.
- The second commit bumps the descriptors used in tests.
- The third one adds the changelog entry for the 0.32.1 release.
- The fourth commit updates the readme files and API docs landing page for bdk-jvm and bdk-android
- The fifth commit updates the development version of the libraries. I'm using 1.2.0-dev for the Rust library, which is not a format recognized by Cargo but is recommended for these types of internal uses and when the library is not actually released on crates.io. This indicates it's an "in-development" version. Our other choices are to use `alpha.0`, but I found this odd because it's not an alpha at this point, it's literally just a development version, or to keep it at 1.1.0 until we release 1.2.0, and bump it then.
- The sixth commit adds the changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -161,6 +161,11 @@ This release is the first alpha release of the 1.0 API for the bindings librarie
 - Create and sign transactions using the transaction builder
 - Broadcast transactions
 
+## [v0.32.1]
+This is a patch release, updating the bindings libraries to bdk version 0.30.2, fixing an issue with syncing very large wallets.
+
+See https://github.com/bitcoindevkit/bdk/releases/tag/v0.30.2 for details.
+
 ## [v0.31.0]
 This release updates the bindings libraries to bdk version 0.29.0, updating rust-bitcoin to version 0.30.2.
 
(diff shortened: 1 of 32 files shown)


reez (approved):
ACK 07ca6229b296e5e52052cf1746963cd09fa198c3

ACKing before asked for review because it looks good to me and I was already reviewing another similar, but I'll take another look if more gets added, and keep an eye on tests passing

reez:
@thunderbiscuit tested this new descriptor on a separate local branch where I was doing some additional logging+testing around the recent live-test CI failures:

1. Previously, some of the tests around the crash/timeout were taking over 1 minute (I was able to reproduce the crash/timeout)
Now, those same tests are taking ~5 seconds.

2. I also can't reproduce the crash/timeout anymore using this new descriptor. So I think we're looking good.

3. Also before using this new descriptor, I also just started getting this on an Electrum test: 

```
"failed: caught error: "Protocol(errorMessage: "{\"code\":-32603, \ "mess age\":\"Too many history transactions (>500). Contact support to raise limits.\"
} " )"
```

Not sure if you saw that too and its why you mentioned "Our wallet was over 500 indices deep", but yeah, I dont get that error anymore with this new descriptor.

thunderbiscuit:
Yeah that was the same error I was getting.

PR #693, opened 2025-03-11
Migrate FeeRate to procedural macros

thunderbiscuit:
This PR simply migrates the FeeRate type to procedural macros. Notice that because it is used in the UDL more places than simply its definition, we must keep a single line indicating its definition in proc macros.

These lines can be removed once no reference to the FeeRate type are required in the UDL (for example once we migrate the `Wallet` type). 

I also fixed here the arguments to both constructors of the `FeeRate` type to [bring them in line with the Rust API](https://docs.rs/bitcoin/latest/bitcoin/struct.FeeRate.html).

#### All Submissions:

### DIFF
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -1475,19 +1475,7 @@ interface Amount {
   f64 to_btc();
 };
 
-interface FeeRate {
-  [Name=from_sat_per_vb, Throws=FeeRateError]
-  constructor(u64 sat_per_vb);
-
-  [Name=from_sat_per_kwu]
-  constructor(u64 sat_per_kwu);
-
-  u64 to_sat_per_vb_ceil();
-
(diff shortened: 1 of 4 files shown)


bdk-ffi/src/bdk.udl
```diff
@@ -1469,19 +1469,7 @@ interface Amount {
   f64 to_btc();
 };
 
-interface FeeRate {
```
rustaceanrob, line 1472: Should we have sort of policy when migrating to proc macros, if a type did not have documentation defined, a second commit introduces docs too? Seems like a good opportunity to get those out of the way

thunderbiscuit: I think that's a great idea.

reez (approved):
ACK 5fd3c71c203d446079fa94f35856aca3ac05d212

Just add docs in `bitcoin.rs` as planned before merge

PR #697, opened 2025-03-12
Fix live test for Kyoto in python

rustaceanrob:
### Description

I recently installed `mypy` and `ruff` for python linting and got a bunch of lints on the live Kyoto test:
1. Named imports are preferred over `*` 
2. The `update` method returns an optional
3. I never actually defined the `log_task` variable somehow, which is a log loop that prints to the console.

#### All Submissions:

### DIFF
--- a/.github/workflows/test-python.yaml
+++ b/.github/workflows/test-python.yaml
@@ -192,3 +192,21 @@ jobs:
 
       - name: "Run tests"
         run: python -m unittest discover --start "./tests/" --pattern "test_offline_*.py" --verbose
+
+  ruff:
+    name: "Lint tests"
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: bdk-python
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+
+      - name: "Install Ruff"
+        run: curl -LsSf https://astral.sh/ruff/install.sh | sh
+
+      - name: "Lint test targets"
+        run: ruff check ./tests/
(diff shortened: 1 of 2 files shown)


thunderbiscuit:
I like the idea of using linters, and we probably should have been using them for a while. 

Would you mind adding a commit that passes all python tests through the linter? I'm not super familiar with how the linters work on Python (or the different options, but I'll trust your choice of ruff), but I also assume there might be files to commit that define rules and whatnot? Feel free to commit those as well so that other users can run the linter too.

rustaceanrob:
> Would you mind adding a commit that passes all python tests through the linter?

Interesting, hadn't thought of adding them to CI. I should've been more specific, but I installed these as neovim plug-ins, so I will have to look into a `ruff` CLI tool or something like that. Do you mind if we punt that to another PR and keep this one as a test fix? The log function introduced in this could help debugging a failed CI

thunderbiscuit:
Oh funny enough I didn't mean adding the lint check in the CI but just running the other test files through haha. But I actually don't mind having it there either. However I feel like the CI runs will should probably all fail now (except for yours that you ran through the linter). 

Can you also add another commit that fixes the other test files (does the linter auto-correct the files if it can? Or does it simply error out with the errors and violations when you run it?

rustaceanrob:
Yeah, it should error out. I checked out master and ran it to get the Kyoto errors, then switched to the feature branch and they passed. I think the other tests were just written well

thunderbiscuit:
Oh awesome! I'm surprised they pass because I don't know the Python standards enough and even when I run Kotlin linters on my code it tends to fix a bunch of stuff so.. that's good. Ready to go then.

thunderbiscuit (approved):
ACK af17607f7c62e154a196146faa782196083d6aea.

Issue #636, opened 2024-12-10
Add apply_unconfirmed_txs to Wallet


andreasgriffin:
**Describe the enhancement**  

Add [apply_unconfirmed_txs](https://docs.rs/bdk/1.0.0-alpha.11/bdk/wallet/struct.Wallet.html#method.apply_unconfirmed_txs) to the bdk-ffi Wallet

**Use case**  

- It is useful to include local transactions to the wallet, to build new transactions ontop without actually broadcasting them yet.   
- It is also good to add transactions you know are in (most) mempools, but the client (e.g. kyoto) doesn't know about them

thunderbiscuit:
This relates to a broader discussion, so I'll link some of those here:

1. bitcoindevkit/bdk#1748
2. bitcoindevkit/bdk#1666
3. bitcoindevkit/bdk#849

reez:
If possible if we can add to 1.2 since we will be releasing Kyoto in 1.2?

PR #704, opened 2025-03-20
feat: apply_unconfirmed_txs on wallet

reez:
### Description

Add [apply_unconfirmed_txs](https://docs.rs/bdk_wallet/latest/bdk_wallet/struct.Wallet.html#method.apply_unconfirmed_txs) to Wallet, resolving #636  potentially. 

### Notes to the reviewers

Created a `UnconfirmedTx` type, similar to how we needed to create a `SentAndReceivedValues` type instead of returning a tuple. Totally open to naming it something different, but just needed a type instead of a tuple.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -737,7 +737,13 @@ interface Wallet {
 
   [Throws=SqliteError]
   boolean persist(Connection connection);
+
+  /// Apply relevant unconfirmed transactions to the wallet.
+  /// Transactions that are not relevant are filtered out.
+  void apply_unconfirmed_txs(sequence<UnconfirmedTx> unconfirmed_txs);
 };
+
+typedef dictionary UnconfirmedTx;
 
 interface Update {};
 
--- a/bdk-ffi/src/types.rs
+++ b/bdk-ffi/src/types.rs
@@ -592,3 +592,11 @@ impl From<BdkTx> for Tx {
         }
     }
 }
+
+/// This type replaces the Rust tuple `(tx, last_seen)` used in the Wallet::apply_unconfirmed_txs` method,
+/// where `last_seen` is the timestamp of when the transaction `tx` was last seen in the mempool.
+#[derive(uniffi::Record)]
+pub struct UnconfirmedTx {
+    pub tx: Arc<Transaction>,
+    pub last_seen: u64,
+}
--- a/bdk-ffi/src/wallet.rs
+++ b/bdk-ffi/src/wallet.rs
@@ -7,7 +7,7 @@ use crate::error::{
 use crate::store::Connection;
 use crate::types::{
     AddressInfo, Balance, CanonicalTx, FullScanRequestBuilder, KeychainAndIndex, LocalOutput,
-    Policy, SentAndReceivedValues, SignOptions, SyncRequestBuilder, Update,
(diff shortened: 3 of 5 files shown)


bdk-ffi/src/bdk.udl
```diff
@@ -737,6 +737,15 @@ interface Wallet {
...
+
+  /// Apply relevant unconfirmed transactions to the wallet.
+  /// Transactions that are not relevant are filtered out.
+  void apply_unconfirmed_txs(sequence<UnconfirmedTx> unconfirmed_txs);
+};
+
+dictionary UnconfirmedTx {
+    Transaction tx;
+    u64 last_seen;
 };
```
thunderbiscuit, line 749: Applying our idea of using macros when possible on changes and new things, this block here can be replaced with

```txt
typedef dictionary UnconfirmedTx;
```

and the Rust code can simply add
```rust
#[derive(uniffi::Record)]
pub struct UnconfirmedTx {
    pub tx: Arc<Transaction>,
    pub last_seen: u64,
}
```

To leverage the procedural macros.

bdk-ffi/src/types.rs
```diff
@@ -592,3 +592,7 @@ impl From<BdkTx> for Tx {
         }
     }
 }
+pub struct UnconfirmedTx {
```
thunderbiscuit, line 595: If we go with the proc macros proposed above, a little bit of API docs to describe this type might be nice.

bdk-ffi/src/wallet.rs
```diff
@@ -130,6 +130,14 @@ impl Wallet {
             .map_err(CannotConnectError::from)
     }
 
+    pub fn apply_unconfirmed_txs(&self, unconfirmed_txs: Vec<UnconfirmedTx>) {
+        self.get_wallet().apply_unconfirmed_txs(
+            unconfirmed_txs
+                .into_iter()
+                .map(|utx| (Arc::new((&*utx.tx).into()), utx.last_seen)),
```
thunderbiscuit, line 137: I find this line hard to parse, but I understand it's required. You need to dereference the transaction in the Arc before taking a reference to it. I think in this case using 

```rust
.map(|utx| (Arc::new(utx.tx.as_ref().into()), utx.last_seen)),
```

would work as well and might be easier to read? Just a nit, pick whatever line you like best.

thunderbiscuit (commented):
A few ideas.

bdk-ffi/src/wallet.rs
reez: Thank you sir, much much easier to read, so thanks for calling this out. Updated

reez:
> We can merge this one next as soon as the tests pass, and I'll rebase my 705 and we can merge that after.

cool catch you in a few hours when tests pass 😆

bdk-ffi/src/types.rs
thunderbiscuit: Maybe something like this?

```rust
This type replaces the Rust tuple `(tx, last_seen)` used in the Wallet::apply_unconfirmed_txs` method, where `last_seen` is the timestamp of when the transaction `tx` was last seen in the mempool.
```

reez: perfect, updated

thunderbiscuit (approved):
ACK 4db7af0c16642357710304048355ae8c46a5de59.

PR #708, opened 2025-03-24
Migrate Amount to procedural macros

ItoroD:
### Description

This PR address issue #706 and migrates `Amount` to procedural macros

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:

* [x] I've added tests for the new feature

#### Bugfixes:

* [x] I'm linking the issue being fixed by this PR

PR #691, opened 2025-03-10
Migrate TxBuilder type to procedural macros

thunderbiscuit:
Trying out the proc macros on the TxBuilder.

A few things I want to keep in mind and look out for:
- Are the docs migrating correctly?
- Errors

PR #710, opened 2025-03-25
Add API docs for `Amount` type

thunderbiscuit:
Sorry I was too quick to merge #708! My bad. 

This fixes that.

See https://docs.rs/bitcoin/latest/bitcoin/struct.Amount.html for details.

### DIFF
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -93,26 +93,35 @@ impl FeeRate {
 impl_from_core_type!(BdkFeeRate, FeeRate);
 impl_into_core_type!(FeeRate, BdkFeeRate);
 
+/// The Amount type can be used to express Bitcoin amounts that support arithmetic and conversion
+/// to various denominations. The operations that Amount implements will panic when overflow or
+/// underflow occurs. Also note that since the internal representation of amounts is unsigned,
+/// subtracting below zero is considered an underflow and will cause a panic.
 #[derive(Debug, Clone, PartialEq, Eq, uniffi::Object)]
 pub struct Amount(pub BdkAmount);
 
 #[uniffi::export]
 impl Amount {
+    /// Create an Amount with satoshi precision and the given number of satoshis.
     #[uniffi::constructor]
     pub fn from_sat(satoshi: u64) -> Self {
         Amount(BdkAmount::from_sat(satoshi))
     }
 
+    /// Convert from a value expressing bitcoins to an Amount.
     #[uniffi::constructor]
     pub fn from_btc(btc: f64) -> Result<Self, ParseAmountError> {
         let bitcoin_amount = BdkAmount::from_btc(btc).map_err(ParseAmountError::from)?;
         Ok(Amount(bitcoin_amount))
     }
 
+    /// Get the number of satoshis in this Amount.
     pub fn to_sat(&self) -> u64 {
         self.0.to_sat()
     }
 
+    /// Express this Amount as a floating-point value in Bitcoin. Please be aware of the risk of
+    /// using floating-point numbers.
     pub fn to_btc(&self) -> f64 {
         self.0.to_btc()
     }


Itorouk (commented):
Please delete this comment. I commented with wrong git account

ItoroD (commented):
I have looked at the generated doc. Looks good. Only weird thing is the constructor doc. This was mentioned for TxBuilder [here ](https://github.com/bitcoindevkit/bdk-ffi/pull/691#discussion_r1990033577)promarcos as well hence I believe its a known thing. Besides this I could go ahead and approve.

ItoroD (approved):

thunderbiscuit:
Yeah I agree we should look at exactly where to put the docstrings. Let me confirm where is best and I'll push the changes if there are any.

thunderbiscuit:
I have played around with docstring placement.

I think the best place to have the general type definition is on the `struct` line directly, with smaller documentation on the constructors and methods. Note that docstrings added to the `impl` block will simply be ignored.

## Example

```rust
/// This is the definition used on the struct.
#[derive(Clone, Debug, uniffi::Object)]
pub struct Script(pub BdkScriptBuf);

/// This is the definition used on the impl block.
#[uniffi::export]
impl Script {
    /// This is the docstring for the constructor.
    #[uniffi::constructor]
    pub fn new(raw_output_script: Vec<u8>) -> Self {
        let script: BdkScriptBuf = raw_output_script.into();
        Script(script)
    }

    /// This is a docstring for a method.
    pub fn to_bytes(&self) -> Vec<u8> {
        self.0.to_bytes()
    }
}
```

Resulting API docs:

image

rustaceanrob:
I see. When using the actual JVM docs this is the best configuration. My autocomplete was showing the constructor docs, but that's more of an IDE quark. We should go with the style you have here

rustaceanrob:
ACK 807fd9f64dadc0e59d2c41e18b4bdbe7d94e6ace

ItoroD:
I am wondering if fromBTC() and fromSat() are constructors. As I can see from your doc addition below how it is displayed in the Constructor section in your screen shot. fromBTC() and fromSat() are not displayed that way. Nor do we use them like constructors in our test at least. Your Script example in your comment seems to be returning a reference of script directly which makes it more of a constructor. 

`fromBTC() and fromSat()` are just companion objects. Hence why in API doc you do not see it under constructor


>     /// This is the docstring for the constructor.
>     #[uniffi::constructor]
>     pub fn new(raw_output_script: Vec<u8>) -> Self {
>         let script: BdkScriptBuf = raw_output_script.into();
>         Script(script)
>     }

ItoroD:
if I am making any sense its my fault for adding the #[uniffi::constructor] on frombtc() and fromSat() as I added it in #708

thunderbiscuit:
@rustaceanrob @ItoroD good points. I also wanted to make sure I got correct behaviour on my IDE and decided to go one step further in my digging. Here is what I found after testing a bunch of stuff.

## Setup

I used the `Address` type because it's a full object (interface in uniffi), and has a struct block, and impl block, a default constructor, an alternate constructor, and standard methods.

```rust
/// These are the type-level API docs declared on the struct block.
#[derive(Debug, PartialEq, Eq, uniffi::Object)]
pub struct Address(BdkAddress<NetworkChecked>);

/// These are the API docs declared on the impl block.
#[uniffi::export]
impl Address {
    /// These are the API docs declared on the default constructor.
    #[uniffi::constructor]
    pub fn new(address: String, network: Network) -> Result<Self, AddressParseError> {
        let parsed_address = address.parse::<bdk_wallet::bitcoin::Address<NetworkUnchecked>>()?;
        let network_checked_address = parsed_address.require_network(network)?;

        Ok(Address(network_checked_address))
    }

    /// These are the API docs declared on an alternative constructor.
    #[uniffi::constructor]
    pub fn from_script(script: Arc<Script>, network: Network) -> Result<Self, FromScriptError> {
        let address = BdkAddress::from_script(&script.0.clone(), network)?;

        Ok(Address(address))
    }

    /// These are API docs declared on a method.
    pub fn script_pubkey(&self) -> Arc<Script> {
        Arc::new(Script(self.0.script_pubkey()))
    }
```

## Results

- If you have docstrings on the type as well as the constructor, the overall type documentation appears in the tooltip when hovering over or requesting API docs on the type declaration if using explicit type declaration like so:



- If you are instantiating a type using the default constructor however, the IDE will pull the docstring of the constructor _if it exists_, otherwise the overall type API docs will simply be displayed:



- Calling a method of course displays the method API docstring.



- API docs written on the impl block never show up anywhere (they're not pulled into the glue code).

thunderbiscuit:
@ItoroD about the constructors:

Yes uniffi doesn't allow for constructor overloading, and so only one standard constructor can be created per type. Rust usually uses the `Type::new()` method to declare its main constructor, and uniffi pulls that method (`new()`) as the main constructor by default. If it exist, this is the method that will translate into something like `Address(arg1, arg2)`. Other constructors will be named, and exist as static functions on the type (and are therefore put into the companion object), i.e. `Address.fromScript()` and `Amount.fromSat()`.

The companion object has always been a little quirk of Kotlin that I don't know was ever that clear, and I have even heard Andrey Breslav say it was one of the things he regrets introducing (complexity that didn't add much). They're really just static methods on the type.

Pardon me going off on this if you're super familiar with Kotlin idioms I just wasn't sure.

ItoroD:
> @ItoroD about the constructors:
> 
> Yes uniffi doesn't allow for constructor overloading, and so only one standard constructor can be created per type. Rust usually uses the `Type::new()` method to declare its main constructor, and uniffi pulls that method (`new()`) as the main constructor by default. If it exist, this is the method that will translate into something like `Address(arg1, arg2)`. Other constructors will be named, and exist as static functions on the type (and are therefore put into the companion object), i.e. `Address.fromScript()` and `Amount.fromSat()`.
> 
> The companion object has always been a little quirk of Kotlin that I don't know was ever that clear, and I have even heard Andrey Breslav say it was one of the things he regrets introducing (complexity that didn't add much). They're really just static methods on the type.
> 
> Pardon me going off on this if you're super familiar with Kotlin idioms I just wasn't sure.

That makes sense. 👍

Issue #662, opened 2025-02-04
Better clients


thunderbiscuit:
Our Electrum and Esplora clients are not offering all that's available to standard Electrum and Esplora clients. Many of the available methods are not useful at the bdk-ffi layer (things like subscribing to specific script updates), but some could be (like requesting a particular block header, or requesting a random transaction given its Txid).

After looking at the APIs for [Electrum](https://docs.rs/electrum-client/0.22.0/electrum_client/trait.ElectrumApi.html) and [Esplora](https://docs.rs/esplora-client/0.11.0/esplora_client/blocking/struct.BlockingClient.html), here are a few I think could be useful for folks interacting with these clients.

### Electrum
- `ElectrumApi::block_header()`
- `ElectrumApi::relay_fee()`
- `ElectrumApi::transaction_get_raw()` or `ElectrumApi::transaction_get()`
- `ElectrumApi::ping()`

### Esplora
- `BlockingClient::get_tx()`
- `BlockingClient::get_tx_status()`
- `BlockingClient::get_tx_info()`
- `BlockingClient::get_block_by_hash()`

reez:
**Electrum**

`ElectrumApi::block_headers_subscribe()`

ConorOkus:
https://github.com/bitcoindevkit/bdk/issues/603

reez:
> 
> ### Esplora
> 
> * `BlockingClient::get_tx()`

We've got `get_tx()` https://github.com/bitcoindevkit/bdk-ffi/blob/master/bdk-ffi/src/esplora.rs#L87 so I think we can cross that one off

andreasgriffin:
Please also include 
- proxy in electrum: https://docs.rs/electrum-client/0.23.0/electrum_client/struct.ConfigBuilder.html
- proxy in esplora: https://docs.rs/esplora-client/0.11.0/esplora_client/struct.Builder.html#method.proxy

reez:
> Please also include
> 
> * proxy in electrum: https://docs.rs/electrum-client/0.23.0/electrum_client/struct.ConfigBuilder.html
> * proxy in esplora: https://docs.rs/esplora-client/0.11.0/esplora_client/struct.Builder.html#method.proxy

This part is resolved by #711 now

PR #711, opened 2025-03-26
Add proxy for Esplora & Electrum

reez:
### Description

Add proxy to [Esplora](https://docs.rs/esplora-client/0.11.0/esplora_client/struct.Builder.html#method.proxy) and [Electrum](https://docs.rs/electrum-client/0.23.0/electrum_client/struct.ConfigBuilder.html)

Resolves comment https://github.com/bitcoindevkit/bdk-ffi/issues/662#issuecomment-2692053222

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -921,7 +921,8 @@ interface Descriptor {
 /// cache to avoid re-fetching already downloaded transactions.
 interface EsploraClient {
   /// Creates a new bdk client from a esplora_client::BlockingClient
-  constructor(string url);
+  /// Optional: Set the proxy of the builder
+  constructor(string url, optional string? proxy = null);
 
   /// Scan keychain scripts for transactions against Esplora, returning an update that can be
   /// applied to the receiving structures.
@@ -979,8 +980,9 @@ interface EsploraClient {
 /// cache to avoid re-fetching already downloaded transactions.
 interface ElectrumClient {
   /// Creates a new bdk client from a electrum_client::ElectrumApi
+  /// Optional: Set the proxy of the builder
   [Throws=ElectrumError]
-  constructor(string url);
+  constructor(string url, optional string? socks5 = null);
 
   /// Full scan the keychain scripts specified with the blockchain (via an Electrum client) and
   /// returns updates for bdk_chain data structures.
--- a/bdk-ffi/src/electrum.rs
+++ b/bdk-ffi/src/electrum.rs
@@ -25,9 +25,15 @@ use std::sync::Arc;
 pub struct ElectrumClient(BdkBdkElectrumClient<bdk_electrum::electrum_client::Client>);
 
 impl ElectrumClient {
-    pub fn new(url: String) -> Result<Self, ElectrumError> {
-        let inner_client: bdk_electrum::electrum_client::Client =
-            bdk_electrum::electrum_client::Client::new(url.as_str())?;
+    pub fn new(url: String, socks5: Option<String>) -> Result<Self, ElectrumError> {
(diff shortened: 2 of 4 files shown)


thunderbiscuit:
One comment I have is that this works fairly well with the addition of the new constructor, but if we wanted to expose more of the builder options we'd be stuck with a crazy amount of constructors (one for each combination of arguments the users might need):

- new_with_proxy
- new_with_proxy_and_timeout
- new_with_timeout_and_retry
- new_with_retry_and_validate_domain
- new_with_validate_domain_and_proxy

ect.

An alternative approach would be to make use of default arguments in the main constructor. The new parameters can then be added without creating breaking changes.

reez:
> One comment I have is that this works fairly well with the addition of the new constructor, but if we wanted to expose more of the builder options we'd be stuck with a crazy amount of constructors (one for each combination of arguments the users might need):
> 
> * new_with_proxy
> * new_with_proxy_and_timeout
> * new_with_timeout_and_retry
> * new_with_retry_and_validate_domain
> * new_with_validate_domain_and_proxy
> 
> ect.
> 
> An alternative approach would be to make use of default arguments in the main constructor. The new parameters can then be added without creating breaking changes.

Absolutely! Updated

thunderbiscuit:
I don't personally like the naming much but I see that the electrum client calls its method on the builder `socks5` (whereas the Esplora calls it the more intuitive name `proxy`).

I'm sort of open to keeping the name proxy here but it is a slight rename of the true Rust method.

thunderbiscuit:
Ok scratch my previous comment. After asking chatGPT I get this interesting answer:

### ChatGPT Question and Answer

_The rust electrum-client library has a builder with a socks5 method. Is this socks5 method always used for proxies? Is it used for anything else?_

Great question.

In the context of the electrum-client Rust library, the .socks5() method is specifically and only used to configure a SOCKS5 proxy. It’s not used for anything else — not for authentication, TLS settings, or other types of proxy protocols like HTTP proxies.

TL;DR

✅ Used for:
	•	Connecting to an Electrum server via a SOCKS5 proxy

❌ Not used for:
	•	HTTP proxies
	•	Authentication (username/password for the SOCKS5 proxy — not supported out of the box)
	•	TLS/SSL settings
	•	General connection options
	
## My thoughts

I think that in that way since proxy5 seems to be a _subset_ of what a proxy means, it's probably more accurate to keep the name socks5 (even though I didn't really know what it meant the first time I saw it).

reez:
@thunderbiscuit yeah I thought the naming was tricky so I went for proxy on both, but totally open to renaming. To make sure I'm on the same page, are you thinking:

`constructor(string url, optional string? socks5 = null);`

instead of

`constructor(string url, optional string? proxy = null);`

or something different on the Electrum specifically?

thunderbiscuit:
Yeah that's what I had in mind.

```
// Electrum
constructor(string url, optional string? socks5 = null);

// Esplora
constructor(string url, optional string? proxy = null);
```

The argument on the `socks5` method on the Electrum builder is actually `socks5_config`, but in this case because we are transforming the builder pattern into a constructor, I think keeping the name of the method is clearer.

reez:
> Yeah that's what I had in mind.
> 
> ```
> // Electrum
> constructor(string url, optional string? socks5 = null);
> 
> // Esplora
> constructor(string url, optional string? proxy = null);
> ```
> 
> The argument on the `socks5` method on the Electrum builder is actually `socks5_config`, but in this case because we are transforming the builder pattern into a constructor, I think keeping the name of the method is clearer.

Updated 👍

thunderbiscuit (approved):
ACK a3dce7a611a5277aad18df36fae4b334a1845239.

reez:
rebased, will merge after tests pass

PR #712, opened 2025-03-26
Fix: export BumpFeeTxBuilder to uniffi

thunderbiscuit:
This bug was introduced in #691. The `BumpFeeTxBuilder` was not exported through proc macros and was therefore not available in the bindings.

Note that this would not have happened when using the UDL, so it's a good thing to keep in mind! Everything compiled fine, but the type created in the glue code file was actually empty. Another good reason to keep a healthy test suite, and maybe adding the BumpFeeTxBuilder to one of the live the tests.


reez (approved):
ACK 7b5f925cd228f19de0c18d9d4d72a70a3bd4165f

PR #713, opened 2025-03-26
Fix visibility modifiers

thunderbiscuit:
This PR fixes our incorrect use of visibility modifiers on some of our types. Note that when exporting to uniffi, visibility modifiers were not respected anyway? Not sure how that works or if it's my understanding of Rust that is lacking here, but a private or crate-private struct should not be available to outside users, yet definitely was.

In any case this PR makes public the types and methods we want to expose to users of the bindings explicit.

### Changelog notice

No changelog.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/tx_builder.rs
+++ b/bdk-ffi/src/tx_builder.rs
@@ -22,30 +22,30 @@ use std::sync::Arc;
 /// calling `finish` to consume the builder and generate the transaction.
 #[derive(Clone, uniffi::Object)]
 pub struct TxBuilder {
-    pub(crate) add_global_xpubs: bool,
-    pub(crate) recipients: Vec<(BdkScriptBuf, BdkAmount)>,
-    pub(crate) utxos: Vec<BdkOutPoint>,
-    pub(crate) unspendable: Vec<BdkOutPoint>,
-    pub(crate) internal_policy_path: Option<BTreeMap<String, Vec<usize>>>,
-    pub(crate) external_policy_path: Option<BTreeMap<String, Vec<usize>>>,
-    pub(crate) change_policy: ChangeSpendPolicy,
-    pub(crate) manually_selected_only: bool,
-    pub(crate) fee_rate: Option<FeeRate>,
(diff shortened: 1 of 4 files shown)


bdk-ffi/src/tx_builder.rs
```diff
@@ -289,13 +289,13 @@ impl TxBuilder {
     /// This will be used to:
     ///
     /// 1. Set the `nLockTime` for preventing fee sniping. Note: This will be ignored if you manually specify a
-    /// `nlocktime` using `TxBuilder::nlocktime`.
+    ///    `nlocktime` using `TxBuilder::nlocktime`.
```
reez, line 292: Is the extra spacing here and in a couple other spots in the diff for any specific reason?

thunderbiscuit: Yeah those were warnings from cargo's linter (the `fmt` tool). I can't find the commit right now but do you remember when we had to turn off the "fail" on the fmt CI workflow when warnings were issued, and only fail on fmt's failures?

This had the downside of now letting us merge stuff even if there is a warning by cargo's linter. Not a big deal, but in this case I just thought I'd deal with the warning. The commit that did that is 78f38cff62fe45d8be3e5eae5281e668ca13a0ec.

reez (approved):
ACK 03c753529831772f039f05bd0eed25fb4cc621b0

Thanks for this PR, for one I always love this sorts of changes that just tighten up the codebase in some really nice way.

It also helped me discover a bit more about "when exporting to uniffi, visibility modifiers were not respected anyway", so I think I understand now that while `pub(crate)` was technically correct from a Rust perspective, it could be misleading because uniffi was exposing these functions regardless. 

So this PR makes the code more honest/explicit about its actual visibility in the bindings 👍

thunderbiscuit:
Yeah love it too. It sort of bothered me at first that it appeared inconsistent but I put it out of my mind for a while. The macros just brought me more into the Rust side of things lately and I'm finding little snags like that.

PR #714, opened 2025-03-27
Build the release-smaller profile when using just command

thunderbiscuit:
The command currently builds both the `dev` and `release-smaller` profiles when calling `just build`, which is twice the work. This fix ensures we always attempt to build the release-smaller profile.


reez (approved):
ACK 039d1f02d5a56e569208edd49515bd9c83ae6696

PR #725, opened 2025-04-08
Bump ubuntu image in ci to 22.04

### LINKED ISSUE #724: Bump Ubuntu images used in CI
The Ubuntu 20.04 image is being sunset and will stop working on April 15.

https://github.com/actions/runner-images/issues/11101

Some of our tests have already started failing because of their "brownouts".

thunderbiscuit:
See #724.

Fixes #724.

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   security_audit:
     name: Security audit
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     defaults:
       run:
         working-directory: bdk-ffi
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   build-test:
     name: "Build and test"
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     defaults:
       run:
         working-directory: bdk-ffi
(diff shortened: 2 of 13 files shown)


thunderbiscuit:
Actually small correction: I think we also had issues with the `ring` crate not being able to compile on a newer version of the NDK, and that had us stuck on NDK 21 for like a year which we had to download + install on each build instead of just pulling the default one on Ubuntu. But yeah basically I don't like changes in the NDK haha.

thunderbiscuit:
> using latest in Python and Rust runs

Yeah in theory it would. I sort of like having it all build/run/test on the same image so it's cleaner to think through issues if there are any that are related to the CI images, but also they haven't caused any other issues than the one mentioned above really. So yeah I'm not feeling too strongly one way or the other. 

In this case my local ndk is 25.X so I'm looking at bumping that to the 27 version that's the LTS version [mentioned here](https://developer.android.com/ndk/downloads). I won't merge until I've figured this thing out...

thunderbiscuit:
I'm saying all of this but in fact _they still change what's in the images_ even the images don't change. So you can pin the version of Ubuntu all you want but I remember once the Python wheels failed for some obscure reason and it was that the build tools included in the 20.04 image had actually been updated... forget about reproducibility it's just a circus out there lol.

rustaceanrob:
Strange the python linter didn't catch that unknown import. Oh well.

Since the philosophy is to keep all versions the same in CI to rule out failures due to OS mismatch, I think this all makes sense.

ACK 749c62fba50912c9e65ac8c852f8326e71ab3ef2

rustaceanrob (approved):

thunderbiscuit:
For some reason the Swift live tests fail (https://github.com/bitcoindevkit/bdk-ffi/actions/runs/14357615687/job/40254293039#step:4:624) but also the logs show that they all pass?

I ran all live tests locally and if everything is green I'll merge.

PR #728, opened 2025-04-11
Prepare libraries for 1.2 release: update readmes

thunderbiscuit:
This PR updates the readme files. It's a step in our [release process](https://github.com/bitcoindevkit/bdk-ffi/issues/722) that just prevents documenation rot by doing a quick check on every release.

In this case most of the changes are related to formatting. I try to follow [Google's Markdown Style Guide](https://google.github.io/styleguide/docguide/style.html).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Native language bindings for BDK
+# Language bindings for BDK
 
 <p>
     <a href="https://github.com/bitcoindevkit/bdk-ffi/blob/master/LICENSE"><img alt="MIT or Apache-2.0 Licensed" src="https://img.shields.io/badge/license-MIT%2FApache--2.0-blue.svg"/></a>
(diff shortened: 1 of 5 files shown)


rustaceanrob:
ACK a421d16b28857dc2fd0c1a2702240ecee28d259c

Did you find an auto-formatter or was this just by eye? It would be nice for the markdown style to be format-on-save in the repo. Also what is up with with NDK version numbers? Surely they aren't on their 12,479,018 patch version?

rustaceanrob (approved):

thunderbiscuit:
No I don't have an auto-formatter but that'd be great. Right now it's just eyeballing the basics.

As for the Android NDK, yeah it's super weird. I think it goes way back. The "patch" part is actually the build number, an internal artifact of AOSP. I don't really know why they're still using this naming scheme!

Thanks for the review.

PR #731, opened 2025-04-22
Migrate `Connection` type to procedural macros

thunderbiscuit:
Migration of the `Connection` type.

Docs pulled directly from the rusqlite crate: https://docs.rs/rusqlite/0.31.0/rusqlite/struct.Connection.html#method.open

One thing to note is that with proc macros, all methods in an `impl` block will be exported if that block is marked with the `#uniffi::export` annotation, meaning that if we want to keep some methods private/out of the public API for the bindings (for example the `Connection::get_store()`), they simply need to be brought into a separate impl block.

### DIFF
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -805,13 +805,7 @@ dictionary Condition {
 // bdk_sqlite crate
 // ------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


reez (approved):
ACK 2fd7aa16c4bffd21e7e70982af87f45f8960202a

PR #732, opened 2025-04-22
refactor: migrate Wallet type to procedural macros

thunderbiscuit:
Migrate the Wallet and related errors.

Note that default arguments (here `sign_options`) require special annotations:

```rust
#[uniffi::method(default(sign_options = None))]
pub fn sign()
```

As more and more types in the UDL are declared with
```txt
typedef interface ExampleType;
```

The types that were only used in the UDL as part of that type (arguments on methods or return values of `ExampleType`) can be removed entirely from the UDL. So far these come up as unused types warnings when running clippy so they're easy to identify, but I wonder if I missed some.


reez:
> Note that default arguments (here sign_options) require special annotations:

I assume this is the rule for any optional?

reez:
> The types that were only used in the UDL as part of that type (arguments on methods or return values of ExampleType) can be removed entirely from the UDL. So far these come up as unused types warnings when running clippy so they're easy to identify, but I wonder if I missed some.

Thanks for noting this

PR #733, opened 2025-04-22
refactor: migrate errors thrown by wallet to procedural macros

thunderbiscuit:
This one is bigger but simpler. Errors migrated here were all only used by the `Wallet` type, and can therefore be removed completely from the UDL file.

The 3 PRs from this morning amount to a fair amount of changes. I'm [running all live tests on this branch just for an extra check](https://github.com/bitcoindevkit/bdk-ffi/actions/runs/14599890734).


reez (approved):
ACK 652edcf7295053b0b5fb3e7bbcec119a8027c87b

+8 -71 🥳

ItoroD:
> his one is bigger but simpler. Errors migrated here were all only used by the `Wallet` type, and can therefore be removed completely from the UDL file.

Then I guess `typedef interface Wallet` can be removed from the udl as well. Since its not being referenced there ?

PR #736, opened 2025-04-23
Migrate Psbt to pro-macros

ItoroD:
### Notes to the reviewers
- `json_serialize` and `finalize` do not have direct mapping doc in rust api docs. I wrote my own doc comments there.
- I noticed that when bringing in docs from rust side, we might want to change rust specific terminology to more understandable ones for the bindings. 

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -1,5 +1,6 @@
 use crate::error::{
-    AddressParseError, FeeRateError, FromScriptError, PsbtError, PsbtParseError, TransactionError,ExtractTxError
+    AddressParseError, ExtractTxError, FeeRateError, FromScriptError, PsbtError, PsbtParseError,
+    TransactionError,
 };
 use crate::error::{ParseAmountError, PsbtFinalizeError};
 use crate::{impl_from_core_type, impl_into_core_type};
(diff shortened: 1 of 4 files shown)


ItoroD:
Here is an example of what I mean by rust specific terminology in our jvm doc

image

bdk-ffi/src/bitcoin.rs
```diff
@@ -11,7 +12,6 @@ use bdk_wallet::bitcoin::blockdata::block::Header as BdkHeader;
 use bdk_wallet::bitcoin::consensus::encode::serialize;
 use bdk_wallet::bitcoin::consensus::Decodable;
 use bdk_wallet::bitcoin::io::Cursor;
-use bdk_wallet::bitcoin::psbt::ExtractTxError;
```
ItoroD, line 14: This cause issues when I was testing after migrating to pro-marcos in this pr. Was there a reason why we were not using `ExtractTxError` from the error crate like the rest of the errors? 

I updated this to how we use the other errors like PsbtError, PsbtParseError and the issue was resolved.

bdk-ffi/src/bitcoin.rs
```diff
@@ -322,25 +322,48 @@ impl From<&Transaction> for BdkTransaction {
...
+#[uniffi::export]
 impl Psbt {
+    /// Creates a new `Psbt` instance from a base64-encoded string.
+    #[uniffi::constructor]
     pub(crate) fn new(psbt_base64: String) -> Result<Self, PsbtParseError> {
         let psbt: BdkPsbt = BdkPsbt::from_str(&psbt_base64)?;
         Ok(Psbt(Mutex::new(psbt)))
     }
 
+    /// Serialize the PSBT into a writer.
```
thunderbiscuit, line 338: How do you feel about this line instead? It feels like it would match your doc from the constructor above.

_Serialize the PSBT into a base64-encoded string._

bdk-ffi/src/bitcoin.rs
```diff
@@ -350,13 +373,19 @@ impl Psbt {
...
+    ///
+    /// In accordance with BIP 174 this function is commutative i.e., `A.combine(B) == B.combine(A)`
     pub(crate) fn combine(&self, other: Arc<Psbt>) -> Result<Arc<Psbt>, PsbtError> {
         let mut original_psbt = self.0.lock().unwrap().clone();
         let other_psbt = other.0.lock().unwrap().clone();
         original_psbt.combine(other_psbt)?;
         Ok(Arc::new(Psbt(Mutex::new(original_psbt))))
     }
 
+    /// Finalizes the current PSBT (Partially Signed Bitcoin Transaction) and produces a result indicating
```
thunderbiscuit, line 386: I would remove the "(Partially Signed Bitcoin Transaction)" here and on the `json_serialize()` method docs, since we're in the type itself and so there is little chance of confusion.

thunderbiscuit (changes requested):
A few small tweaks but this is looking good!

thunderbiscuit:
Agreed that the docs are not always clear and appropriate. It's a bit up in the air what our policy with those are at the moment; in some cases you'd want the docs to be a little different between languages, which is not possible. In other cases the Rust docs clearly refer to things that don't exist in the ffi layer (in which case I have taken the liberty of removing those lines). 

We should talk about this over the next dev call. For now we've been trying to just copy/paste the Rust docs, but some cleanup might be required.

bdk-ffi/src/bitcoin.rs
ItoroD: ok. Great!

bdk-ffi/src/bitcoin.rs
ItoroD: Much better

bdk-ffi/src/bitcoin.rs
ItoroD: Yes. I will update

thunderbiscuit:
My bad! After looking into an odd cargo warning, I realize that the FinalizedPsbtResult should use `#[derive(uniffi::Record)]`

https://mozilla.github.io/uniffi-rs/latest/proc_macro/records.html

thunderbiscuit:
Yes exactly. You either get fields or methods (one of the quirks of uniffi). This is why we have methods on Transaction that are actually just fields in Rust, because we cannot have both, and in cases where Rust uses both we expose the fields through getters.

ItoroD:
Ok I will create pr for the fix now

ItoroD:
or would you rather do this in your pr? @thunderbiscuit

thunderbiscuit:
No feel free to open a PR for it!

PR #737, opened 2025-04-23
Migrate `ChangeSpendPolicy` type to procedural macros

thunderbiscuit:
This one is done as a [remote type](https://mozilla.github.io/uniffi-rs/latest/types/remote_ext_types.html#remote-types).

Let me know if you have any questions!


bdk-ffi/src/tx_builder.rs
```diff
@@ -18,6 +17,8 @@ use std::convert::{TryFrom, TryInto};
 use std::str::FromStr;
 use std::sync::Arc;
 
+type ChangeSpendPolicy = bdk_wallet::ChangeSpendPolicy;
```
reez, line 20: more out of curiosity (and not having as many reps as you on this) than me thinking its wrong or anything, what made you have to do this (function, aesthetic, etc)?

reez (approved):
ACK 8429a0ee51f21b94dbd87945e22330c4d6d2b94b

I'm assuming tests pass

PR #738, opened 2025-04-23
FinalizedPsbtResult : Change pro-macros to record

ItoroD:
The macro should have been #[derive(uniffi::Record)] instead of #[derive(uniffi::Object)] thanks for pointing this out @thunderbiscuit 

I see unused imports have been handled in #737. Thanks!

### DIFF
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -483,7 +483,7 @@ impl From<BdkPsbt> for Psbt {
     }
 }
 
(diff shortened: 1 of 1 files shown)


thunderbiscuit (approved):
ACK ccf2748f58f887ffd2734a23c2ee63d524bc6049.

ItoroD:
> Hey man! Not sure why but your commit here is not gpg signed, which means I can't merge it. Mind fixing that?

Strange. Will update

PR #742, opened 2025-04-25
Migrate descriptor types to proc macros

rustaceanrob:
### Description

Move `Descriptor`, `DescriptorPublicKey`, `DescriptorPrivateKey`, `DerivationPath`, and `Mnemonic` to proc macros. Adds documentation when appropriate, including links. Uses `uniffi::export(Display)` to derive `Display`, think that is the proper way to do it?

### Notes to the reviewers

There was a `TODO` questioning the use of `thread_rng` to create a mnemonic. Yes, `thread_rng` implements `CryptoRng`: https://docs.rs/rand/latest/rand/rngs/struct.ThreadRng.html#impl-CryptoRng-for-ThreadRng

### Changelog

Deprecated `as_string` for `DescriptorSecretKey`/`DescriptorPublicKey` in favor of `to_string` via `Display`

#### All Submissions:

### DIFF
--- a/bdk-ffi/src/keys.rs
+++ b/bdk-ffi/src/keys.rs
@@ -19,9 +19,15 @@ use std::ops::Deref;
 use std::str::FromStr;
 use std::sync::{Arc, Mutex};
 
+/// A mnemonic seed phrase to recover a BIP-32 wallet.
+#[derive(uniffi::Object)]
+#[uniffi::export(Display)]
 pub struct Mnemonic(BdkMnemonic);
 
+#[uniffi::export]
 impl Mnemonic {
+    /// Generate a mnemonic given a word count.
+    #[uniffi::constructor]
     pub fn new(word_count: WordCount) -> Self {
         // TODO 4: I DON'T KNOW IF THIS IS A DECENT WAY TO GENERATE ENTROPY PLEASE CONFIRM
         let mut rng = rand::thread_rng();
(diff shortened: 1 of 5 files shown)


rustaceanrob:
751134c2563314ddcf2f1cafe1a17ba1bdc2a91a is rebased

rustaceanrob:
2130c64abec19e961af83f33dc82dc0ab019e826 rebased once more

bdk-ffi/src/keys.rs
```diff
@@ -220,14 +253,17 @@ impl DescriptorPublicKey {
         }
     }
 
+    /// Convert the descriptor to a string.
     pub fn as_string(&self) -> String {
         self.0.to_string()
     }
```
thunderbiscuit, line 259: Same as above: this is a vestige of the times where we could not get the Display trait on types.

I think we should add the Display in this case so the `.toString()` method is implemented for `DescriptorPublicKey`.

bdk-ffi/src/keys.rs
```diff
@@ -158,21 +184,27 @@ impl DescriptorSecretKey {
         secret_bytes
     }
 
+    /// Convert this secret key to a string.
     pub fn as_string(&self) -> String {
         self.0.to_string()
     }
```
thunderbiscuit, line 190: This is a vestige of the times where we could not get the Display trait on types.

I think we should add the Display in this case so the `.toString()` method is implemented for `DescriptorSecretKey`.

thunderbiscuit (commented):
I'm trying to decide if these are part of the API we agreed not to break as part of the 1.0, because the renaming of the `keychain` argument to `keychainKind` is a breaking change.

The removal of the `as_string()` methods in favour of the Display trait which would give us the better `toString()` would also be, but in this case we can simply have both and not create a break.

rustaceanrob:
`as_string` removed for `DescriptorSecretKey` and `DescriptorPublicKey` in lieu of `Display`. 1c9574b610f823b73d49085dd7bfffc7e9145e7e is rebased

rustaceanrob:
58e06a91f369f705871d42575457506dcc6e2b70 rebased

thunderbiscuit (approved):
This one is bigger and so I'd like to get at least a second ACK (@reez or @ItoroD?)

ACK 58e06a91f369f705871d42575457506dcc6e2b70.

Because it has breaking changes, do you mind writing a small changelog block in your initial PR comment (something I steal on all PRs when I build the changelog).

rustaceanrob:
Go for it 👍

reez:
(reviewing right now)

reez:
> I'm trying to decide if these are part of the API we agreed not to break as part of the 1.0, because the renaming of the keychain argument to keychainKind is a breaking change.

PR code looks solid, but I'm more wondering about this commend from thunder, did you come to a consensus on if this was something we agreed to not break as a part of the 1.0?

rustaceanrob:
Since the next release will be a BDK wallet 2.0.0, I want to try to get arbitrary persistence available to FFI users. Would require breaking changes but we have a major release to do so

thunderbiscuit:
Yeah @reez we fell down the same trap. I just forgot that the next release was major. It's a good time to fix some of those little (breaking) quirks IMO.

reez:
@rustaceanrob @thunderbiscuit yup makes sense

reez (approved):
ACK 58e06a91f369f705871d42575457506dcc6e2b70

PR #747, opened 2025-04-27
feat: export xpub, tpub or vpub from Descriptor

rubensmachion:
### Description

Created a convenient method to extract xpub from descriptors

### Notes to the reviewers

• Added new dependency: regex
• Created a new function to_x_pub in the Descriptor class to expose the xpub
• Added a test to validate `Descriptor::to_x_pub()`

#### All Submissions:

* [X] I've signed all my commits
* [X] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [X] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:

* [X] I've added docs for the new feature

### DIFF
--- a/bdk-ffi/Cargo.lock
+++ b/bdk-ffi/Cargo.lock
@@ -30,6 +30,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "aho-corasick"
+version = "1.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
 name = "anstyle"
 version = "1.0.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


rubensmachion:
Hi @reez, just checking — do you find this relevant?

reez:
Regarding the current approach using regex: While it works for the common cases, our general philosophy for bdk-ffi is to leverage the underlying bdk and rust-bitcoin libraries as much as possible. These libraries already have pretty robust + well-tested descriptor parsing capabilities. Building a custom regex parser possibly duplicates that effort and might miss edge cases or become outdated if the descriptor standards evolve. This way we can rely on the core library's parsing logic which is more robust and maintainable in the long run.

If you want to look at rust-bitcoin+bdk to find the solution they provide for doing this I think that might a good place to start. Let me know if that makes sense, happy to help if ya need me, and thanks for taking a look at this!

rubensmachion:
Thanks so much for the review. As a beginner, I didn’t realize this was already available. I created this PR because I was looking for a way to export the xpub and display it in BDKSwiftSampleWallet. I’ll message you later to talk more about this.

PR #754, opened 2025-05-01
Introduce hash-like types

rustaceanrob:
Types like "Txid", "BlockHash", "DescriptorId" are all just 32 byte arrays that represent hashes with different meanings. Currently they are represented as strings at the FFI layer, but they are also meaningful arrays of bytes. Particularly if a user wants to implement persistence over the FFI layer, they would want to efficiently serialize these types.

Here I am introducing a new group of types that all implement display, allow serialization to bytes, and may be constructed from an array of bytes. I went with a "rule of 3s" here, and also introduced a macro to do these implementations because there was a lot of boilerplate involved.

Note that all of these are included in the wallet changeset, which is required to represent in-full for FFI-layer custom persistence: https://docs.rs/bdk_wallet/latest/bdk_wallet/struct.ChangeSet.html

edit: Added `TxMerkleNode`, which may be interesting in the future if merkle inclusion proofs come to the bindings layer at some point for Electrum users.

edit: Added `Wtxid`, which is how transactions are identified over the p2p network as of SegWit. This is useful for Kyoto users.

### Notes to the reviewers

- I understand macros may not be the most legible, but this case in particular saves a lot of repeated logic.
- "DescriptorId" from `bdk_chain` is mostly an alias for `sha256::Hash`

### Changelog notice

- Introduce object types for `Txid`, `BlockHash`, `DescriptorId`, `Wtxid`, `TxMerkleNode`

#### All Submissions:

### DIFF
--- a/bdk-ffi/src/macros.rs
+++ b/bdk-ffi/src/macros.rs
@@ -19,3 +19,32 @@ macro_rules! impl_into_core_type {
         }
     };
 }
+
+#[macro_export]
+macro_rules! impl_hash_like {
+    ($ffi_type:ident, $core_type:ident) => {
+        #[uniffi::export]
+        impl $ffi_type {
+            /// Construct a hash-like type from 32 bytes.
+            #[uniffi::constructor]
+            pub fn from_bytes(bytes: Vec<u8>) -> Result<Self, HashParseError> {
+                let hash_like: $core_type = deserialize(&bytes).map_err(|_| {
+                    let len = bytes.len() as u32;
+                    HashParseError::InvalidHash { len }
+                })?;
+                Ok(Self(hash_like))
+            }
+
+            /// Serialize this type into a 32 byte array.
+            pub fn serialize(&self) -> Vec<u8> {
+                serialize(&self.0)
+            }
+        }
+
+        impl std::fmt::Display for $ffi_type {
+            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                self.0.fmt(f)
+            }
+        }
+    };
+}
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -1,27 +1,33 @@
 use crate::error::{
-    AddressParseError, ExtractTxError, FeeRateError, FromScriptError, PsbtError, PsbtParseError,
-    TransactionError,
+    AddressParseError, ExtractTxError, FeeRateError, FromScriptError, HashParseError, PsbtError,
+    PsbtParseError, TransactionError,
 };
 use crate::error::{ParseAmountError, PsbtFinalizeError};
-use crate::{impl_from_core_type, impl_into_core_type};
+use crate::{impl_from_core_type, impl_hash_like, impl_into_core_type};
 
 use bdk_wallet::bitcoin::address::NetworkChecked;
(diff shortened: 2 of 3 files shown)


notmandatory:
Concept ACK, and proc macros aren't too hard to read, especially simple ones like you added.

rustaceanrob:
A lot of my work on persistence relies on this change. @ItoroD can I get a review on this so I can respond to any outstanding feedback?

ItoroD:
> A lot of my work on persistence relies on this change. @ItoroD can I get a review on this so I can respond to any outstanding feedback?

Sure! will take a look in a bit

bdk-ffi/src/macros.rs
```diff
@@ -19,3 +19,32 @@ macro_rules! impl_into_core_type {
...
+#[macro_export]
+macro_rules! impl_hash_like {
+    ($ffi_type:ident, $core_type:ident) => {
+        #[uniffi::export]
+        impl $ffi_type {
+            /// Construct a hash-like type from 32 bytes.
+            #[uniffi::constructor]
+            pub fn from_bytes(bytes: Vec<u8>) -> Result<Self, HashParseError> {
+                let hash_like: $core_type = deserialize(&bytes).map_err(|_| {
+                    let len = bytes.len() as u32;
```
ItoroD, line 32: I tried doing some serializations with this. One observation I have is, when we get a txid from a transaction, in kotlin `transaction.computeTxid()` we get a 64 char Hex. So I had to convert to a 32 bytes array before passing it in here. So I guess the idea is anyone looking to store will be storing them as bytes (32 bytes) That pretty standard.

 The conversion does work well!

ItoroD, line 32: I tried doing some serializations with this. One observation I have is, when we get a txid from a transaction, in kotlin `transaction.computeTxid()` we get a 64 char Hex. So I had to convert to a 32 bytes array before passing it in here. So I guess the idea is anyone looking to store will be storing them as bytes (32 bytes) That pretty standard.

 The conversion does work well!

ItoroD:
ACK [c533b0a](https://github.com/bitcoindevkit/bdk-ffi/pull/754/commits/c533b0a8a6c3f25e134205f44aba5d1e870134d0)

reez (approved):
ACK c533b0a8a6c3f25e134205f44aba5d1e870134d0

bdk-ffi/src/macros.rs
thunderbiscuit: Yeah in Kotlin this conversion can be done using the `hexToByteArray()` extension function. So you could do for example `"abcd0123".hexToByteArray()` to get your bytes.

rustaceanrob: Ideally this type is used across the library, so you would be able to `serialize` to bytes instead of going through the string implementation

thunderbiscuit:
Itoro's comment made me realize that a follow-up PR could use the `Txid` type across the codebase, including the `Transaction::compute_txid()` method.

I found a few more:
- `Transaction::compute_txid()`
- The `txid` field on the `OutPoint` type
- The `BumpFeeTxBuilder` constructor
- `EsploraClient::get_tx`
- `EsploraClient::get_tx_info`
- `EsploraClient::get_tx_status`
- `Wallet::get_tx`
- The `txid` field on the `Tx` type

thunderbiscuit (approved):
ACK 22b213c7f45c94d4fb0ce559cc42bacdf91650d4.

PR #755, opened 2025-05-06
Update `bdk_kyoto` to `0.9.0`

rustaceanrob:
Debug strings and info messages from Kyoto are on separate channels now. `TxSent` is renamed to `TxGossiped` and is only emitted if we are sure that the remote node has requested the transaction from us and we could successfully send it. The node accepts a Socks5 proxy to route traffic over Tor. I added a specific `Socks5Proxy` for this. We also return a `Update` directly instead of an optional. This makes it clear Kyoto has synced, even if there may be duplicate information. A handful of methods on the client were also converted to synchronous calls.

Internally the `bdk_kyoto` crate simply provides an extension trait over the Kyoto node builder, so users can take full advantage of the Kyoto features build still build for a specific `Wallet`. I will follow up with some additional configurations on the builder that are non-breaking.

### Changelog notice

- Add Socks5 proxy to Kyoto client options
- Separate debug strings and `Info` messages for `CbfClient`
- `CbfClient` returns an `Update` instead of optional

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/Cargo.lock
+++ b/bdk-ffi/Cargo.lock
@@ -166,9 +166,9 @@ dependencies = [
 
 [[package]]
 name = "bdk_kyoto"
-version = "0.8.0"
+version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8922b9a7b279e260bca897fa14fd6f735b529c6ac95d87a25d2b170799e103b0"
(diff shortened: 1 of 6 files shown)


reez (approved):
ACK 78a1134cd7d16b8fb931ff3ec311d036ef790bb0

thunderbiscuit:
This one needs a quick rebase and is ready to go.

rustaceanrob:
Rebased

PR #756, opened 2025-05-07
Express the wallet changeset over the FFI layer

### LINKED ISSUE #739: Continued discussion on encrypting persisted data
**Describe the enhancement**  

Per bitcoindevkit/bdk#1948, encrypting to disk is a desirable property for bindings users. In Rust, this is possible by using either `rusqlite:bundled-sqlcipher` or `rusqlite:bundled-sqlcipher-vendored-openssl`, which allows for a sqlite connection to be encrypted by using a `pragma_update` step. 

```rust
let conn = Connection::open(PATH).unwrap();
conn.pragma_update(None, "MY_KEY", "MY_SECRET").unwrap();
```

While the `bdk_wallet` does not ship with this by default, it's trivial to just add the additional "bundled-sqlcipher-vendored-openssl" feature to a `rusqlite` import, so I would consider this completed. 

**Use case**  

Apparently there is some history of viruses/trojans snooping around for `wallet.dat` files. Either way, @andreasgriffin has at least made a case for encrypting user data multiple times, and it would be nice to have feature parity with Rust in this regard.

**Additional context**

The `minreq` client currently ships with `rustls`, and I recall somehow that the `native-tls` crate was a problem for bindings? If it was the case that `minreq` shipped with `native-tls` and `rusqlite` shipped with bundled OpenSSL, they _should_ be able to share the OpenSSL implementation, but I will have to research both the binary size changes and if it is possible to link a shared `openssl-sys` dependency between `minreq` and `rusqlite`

rustaceanrob:
### Description

Builds on #754 to create a fully expressive and serializable `ChangeSet` type. The goal is for this to enable arbitrary persistence backends across the FFI layer. Ultimately that would close #739 and any further database requests that are outside the immediately supported options.

For context: The iterators in the BDK changesets are `BTree` based to be `no_std` compatible and have deterministic ordering of duplicate data across all targets. UniFFI exposes only two iterators across the boundary, `Vec` and `HashMap`, but I can try to lay out why I think using these is acceptable. 

The goal here is to convert between `bdk_wallet::ChangeSet` and FFI `ChangeSet`, and the easiest way to go about that is to just make each field convertible. For the `LocalChain::ChangeSet`, I iterate over the `BTreeMap`, which will return the key-value pairs in order. That means the vector will be in order over the FFI layer. Everything else coming from BDK _could_ be ordered by key, but is not required in my opinion to implement the persistence. For instance any mapping from `DescriptorId` to last revealed index is okay, as the `DescirptorId` do not necessarily have to be in any particular order. Same goes for `Txid` or `Outpoint` mapping to some value. While it is nice to be reproducible, these keys are suitable for a `HashMap`.

Can elaborate on the next call

### Notes to the reviewers

In Swift, and key of a `HashMap` must implement `Equitable` and `Hashable`. We can export `Eq` and `Hash` from Rust for _Objects_, but not for _Records_ from what I can tell. As a result, I made a kind-of-gross wrapper called `HashableOutPoint` that is an `Object` so it can be used as a key, but it is convertible to the underlying `Record` if required. Open to ideas on that one.

Also, I cannot for the life of me figure out the best way to go from a `Descriptor<DescriptorPublicKey>` into `DescriptorPublicKey` and vise versa without introducing an `unwrap`. Feels like we can improve that.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/bdk.udl
+++ b/bdk-ffi/src/bdk.udl
@@ -304,10 +304,6 @@ interface FullScanScriptInspector {
   void inspect(KeychainKind keychain, u32 index, Script script);
 };
 
-/// A changeset for [`Wallet`].
-[Remote]
-interface ChangeSet {};
-
 // ------------------------------------------------------------------------
 // bdk_wallet crate - wallet module
 // ------------------------------------------------------------------------
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -72,13 +72,22 @@ impl From<OutPoint> for BdkOutPoint {
     }
 }
 
-/// An [`OutPoint`] suitable as a key in a hash map.
+/// An [`OutPoint`] used as a key in a hash map.
+///
+/// Due to limitations in generating the foreign language bindings, we cannot use [`OutPoint`] as a
+/// key for hash maps.
 #[derive(Debug, PartialEq, Eq, std::hash::Hash, uniffi::Object)]
 #[uniffi::export(Debug, Eq, Hash)]
 pub struct HashableOutPoint(pub(crate) OutPoint);
 
 #[uniffi::export]
 impl HashableOutPoint {
+    /// Create a key for a key-value store from an [`OutPoint`]
+    #[uniffi::constructor]
+    pub fn new(outpoint: OutPoint) -> Self {
+        Self(outpoint)
+    }
+
     /// Get the internal [`OutPoint`]
     pub fn outpoint(&self) -> OutPoint {
         self.0.clone()
--- a/bdk-ffi/src/lib.rs
+++ b/bdk-ffi/src/lib.rs
@@ -43,7 +43,6 @@ use crate::types::SyncScriptInspector;
 
 use bdk_wallet::bitcoin::Network;
 use bdk_wallet::keys::bip39::WordCount;
-use bdk_wallet::ChangeSet;
 use bdk_wallet::KeychainKind;
 
 uniffi::include_scaffolding!("bdk");
--- a/bdk-ffi/src/types.rs
+++ b/bdk-ffi/src/types.rs
@@ -2,6 +2,7 @@ use crate::bitcoin::{
     Address, Amount, BlockHash, DescriptorId, HashableOutPoint, OutPoint, Script, Transaction,
     TxOut, Txid,
 };
+use crate::descriptor::Descriptor;
 use crate::error::{CreateTxError, RequestBuilderError};
 
 use bdk_core::bitcoin::absolute::LockTime as BdkLockTime;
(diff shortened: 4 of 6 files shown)


rustaceanrob:
Best left for after #764

rustaceanrob:
Rebased and ready for discussion next call

rustaceanrob:
To clarify, when trying to compile 

```rust
/// A reference to an unspent output by TXID and output index.
#[derive(Debug, Clone, Eq, PartialEq, std::hash::Hash, uniffi:: Record)]
#[uniffi::export(Eq, Hash)]
pub struct OutPoint {
    /// The transaction.
    pub txid: Arc<Txid>,
    /// The index of the output in the transaction.
    pub vout: u32,
}
```

UniFFI does not allow the `export(Eq, Hash)` which are required for keys in a hashmap. This is why `HashableOutPoint` exists.

rustaceanrob:
#771 demonstrates how this is used

rustaceanrob:
afc9fb832473eaed185a0ab12861f97af11a410f rebased

thunderbiscuit:
Question for @rustaceanrob: given the approach of #771 (which I have not reviewed yet), it's the job of the application layer to serialize the `ChangeSet` type into bytes to prepare for encryption into custom persistence right? i.e., we don't need a sort of `to_bytes()` or `serialize()` method on the ChangeSet type if I understand correctly your proposed approach?

rustaceanrob:
Yeah, I think it makes sense to spell this out explicitly for future reference:

Primitive objects that implement to/from bytes:
`Txid`, `Script`, `u32`, `u64`, `Transaction`, `BlockHash`, `DescriptorId`, `Network` (is an integer), `Descriptor` (UTF-8 string).

The records in the change sets that are made up of the primitives: 
`ChainChange`: `u32`, `Option<BlockHash>`
`BlockId`: `u32`, `BlockHash`
`ConfirmationBlockTime`: `BlockId`, `u64`
`Anchor`: `ConfirmationBlockTime`, `Txid`
`TxOut`: `Script`, `u64`
`OutPoint`: `Txid`, `u32`
`HashableOutPoint`: `OutPoint`
`TxGraphChangeSet`: `Transaction`, `Anchor`, `HashableOutPoint`, `Txid`, `TxOut`

By representing each of the smallest sub-components as serializable, the user can determine how they will store each one of these fields. For instance, maybe the `Descriptor` go to the secure element, the chain is held in a native K-V store, and the `TxGraph` is encrypted on disk. 

Some last minute clean-up:
- `HashableOutPoint` need a constructor
- I got mixed up with the descriptor types, there is no need to do any weird string serialization. All `unwrap` are removed. On the Rust side, FWIW I think `ExtendedDescriptor` is a stupid type alias

rustaceanrob:
They are [aware of this bug for Swift](https://github.com/mozilla/uniffi-rs/issues/2409), so hopefully we can remove `HashableOutPoint` in the future.

Looking for re-ACK on 30675dfec4773180d9ed99365e7538eb97942dc8

PR #761, opened 2025-05-12
Expose latest_checkpoint in Wallet implementation

ItoroD:
#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/src/types.rs
+++ b/bdk-ffi/src/types.rs
@@ -1,8 +1,9 @@
-use crate::bitcoin::{Address, Amount, OutPoint, Script, Transaction, TxOut};
+use crate::bitcoin::{Address, Amount, BlockHash, OutPoint, Script, Transaction, TxOut};
 use crate::error::{CreateTxError, RequestBuilderError};
 
 use bdk_core::bitcoin::absolute::LockTime as BdkLockTime;
 use bdk_core::spk_client::SyncItem;
+use bdk_core::BlockId as BdkBlockId;
 
 use bdk_wallet::bitcoin::Transaction as BdkTransaction;
 use bdk_wallet::chain::spk_client::FullScanRequest as BdkFullScanRequest;
@@ -66,7 +67,7 @@ impl From<BdkChainPosition<BdkConfirmationBlockTime>> for ChainPosition {
             } => {
                 let block_id = BlockId {
                     height: anchor.block_id.height,
-                    hash: anchor.block_id.hash.to_string(),
+                    hash: Arc::new(BlockHash(anchor.block_id.hash)),
                 };
                 ChainPosition::Confirmed {
                     confirmation_block_time: ConfirmationBlockTime {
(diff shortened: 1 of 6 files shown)


ItoroD:
Not sure if the println added to test `LiveTransactionTest.kt` is necessary. I figured it might be worth putting it there to test that the lastest_checkpoint addition works. Creating a whole new test for just that seemed overkill. Let me know it I should take it out.

bdk-ffi/src/types.rs
```diff
@@ -101,6 +102,15 @@ pub struct BlockId {
     pub hash: String,
 }
 
+impl From<BdkBlockId> for BlockId {
+    fn from(block_id: BdkBlockId) -> Self {
+        BlockId {
+            height: block_id.height,
+            hash: block_id.hash.to_string(),
```
thunderbiscuit, line 109: Here I think you can use the new [BlockHash](https://github.com/bitcoindevkit/bdk-ffi/blob/ed7064f34e27e150479ad48570ffa426fdfde68c/bdk-ffi/src/bitcoin.rs#L589) type if you strong-type the `hash` field in the `BlockId` type.

thunderbiscuit (changes requested):
Looks good! One little comment and I think it's good to go.

bdk-ffi/src/types.rs
ItoroD: When I attempt to update hash type to BlockHash, I get an error at compile. 

> `the trait uniffi::Lower<UniFfiTag> is not implemented for bitcoin::BlockHash`

Assuming `BlockHash` works, I will need to deal with the other places using hash of BlockId type I have identified just 1 [place ](https://github.com/bitcoindevkit/bdk-ffi/blob/ed7064f34e27e150479ad48570ffa426fdfde68c/bdk-ffi/src/types.rs#L69) so far. (FYI)

Itorouk: Ok. Issue was that I did not add the Arc around the BlockHash. My bad!

ItoroD: Ok it works. Issue was that I did not add the Arc around the BlockHash. My bad!

ItoroD:
I noticed some tests are expecting String for BlockHash (hence the failing test). I will update them.

bdk-ffi/src/types.rs
```diff
@@ -98,7 +99,16 @@ pub struct BlockId {
...
     /// The hash of the block.
-    pub hash: String,
+    pub hash: Arc<BlockHash>,
+}
+
+impl From<BdkBlockId> for BlockId {
+    fn from(block_id: BdkBlockId) -> Self {
+        BlockId {
+            height: block_id.height,
+            hash: Arc::from(BlockHash(block_id.hash)),
```
thunderbiscuit, line 109: I think here you should use `Arc::new()` instead of `Arc::from()`.

thunderbiscuit (changes requested):
Just one last little thing!

As for the tests, you can actually remove the blockId variables from the tests. They were intended to test the bindings generation for bitcoin-ffi and bdk-ffi (as per the comments), but now we only use bdk-ffi so you can just keep the network enum and the tests will do what we need them to do. Just to be clear this is what I mean:

```swift
let network = Network.testnet
```

```kotlin
val network = Network.TESTNET
```


```python
class TestBdk(unittest.TestCase):

    def test_some_enum(self):
        network = Network.TESTNET

if __name__=='__main__':
    unittest.main()
```

bdk-ffi/src/types.rs
ItoroD: Yes. I noticed. Thanks!

thunderbiscuit (approved):
ACK 34aa59f067aeb27d70b8a09467c36f5710844fd6.

PR #763, opened 2025-05-13
Export public types from `Script`

rustaceanrob:
### Description

I was developing on the latest commit in master and the script API was missing because it was not annotated with `uniffi::export`. I did a quick audit of the `impl` blocks but probably worth another pass by someone else to see if there are impl blocks missing exports

#### All Submissions:


thunderbiscuit (approved):
ACK ed7064f34e27e150479ad48570ffa426fdfde68c.

PR #764, opened 2025-05-14
Use `Txid`, `BlockHash`, `DescriptorId`, and `TxMerkleNode` where applicable

rustaceanrob:
Explicitly leaving out `BlockId` to avoid merge conflicts. The refactor here is moslty uneventful, however it removes a handful of `unwrap` when parsing transaction IDs.

### Changelog notice

- Use new hash types across public APIs

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/descriptor.rs
+++ b/bdk-ffi/src/descriptor.rs
@@ -1,3 +1,4 @@
+use crate::bitcoin::DescriptorId;
 use crate::error::DescriptorError;
 use crate::keys::DescriptorPublicKey;
 use crate::keys::DescriptorSecretKey;
@@ -5,6 +6,7 @@ use crate::keys::DescriptorSecretKey;
 use bdk_wallet::bitcoin::bip32::Fingerprint;
 use bdk_wallet::bitcoin::key::Secp256k1;
 use bdk_wallet::bitcoin::Network;
+use bdk_wallet::chain::DescriptorExt;
 use bdk_wallet::descriptor::{ExtendedDescriptor, IntoWalletDescriptor};
 use bdk_wallet::keys::DescriptorPublicKey as BdkDescriptorPublicKey;
 use bdk_wallet::keys::{DescriptorSecretKey as BdkDescriptorSecretKey, KeyMap};
@@ -294,6 +296,12 @@ impl Descriptor {
     /// Does this descriptor contain paths: https://github.com/bitcoin/bips/blob/master/bip-0389.mediawiki
     pub fn is_multipath(&self) -> bool {
         self.extended_descriptor.is_multipath()
+    }
+
+    /// A unique identifier for the descriptor.
+    pub fn descriptor_id(&self) -> Arc<DescriptorId> {
+        let d_id = self.extended_descriptor.descriptor_id();
+        Arc::new(DescriptorId(d_id.0))
     }
 
     /// Return descriptors for all valid paths.
--- a/bdk-ffi/src/electrum.rs
+++ b/bdk-ffi/src/electrum.rs
@@ -1,4 +1,4 @@
-use crate::bitcoin::{Header, Transaction};
+use crate::bitcoin::{BlockHash, Header, Transaction, Txid};
 use crate::error::ElectrumError;
 use crate::types::Update;
 use crate::types::{FullScanRequest, SyncRequest};
@@ -131,12 +131,12 @@ impl ElectrumClient {
     }
 
     /// Broadcasts a transaction to the network.
-    pub fn transaction_broadcast(&self, tx: &Transaction) -> Result<String, ElectrumError> {
+    pub fn transaction_broadcast(&self, tx: &Transaction) -> Result<Arc<Txid>, ElectrumError> {
         let bdk_transaction: BdkTransaction = tx.into();
         self.0
             .transaction_broadcast(&bdk_transaction)
             .map_err(ElectrumError::from)
-            .map(|txid| txid.to_string())
+            .map(|txid| Arc::new(Txid(txid)))
     }
 
     /// Returns the capabilities of the server.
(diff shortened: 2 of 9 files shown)


bdk-ffi/src/electrum.rs
```diff
@@ -177,7 +176,7 @@ pub struct ServerFeaturesRes {
     /// Server version reported.
     pub server_version: String,
     /// Hash of the genesis block.
-    pub genesis_hash: String,
+    pub genesis_hash: Arc<BlockHash>,
```
ItoroD, line 179: There is a  kotlin test that fails when I run locally  [here](https://github.com/bitcoindevkit/bdk-ffi/blob/34aa59f067aeb27d70b8a09467c36f5710844fd6/bdk-jvm/lib/src/test/kotlin/org/bitcoindevkit/LiveElectrumClientTest.kt#L44) Its seems the hash type is returning the reverse byte order (little-endian)

`Expected :000000000933ea01ad0ee984209779baaec3ced90fa3f408719526f8d77f4943`

`Actual   :43497fd7f826957108f4a30fd9cec3aeba79972084e90ead01ea330900000000`

This will happen for any other tests like these.

bdk-jvm/lib/src/test/kotlin/org/bitcoindevkit/LiveElectrumClientTest.kt
```diff
@@ -48,7 +48,7 @@ class LiveElectrumClientTest {
 
         assertEquals(
             expected = "000000000933ea01ad0ee984209779baaec3ced90fa3f408719526f8d77f4943",
-            actual = features.genesisHash
+            actual = features.genesisHash.toString()
```
ItoroD, line 51: Makes me wonder if the hashlike type implementation should the conversion to big endian. Else, either we always have to reverse it or it should just be known that you are getting the little endian bytes order.

bdk-ffi/src/electrum.rs
rustaceanrob: I will investigate this, as I am not sure if this is an Electrum protocol quark or the result of `deserialize` from `bitcoin`

ItoroD: Alright 👍

On my test from [here ](https://github.com/bitcoindevkit/bdk-ffi/pull/754#discussion_r2079970546) I did notice `deserializing` does return the reverse (little endian) which I think is normal for bitcoin. Same for `transaction.computeTxid()` if you serialize and deserialize with the `hashlike` type, you get reverse (little endian) of what `transaction.computeTxid()` actually gives

rustaceanrob: I would expect that to be the case, since it is the consensus encoding. That is fine when going to and from bytes. What I am skeptical of is the `Display` implementation of `bitcoin` using little endian. Any time I have seen a block hash printed to the console it is in the expected enidanness. It sounds like the `[u8; 32]` is not returned in the endian-ness expected by `deserialze`, so the `BlockHash` in `ServerFeatureRes` is just completely incorrect. I am almost certain `Display` will return the expected result at all other callsites and this was just incorrect for parsing `ServerFeatureRes`

ItoroD: Ah! I see. 👍

ItoroD: That works!

ItoroD:
ACK [d608592](https://github.com/bitcoindevkit/bdk-ffi/pull/764/commits/d6085926cf83f731bc5aa1def22b4f7543317a49)

bdk-ffi/src/electrum.rs
rustaceanrob: Thanks for that catch!

rustaceanrob:
Rebased

bdk-ffi/src/electrum.rs
rustaceanrob: The old state of the code was that. You should run `TestLiveElectrum.kt` to verify.

The `[u8; 32]` reported by Electrum is not in Bitcoin consensus format, so `deserialize` should produce a backwards blockhash. Parsing `[u8; 32]` as a string does it in the canonical format, not consensus format, so the string correct. That string is used to parse into a blockhash

bdk-ffi/src/descriptor.rs
```diff
@@ -296,6 +298,12 @@ impl Descriptor {
         self.extended_descriptor.is_multipath()
     }
 
+    /// A unique identifier for the descriptor.
+    pub fn descriptor_id(&self) -> Arc<DescriptorId> {
+        let d_id = self.extended_descriptor.descriptor_id();
+        Arc::new(DescriptorId(d_id.0))
```
thunderbiscuit, line 304: I think here we could return a straight `DescriptorId` instead of an Arc. Let me know if you had a reason to prefer the Arc.

bdk-ffi/src/electrum.rs
thunderbiscuit: Oh sorry I ran my tests locally but didn't recompile, so they were passing but it's because I wasn't on my fixed code haha. Yes I see them fail now. Thanks!

thunderbiscuit: Really odd. I'm trying to find out where this behavior is defined. The docs for Electrum are not clear: https://electrumx.readthedocs.io/en/latest/protocol-methods.html#server-features

thunderbiscuit (commented):
Looking good. I have one comment.

bdk-ffi/src/descriptor.rs
rustaceanrob: Everything will be reference counted in the target language anyway, so it doesn't seem like a big deal to go with the `Arc`. The only thing that shouldn't use `Arc` is `uniffi::Record` imo

thunderbiscuit (approved):
ACK.

PR #770, opened 2025-05-28
Add `lookahead` as optional argument

rustaceanrob:
### Description

Users may want to configure the initial `lookahead` value of a wallet, but we can leave this optional.

Using the default argument here, so this should actually be non-breaking. Compared to a builder this is ergonomic in the target languages IMO

### Changelog notice

- Add `lookahead` as optional wallet configuration

#### All Submissions:


bdk-ffi/src/wallet.rs
```diff
@@ -41,23 +41,23 @@ impl Wallet {
     /// Build a new Wallet.
     ///
     /// If you have previously created a wallet, use load instead.
-    #[uniffi::constructor]
+    #[uniffi::constructor(default(lookahead = None))]
```
thunderbiscuit, line 44: I wonder if using the default lookahead in bdk_wallet might make more sense for users looking at the default values. As is, you might think that unless one is set, your lookahead will be `null`, (even though that doesn't really make sense) but that's not true [(the default is 25)](https://docs.rs/bdk_chain/latest/bdk_chain/indexer/keychain_txout/constant.DEFAULT_LOOKAHEAD.html).

thunderbiscuit (commented):
Thanks for adding this! Just one question/comment.

bdk-ffi/src/wallet.rs
rustaceanrob: Answered that for myself. Macro requires a literal, so must be hard coded at 25.

thunderbiscuit (approved):
ACK 04756c061e7fc93c0fde492e8204015bc909454c.

PR #771, opened 2025-05-28
Abitrary persistence for wallet

rustaceanrob:
### Description

With the following approach we can accomplish two things at once:
1. Make it easy to add new Rust backends without sacrifing performance
2. Allow for arbitrary persistence accross the FFI

To accomplish this we can differentiate between a native backend and a foreign backend. The foreign backend interacts with the FFI `ChangeSet`, whereas the native backend can just use the Rust changeset directly. Whenever a new backend is introduced in Rust, a new enum variant may simply be added to the `PersistenceType`. To build a Sqlite backend, or a foreign backend, the user will use the `Persister` structure.

Abitrary persistence is allowed by implementing the `Persistence` trait. This was accomplished with no changes to https://github.com/bitcoindevkit/bdk-ffi/pull/756.

I hope 1. motivates this change thoroughly, as we expect BDK will add support for new backends in the future. I am also interested in the applications of 2., where a user might be able to do encrypted and/or cloud storage.

### Notes to the reviewers

We use an enum to allow for Rust backends to use Rust changesets directly. Otherwise we may convert them to FFI types. I've thought through many approaches, and I am convinced this is the most robust. 

### Changelog notice

- Add possibility for any type of persistence through a `Persister` type, that includes the usual sqlite approach

#### All Submissions:

### DIFF
--- a/bdk-ffi/src/store.rs
+++ b/bdk-ffi/src/store.rs
@@ -1,34 +1,88 @@
-use crate::error::SqliteError;
+use crate::error::PersistenceError;
+use crate::types::ChangeSet;
 
-use bdk_wallet::rusqlite::Connection as BdkConnection;
+use bdk_wallet::{rusqlite::Connection as BdkConnection, WalletPersister};
 
-use std::sync::Mutex;
-use std::sync::MutexGuard;
+use std::ops::DerefMut;
+use std::sync::{Arc, Mutex};
 
-/// A connection to a SQLite database.
+/// Definition of a wallet persistence implementation.
+#[uniffi::export(with_foreign)]
+pub trait Persistence: Send + Sync {
+    /// Initialize the total aggregate `ChangeSet` for the underlying wallet.
+    fn initialize(&self) -> Result<Arc<ChangeSet>, PersistenceError>;
+
+    /// Persist a `ChangeSet` to the total aggregate changeset of the wallet.
+    fn persist(&self, changeset: Arc<ChangeSet>) -> Result<(), PersistenceError>;
+}
+
+pub(crate) enum PersistenceType {
+    Custom(Arc<dyn Persistence>),
+    Sql(Mutex<BdkConnection>),
+}
+
+/// Wallet backend implementations.
 #[derive(uniffi::Object)]
-pub struct Connection(Mutex<BdkConnection>);
+pub struct Persister {
+    pub(crate) inner: Mutex<PersistenceType>,
+}
 
 #[uniffi::export]
-impl Connection {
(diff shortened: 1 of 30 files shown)


reez:
> I may implement a simple database in Swift or Python just to ensure the roundtrips are okay.

I'd def love to see this for Swift, and could help be another tester when you implement for anything you need

andreasgriffin:
Dummy implementation tested in python.  Will check the round trip one.

andreasgriffin:
`HashableOutPoint` is not serializable.

bdk-ffi/src/store.rs
```diff
@@ -1,34 +1,88 @@
...
+pub trait Persistence: Send + Sync {
+    /// Initialize the total aggregate `ChangeSet` for the underlying wallet.
+    fn initialize(&self) -> Result<Arc<ChangeSet>, PersistenceError>;
+
+    /// Persist a `ChangeSet` to the total aggregate changeset of the wallet.
+    fn persist(&self, changeset: Arc<ChangeSet>) -> Result<(), PersistenceError>;
+}
+
+pub(crate) enum PersistenceType {
+    Foreign(Arc<dyn Persistence>),
```
rustaceanrob: Note to use `Custom` terminology everywhere

rustaceanrob:
Good catch, I can add the offline prefix and fix that line tomorrow. Given how thorough the Python test is, I think we can keep the Swift and JVM with the default SQL options. Happy to accept other tests if they are added to my branch, but I think this should be finalized for review.

rustaceanrob:
Test is renamed with `test_offline` prefix and `bdk.Network | Node` is renamed to `Optional[bdk.Network]`

andreasgriffin:
there might be a sorting issue of dicts in older python versions, that will let the test fail.  I can investigate.

andreasgriffin:
> > sorting issue of dicts in older python versions
> 
> Yeah, this is expected as the changeset uses hash maps, which can be initialized differently per two different dictionaries. I don't think the final assertion is necessary. The balance is equal if and only if the `persist` and `initialize` steps are applied properly.

I think it is necessary, to check that persistence writing (not only initialization) works as expected.

andreasgriffin:
I could reproduce the failiure and fixed it in 2ac89a4432fdb85cd6e65ea009469788bc470c00 with more sorting in the serializer.

andreasgriffin:
I fixed the issue in 2ac89a4432fdb85cd6e65ea009469788bc470c00  by using more sorting in the serializer.

reez (approved):

PR #772, opened 2025-05-28
Update Kyoto to latest version

rustaceanrob:
Some minor changes to Kyoto:

We know internally if the node has stopped within the `update` method now, so we can return that error. There is also a new informational message. We also refer to all transactions with `Wtxid`, consistent with the p2p network.

### Changelog notice

- Update `bdk_kyoto` to `0.10.0`

#### All Submissions:

* [x] I ran `cargo fmt` and `cargo clippy` before committing.

### DIFF
--- a/bdk-ffi/Cargo.lock
+++ b/bdk-ffi/Cargo.lock
@@ -166,9 +166,9 @@ dependencies = [
 
 [[package]]
 name = "bdk_kyoto"
-version = "0.9.0"
+version = "0.10.0"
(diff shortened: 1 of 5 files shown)


reez:
Not sure if still in Draft mode with more to come, but taking a look at this its looking good to me

rustaceanrob:
After that field rename it should be all good to go. Gonna do that over coffee tomorrow!

rustaceanrob:
220c70b86d32e62b30fcbe2b2a10d47bf66a4584 rebased

PR #777, opened 2025-06-02
(draft) deps: bdk_wallet 2.0.0-beta

reez:
#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/types.rs
+++ b/bdk-ffi/src/types.rs
@@ -5,9 +5,9 @@ use crate::bitcoin::{
 use crate::descriptor::Descriptor;
 use crate::error::{CreateTxError, RequestBuilderError};
 
-use bdk_core::bitcoin::absolute::LockTime as BdkLockTime;
-use bdk_core::spk_client::SyncItem;
-use bdk_core::BlockId as BdkBlockId;
+use bdk_wallet::bitcoin::absolute::LockTime as BdkLockTime;
+use bdk_wallet::chain::spk_client::SyncItem;
+use bdk_wallet::chain::BlockId as BdkBlockId;
 
 use bdk_wallet::bitcoin::Transaction as BdkTransaction;
 use bdk_wallet::chain::spk_client::FullScanRequest as BdkFullScanRequest;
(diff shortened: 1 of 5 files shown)


reez:
Walking thru iOS https://github.com/bitcoindevkit/BDKSwiftExampleWallet/pull/297 checking out all the breaks updating from 1.2 -> 2.0.0-beta, and adding comments on the expected/unexpected items that broke (and why?)

bdk-ffi/src/types.rs
```diff
@@ -900,6 +901,8 @@ impl From<TxGraphChangeSet> for bdk_wallet::chain::tx_graph::ChangeSet<BdkConfir
             txouts,
             anchors,
             last_seen,
+            first_seen: Default::default(),
```
ItoroD, line 923: Just a comment here: Would we want to update `TxGraphChangeSet` to also have `first_seen` and `last_evicted` ? I see they are present for  `ChangeSet` in `bdk_wallet`. Or maybe that can wait for later.

reez: Good idea! I think we should do that. Let's do it in a separate PR? I'm thinking of this PR as a straight update of the dependency, and then doing follow up PR for something like exposing the new [TxDetails](https://github.com/bitcoindevkit/bdk-ffi/pull/778), so if you want to do a follow up PR

reez:
> LGTM!
> 
> Ran tests locally (kotlin) too.

Thanks for running those kotlin tests locally!

reez:
> > Walking thru iOS [bitcoindevkit/BDKSwiftExampleWallet#297](https://github.com/bitcoindevkit/BDKSwiftExampleWallet/pull/297) checking out all the breaks updating from 1.2 -> 2.0.0-beta, and adding comments on the expected/unexpected items that broke (and why?)
> 
> Thanks for this. (I am taking notes for kotlin sample wallet 😁)

No problem 👍

bdk-ffi/src/types.rs
ItoroD: Ok. Understood! 👍

ItoroD (approved):
ACK [1a5545b](https://github.com/bitcoindevkit/bdk-ffi/pull/777/commits/1a5545b8a9f138549004a4b29c9d28db9cb8a66c)

PR #782, opened 2025-06-10
TxGraphChangeSet: Include first_seen and last_evicted

ItoroD:
### Description
`tx_graph::ChangeSet` now has additional fields as at 2.0.0-beta. `first_seen` and `last_evicted`. This pr adds them to TxGraphChangeSet type.

#### All Submissions:

### DIFF
--- a/bdk-python/tests/test_offline_custom_persist.py
+++ b/bdk-python/tests/test_offline_custom_persist.py
@@ -20,7 +20,7 @@ change_descriptor: bdk.Descriptor = bdk.Descriptor(
 )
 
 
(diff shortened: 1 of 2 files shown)


bdk-python/tests/test_offline_custom_persist.py
```diff
@@ -93,13 +93,30 @@ def sort_key(t):
                 except AttributeError:
                     txid_hex = str(txid_obj)
                 last_seen_dict[txid_hex] = height
```
ItoroD: I noticed last_seen had variable `height` for its value pair. Shouldn't we name that variable timestamp? Since thats what the actual value pair type is in Changeset. Same for `first_seen` and `last_evicted`. Right now I just followed the conversion I saw for last_seen (`height`)

reez: Good call, timestamp is much clearer. Thanks for catching that

reez (approved):
ACK bdd12be9153d39d375a49b87895c2d86cc88b69e

Great work implementing the new fields, changing that naming in the python test is optional https://github.com/bitcoindevkit/bdk-ffi/pull/782#discussion_r2138716278, but I'll let you change that if you want because it's a good idea.

PR #784, opened 2025-06-11
feat: add from_string constructor for hash-like types

reez:
### Description

This PR adds a `from_string` constructor to all hash-like types to enable creating these types from hex strings. This restores the convenience of string-based construction.

### Notes to the reviewers

Tested locally built swift bindings with iOS app, and then was able to use `fromString`(swift) to address some lines of code mentioned in https://github.com/bitcoindevkit/BDKSwiftExampleWallet/pull/297/files

_Example:_ 

Pass a string into `BlockHash.fromString` to construct a `BlockHash` instead of pass bytes into `BlockHash.fromBytes`

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/macros.rs
+++ b/bdk-ffi/src/macros.rs
@@ -35,6 +35,14 @@ macro_rules! impl_hash_like {
                 Ok(Self(hash_like))
             }
 
+            /// Construct a hash-like type from a hex string.
+            #[uniffi::constructor]
+            pub fn from_string(hex: String) -> Result<Self, HashParseError> {
+                hex.parse::<$core_type>()
+                    .map(Self)
+                    .map_err(|_| HashParseError::InvalidHexString { hex })
+            }
+
             /// Serialize this type into a 32 byte array.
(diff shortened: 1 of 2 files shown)


bdk-ffi/src/macros.rs
```diff
@@ -35,6 +35,15 @@ macro_rules! impl_hash_like {
                 Ok(Self(hash_like))
             }
 
+            /// Construct a hash-like type from a hex string.
+            #[uniffi::constructor]
+            pub fn from_string(hex: String) -> Result<Self, HashParseError> {
+                use std::str::FromStr;
```
rustaceanrob: To avoid this import I believe you can do `hex.parse::<$core_type>().map_err(...`

reez: good call thanks!

ItoroD (approved):
ACK [2b0ee96](https://github.com/bitcoindevkit/bdk-ffi/commit/2b0ee964a4308d63077cd426cf4091d312a193c5)

PR #786, opened 2025-06-19
Add public_descriptor to wallet implementation

ItoroD:
### Description

I have been working with wallet and our descriptor type for my work on Psbt and multisig. I noticed that we cannot get the descriptor used to create the wallet directly from the `Wallet` type. Hence exposing [public_descriptor](https://docs.rs/bdk_wallet/latest/bdk_wallet/struct.Wallet.html#method.public_descriptor) will allow us get this directly. A good use case is if you want to create a watch only wallet you could just call `public_descriptor`.

At the moment I am returning the descriptor as a string instead of a [ExtendedDescriptor](https://docs.rs/bdk_wallet/latest/bdk_wallet/descriptor/type.ExtendedDescriptor.html). The downside to this is we loose all the functionality `ExtendedDescriptor` provides. We do not have a ExtendedDescriptor type which is just an alias for `Descriptor<DescriptorPublicKey>`. 

We could create and implement the type if necessary instead of using a string.

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/src/wallet.rs
+++ b/bdk-ffi/src/wallet.rs
@@ -433,6 +433,15 @@ impl Wallet {
             .tx_details(txid.0)
             .map(|details| details.into())
     }
+
+    /// Returns the descriptor used to create addresses for a particular `keychain`.
+    ///
+    /// It's the "public" version of the wallet's descriptor, meaning a new descriptor that has
+    /// the same structure but with the all secret keys replaced by their corresponding public key.
+    /// This can be used to build a watch-only version of a wallet.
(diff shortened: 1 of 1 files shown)


reez (approved):
ACK fdcf68eb6fe25924f6c4f7589f2656b8a6cae0fb

Thanks for adding this. Appreciate you adding the notes about extended descriptor, and I think the way you have it with string is the correct way for our codebase currently too.

thunderbiscuit:
lol I can't believe we didn't have this API exposed before! Thanks Itoro.

PR #789, opened 2025-06-26
refactor: use macros for `WordCount`

reez:
### Description

Migrate `WordCount` to use proc macro

### Notes to the reviewers

Unfortunately `lib.rs` required type alias thing, but this is the same pattern used for KeychainKind in `types.rs`. The type alias is necessary because we can't have both an import and an enum definition with the same name in the same scope.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/lib.rs
+++ b/bdk-ffi/src/lib.rs
@@ -35,6 +35,5 @@ use crate::types::LockTime;
 use crate::types::PkOrF;
 
 use bdk_wallet::bitcoin::Network;
-use bdk_wallet::keys::bip39::WordCount;
 
 uniffi::include_scaffolding!("bdk");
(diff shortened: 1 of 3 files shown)


bdk-ffi/src/lib.rs
```diff
@@ -42,7 +42,17 @@ use crate::types::SyncRequestBuilder;
 use crate::types::SyncScriptInspector;
 
 use bdk_wallet::bitcoin::Network;
-use bdk_wallet::keys::bip39::WordCount;
 use bdk_wallet::KeychainKind;
 
+type WordCount = bdk_wallet::keys::bip39::WordCount;
```
ItoroD: This is good. One small question is why is this put in lib.rs and not type.rs?

reez: good call, updated 👍

ItoroD (approved):
ACK [e449942](https://github.com/bitcoindevkit/bdk-ffi/commit/e449942d72a8281bce0f17d7f74b1d9aa65b0792)

reez:
had to rebase, will merge this first when ci passes

PR #798, opened 2025-07-04
feat: add `spend_utxo` to Psbt

ItoroD:
### Description
 Exposing `spend_utxo`  will be helpful when trying to validate signatures or sign transaction inputs. This will allows signers of a psbt to verify and confirm the amount being spent and ensure it is what they expect before they approve/sign the transaction.

### Notes to the reviewers
`spend_utxo` returns a [SignError](https://docs.rs/bitcoin/0.32.6/bitcoin/psbt/enum.SignError.html) as part of result in case of any errors. We do not have this error type so I skipped this. After testing we still get a fairly descriptive error message when an error occurs. At least for passing an out of bounds index.

### Changelog notice

- Check Psbt utxo at given index

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -560,6 +560,13 @@ impl Psbt {
         let psbt = self.0.lock().unwrap();
         serde_json::to_string(psbt.deref()).unwrap()
     }
+
+    /// Returns the spending utxo for this PSBT's input at `input_index`.
+    pub(crate) fn spend_utxo(&self, input_index: u64) -> String {
+        let psbt = self.0.lock().unwrap();
+        let utxo = psbt.spend_utxo(input_index as usize).unwrap();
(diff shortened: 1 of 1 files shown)


bdk-ffi/src/bitcoin.rs
```diff
@@ -675,6 +682,7 @@ impl_hash_like!(TxMerkleNode, BitcoinDoubleSha256Hash);
 
 #[cfg(test)]
 mod tests {
+    use super::Psbt;
```
reez: nit: are we able to use this `use crate::bitcoin::Psbt;` instead just to keep it consistent with the

```
use crate::bitcoin::Network;
use crate::bitcoin::Address;
```

below it

ItoroD: yes, makes sense

reez: thanks

PR #799, opened 2025-07-04
feat: implement display for `Transaction`

ItoroD:
### Description

I think we need to be able to see the details of a `Transaction`  when printed. In Kotlin, `Transaction` currently prints the memory address and not the display. @reez and I had noticed this and discussed briefly, see [here ](https://github.com/bitcoindevkit/bdk-ffi/pull/778#pullrequestreview-2920191449). I think we have manual implementations of display for struct like `Address`, so why not add it to `Transaction`  too. Its a bit of a hassle, developing and testing and not being able to see the printout for Tx. I understand that swift is ok having just `debug` trait and can printout with no issues. This will at least for now allow JVM users see the actual printouts.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -321,6 +321,7 @@ impl_into_core_type!(Address, BdkAddress);
 /// Bitcoin transaction.
 /// An authenticated movement of coins.
 #[derive(Debug, Clone, PartialEq, Eq, uniffi::Object)]
+#[uniffi::export(Eq, Display)]
 pub struct Transaction(BdkTransaction);
 
 #[uniffi::export]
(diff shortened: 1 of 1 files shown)


rustaceanrob:
I'm not sure it makes sense to derive the display trait for a debug string. The output won't be end-user friendly, so maybe this could be a "debug_string" method?

ItoroD:
> I'm not sure it makes sense to derive the display trait for a debug string. The output won't be end-user friendly, so maybe this could be a "debug_string" method?

Yes, I figured. Thanks for confirming. But without display implementation we can't view or print `Transaction` in Kotlin and this way is probably not the best way to go about it.

Not sure how debug_string will help.

reez (commented):
This looks good per our discussion at Bindings

PR #800, opened 2025-07-05
Add reading and writing PSBT files

rustaceanrob:
The Rust API makes it trivially simple to read and write PSBT to and from files. Bindings users may make use of this in air-gapped signing flows. The `PsbtError` already includes an `Io` variant, so we may make use of that error.

### Notes to the reviewers

Discovered while working on a desktop app.

### Changelog notice

- Read and write PSBT to and from files

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -32,6 +32,8 @@ use bdk_wallet::miniscript::psbt::PsbtExt;
 use bdk_wallet::serde_json;
 
 use std::fmt::Display;
+use std::fs::File;
+use std::io::{BufReader, BufWriter};
 use std::ops::Deref;
 use std::str::FromStr;
 use std::sync::{Arc, Mutex};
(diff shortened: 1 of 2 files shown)


bdk-ffi/src/bitcoin.rs
```diff
@@ -548,6 +559,15 @@ impl Psbt {
         }
     }
 
+    /// Write the `Psbt` to a file. Note that the file must not yet exist.
+    pub(crate) fn write_to_file(&self, path: String) -> Result<(), PsbtError> {
+        let file = File::create(path)?;
```
reez: Wondering your thoughts on this: doc comment says "Note that the file must not yet exist." but `File::create()` "will truncate it if it does" (exist) per [docs](https://doc.rust-lang.org/std/fs/struct.File.html#method.create). Should this use [`File::create_new()`](https://doc.rust-lang.org/std/fs/struct.File.html#method.create_new) instead to prevent overwrites? Or I could be misunderstanding workflow/behavior too.

bdk-ffi/src/error.rs
```diff
@@ -1438,6 +1438,22 @@ impl From<BdkPsbtParseError> for PsbtParseError {
...
+}
+
+impl From<bdk_wallet::bitcoin::io::Error> for PsbtError {
+    fn from(error: bdk_wallet::bitcoin::io::Error) -> Self {
+        PsbtError::Io {
+            error_message: error.to_string(),
+        }
+    }
+}
+
```
reez, line 1456: Thanks for adding these 👍

bdk-ffi/src/bitcoin.rs
```diff
@@ -548,6 +559,15 @@ impl Psbt {
...
+        let mut writer = BufWriter::new(file);
+        let psbt = self.0.lock().unwrap();
+        psbt.serialize_to_writer(&mut writer)?;
```
reez, line 585: https://docs.rs/bitcoin/latest/bitcoin/struct.Psbt.html#method.serialize_to_writer cool

bdk-ffi/src/bitcoin.rs
rustaceanrob: Good call, will swap to create new.

reez: cool, this looks good otherwise so I'll keep an eye on it and merge when swapped (and rebased if needed)

PR #801, opened 2025-07-05
Fix nightly lints

rustaceanrob:
I develop on recent nightly versions and they added some annoying lints. Fixing them here before a later CI fails.

#### All Submissions:


reez (approved):
ACK 650da8cc99b80ee3bd102791dd2a8937f8d1cb95

Thanks for doing this

PR #804, opened 2025-07-07
(draft) feat: add `from_string` to `Transaction`

reez:
_TODO: move away from UDL in PR once other PR's merged (had to use UDL until then)_

### Description

relates to 803 issue

Dont mind the UDL stuff, will rebase on recent changes that move that error out of UDL.

### Notes to the reviewers

Added a From trait for automatic error conversion to make things cleaner, consistent with what we usually try to do anyway.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -14,6 +14,7 @@ use bdk_wallet::bitcoin::consensus::encode::serialize;
 use bdk_wallet::bitcoin::consensus::Decodable;
 use bdk_wallet::bitcoin::hashes::sha256::Hash as BitcoinSha256Hash;
 use bdk_wallet::bitcoin::hashes::sha256d::Hash as BitcoinDoubleSha256Hash;
+use bdk_wallet::bitcoin::hex::FromHex;
 use bdk_wallet::bitcoin::io::Cursor;
 use bdk_wallet::bitcoin::secp256k1::Secp256k1;
 use bdk_wallet::bitcoin::Amount as BdkAmount;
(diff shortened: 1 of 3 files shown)


thunderbiscuit:
I'm not _against_ this per se and will certainly not oppose merging, but notice that this is not part of the Rust API (you cannot create a transaction using something like `Transaction::from_str`. The Rust API requires bytes, and the hex to byte transformation is done elsewhere.

In general I think if we add these sorts of utility APIs, we should have a way to mark them as our own (and not coming from our Rust dependencies). They should also get special attention in tests because of that fact (since they're not tested on the dependencies' side).

reez:
> I'm not _against_ this per se and will certainly not oppose merging, but notice that this is not part of the Rust API (you cannot create a transaction using something like `Transaction::from_str`. The Rust API requires bytes, and the hex to byte transformation is done elsewhere.
> 
> In general I think if we add these sorts of utility APIs, we should have a way to mark them as our own (and not coming from our Rust dependencies). They should also get special attention in tests because of that fact (since they're not tested on the dependencies' side).

Discussion definitely here but I'm interested in touching on this at Bindings call tomorrow too, I just took a quick scan of how many ffi convenience methods that dont mirror the Rust API we have in bdk-ffi (just so I could contextualize for myself)

bdk-ffi/src/bitcoin.rs
```diff
@@ -333,6 +334,13 @@ impl Transaction {
         Ok(Transaction(tx))
     }
 
+    /// Creates a new `Transaction` instance from a hexadecimal string representation.
+    #[uniffi::constructor]
+    pub fn from_string(tx_hex: String) -> Result<Self, TransactionError> {
```
reez, line 339: naming convention>?

bdk-ffi/src/bitcoin.rs
```diff
@@ -333,6 +334,13 @@ impl Transaction {
...
+        let tx_bytes = Vec::from_hex(&tx_hex)?;
```
reez, line 340: enforce consensus transaction?

reez: https://github.com/2140-dev/kyoto/blob/master/src/broadcaster.rs#L46

PR #812, opened 2025-07-17
feat: Add multisig example.

ItoroD:
### Description

Added a 2 of 3 multisig (Psbt) example.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-jvm/examples/src/main/kotlin/MultisigTransaction.kt
+++ b/bdk-jvm/examples/src/main/kotlin/MultisigTransaction.kt
@@ -1,0 +1,158 @@
+package org.bitcoindevkit
+
+import java.nio.file.Paths
+
+fun main() {
+    // Add your regtest URL here. Regtest environment must have esplora to run this.
+    // See here for regtest podman setup https://github.com/thunderbiscuit/podman-regtest-infinity-pro
+    val REGTEST_URL = "http://127.0.0.1:3002"
+
+    //Create 3 wallets
+    val aliceWallet = getNewWallet(ActiveWalletScriptType.P2WPKH, Network.REGTEST)
+    val bobWallet = getNewWallet(ActiveWalletScriptType.P2WPKH, Network.REGTEST)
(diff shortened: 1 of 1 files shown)


reez (approved):
ACK 594d174c5d9271c5d7c2451c57afeff4ee7f77b7

Ill merge after I finish 2.0 release

Good structure and comments! Can you open Issues for these ones that are only being done for bdk-jvm at the moment to make sure we track them so hopefully at some point getting them for Swift etc? They can even be labeled `good-first-issue` since someone just has to peek at what you've already done here to get them a good chunk of the way thru.

ItoroD:
Thanks @reez ! 

Alright.

thunderbiscuit (approved):
Tested ACK 594d174c5d9271c5d7c2451c57afeff4ee7f77b7.

One thing this example shows us is that the `Descriptor` type is probably not as powerful as it could be; having to parse the string manually instead of just querying the descriptor for the path seems like something probably available in the original Rust type (but the rust-bitcoin descriptor types are notorious for being complex so I know we also use wrappers on those in some cases, and BDK itself wraps them? Can't remember the specifics).

I'm wondering if you have a way to trigger those from the command line (in which case we could add just commands for them). Not a blocker. Just a nice to have. Also note that these will need to be migrated to the new [bdk-jvm](https://github.com/thunderbiscuit/bdk-jvm) repository at some point!

ItoroD:
> Tested ACK [594d174](https://github.com/bitcoindevkit/bdk-ffi/commit/594d174c5d9271c5d7c2451c57afeff4ee7f77b7).
> 
> One thing this example shows us is that the `Descriptor` type is probably not as powerful as it could be; having to parse the string manually instead of just querying the descriptor for the path seems like something probably available in the original Rust 

Yes, I was going to point that out too. The manual string thing felt like I was doing a Leetcode (easy) question. I check rust side to see what we have there.

PR #813, opened 2025-07-18
Export `Display` for `Script`

rustaceanrob:
The `rust-bitcoin` type has a nice feature where the human-readable OP-codes are printed when using the `Display` version of `ScriptBuf`. Some users might find this interesting, as this is a feature of the almighty Sparrow wallet.

To see for yourself:
```rust
let mut output = String::new();
addr.fmt_asm(&mut output).unwrap();
println!("{output}");
```
Which should have an output similar to this for a fully built out example:
```
OP_0 OP_PUSHBYTES_20 1cd30fcc6cc9793c08a1105893c6d1e584a57e48
```

### Changelog notice

- Added `Display` for `Script`

#### All Submissions:


thunderbiscuit (approved):
Yep I use this all the time! Thanks for adding it here.

ACK 18046acd17843a19ea06ba4dd6a396e842f6b75e

PR #814, opened 2025-07-28
Make a runtime environment for DNS for Kyoto

rustaceanrob:
The `lookup_host` function requires a `tokio` runtime to execute, but since it is outside of the `Node`, this function panics when called. Add a runtime, which is essentially an OS thread in this context, to complete the call.

### Notes to the reviewers

In the future I could potentially find a more elegant way to do this, but this call should be relatively short and only consume a single thread.

### Changelog notice

- Fix `lookup_host` on `CbfClient`

#### All Submissions:

#### New Features:

#### Bugfixes:


reez (approved):
ACK 6f2e566742dcef43b6b264bf7a66675aab885683

PR #825, opened 2025-08-01
Expose `Wallet::create_single` method

thunderbiscuit:
This PR exposes the [`Wallet::create_single()`](https://docs.rs/bdk_wallet/2.0.0/bdk_wallet/struct.Wallet.html#method.create_single) method.

I prefer to use single-descriptor wallets for my Tatooine faucet (easier to sync and keep track of), and realized we didn't bring this back into the bindings API after it was merged in Rust a while back.

TODO: Needs at least a test or two! Some methods should fail I think when the wallet is a single-descriptor, so we can test for that too.

### Changelog notice

```md
Added:
    - Expose `Wallet::create_single` constructor on Wallet type [#825]

[#825]: https://github.com/bitcoindevkit/bdk-ffi/pull/825
```

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/src/wallet.rs
+++ b/bdk-ffi/src/wallet.rs
@@ -58,9 +58,7 @@ impl Wallet {
                 .network(network)
                 .lookahead(lookahead)
                 .create_wallet(deref)
-                .map_err(|e| CreateWithPersistError::Persist {
-                    error_message: e.to_string(),
-                })?;
+                .map_err(CreateWithPersistError::from)?;
 
         Ok(Wallet {
             inner_mutex: Mutex::new(wallet),
@@ -100,9 +98,7 @@ impl Wallet {
             .network(network)
             .lookahead(lookahead)
             .create_wallet(deref)
-            .map_err(|e| CreateWithPersistError::Persist {
-                error_message: e.to_string(),
-            })?;
+            .map_err(CreateWithPersistError::from)?;
 
         Ok(Wallet {
             inner_mutex: Mutex::new(wallet),
@@ -134,9 +130,7 @@ impl Wallet {
                 .network(network)
                 .lookahead(lookahead)
                 .create_wallet(deref)
-                .map_err(|e| CreateWithPersistError::Persist {
-                    error_message: e.to_string(),
-                })?;
+                .map_err(CreateWithPersistError::from)?;
 
         Ok(Wallet {
             inner_mutex: Mutex::new(wallet),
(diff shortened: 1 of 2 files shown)


bdk-ffi/src/wallet.rs
```diff
@@ -67,6 +67,48 @@ impl Wallet {
...
+    /// Build a new single descriptor `Wallet`.
+    ///
+    /// If you have previously created a wallet, use `Wallet::load` instead.
+    ///
+    /// # Note
+    ///
+    /// Only use this method when creating a wallet designed to be used with a single
+    /// descriptor and keychain. Otherwise the recommended way to construct a new wallet is
+    /// by using `Wallet::create`. It's worth noting that not all features are available
+    /// with single descriptor wallets, for example setting a `change_policy` on `TxBuilder`
```
reez, line 77: > It's worth noting that not all features are available with single descriptor wallets, for example setting a `change_policy` on `TxBuilder` ...

are these just silent failures?

bdk-ffi/src/wallet.rs
```diff
@@ -67,6 +67,48 @@ impl Wallet {
...
 
+    /// Build a new single descriptor `Wallet`.
+    ///
+    /// If you have previously created a wallet, use `Wallet::load` instead.
+    ///
+    /// # Note
+    ///
+    /// Only use this method when creating a wallet designed to be used with a single
+    /// descriptor and keychain. Otherwise the recommended way to construct a new wallet is
+    /// by using `Wallet::create`. It's worth noting that not all features are available
```
reez: For FFI doc comments would  `by using Wallet::create` be `by using Wallet::new`?

bdk-ffi/src/wallet.rs
thunderbiscuit: Yep I think these just silently fail and don't do anything.

bdk-ffi/src/wallet.rs
thunderbiscuit: Good catch!

thunderbiscuit:
Great idea @reez on cleaning up the errors on the constructor. I added that fix in a second commit. Ready for re-review!

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/WalletTest.kt
```diff
@@ -45,4 +45,24 @@ class WalletTest {
             actual = wallet.balance().total.toSat()
         )
     }
+
+    // Single-descriptor wallet returns an address on the external keychain if even if change descriptor is not provided
```
ItoroD: Just a nit here.

comment says `keychain if even if change` first if should be removed

ItoroD: Tests work as expected!

thunderbiscuit: Fixed.

PR #826, opened 2025-08-01
Test Android library in CI

thunderbiscuit:
This is a really big achievement, because we have not been able to run the tests for the library in an Android emulator on CI for years!

Note that the live tests were failing because of:
1. The BDK signet Esplora instance has been down for a while. Once that was fixed, the mempool.space instance,
2. We ran into the low balance issue, where test had emptied the wallet over time. Once that was fixed, we ran into the mempool.space request max issue, where you can only make 700 requests per hour, but on big wallets you exceed that super quick.

So I JUST COMMENTED OUT ALL LIVE TESTS

As discussed over dev calls, we'll be focusing on unit tests here, and add an examples directory with more integration tests. In the long run we are also looking at adding a full regtest environment to the CI so we can run more of those tests that require syncing the wallet, without requiring Signet or Testnet 3 or 4.

### DIFF
--- a/.github/workflows/test-android.yaml
+++ b/.github/workflows/test-android.yaml
@@ -10,30 +10,16 @@ on:
       - "bdk-ffi/**"
       - "bdk-android/**"
 
-# The default Android NDK on the ubuntu-24.04 image is 25.2.9519653
-
 permissions: {}
 
 jobs:
   build:
     runs-on: ubuntu-24.04
     steps:
-      - name: "Show default version of NDK"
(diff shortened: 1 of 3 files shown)


thunderbiscuit:
Taking the action from here: https://github.com/ReactiveCircus/android-emulator-runner.

reez:
> The zizmor workflow fails but I think fixing it is for another PR.

Cool, i think I understand what its asking for so I’ll be able to review that follow up pr pretty easily

PR #827, opened 2025-08-04
feat: Add Kyoto example

ItoroD:
### Description

CBF syncing that works with regtest. 

### Changelog notice
- Add Kyoto example

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-jvm/examples/src/main/kotlin/Kyoto.kt
+++ b/bdk-jvm/examples/src/main/kotlin/Kyoto.kt
@@ -1,0 +1,85 @@
+package org.bitcoindevkit
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.cancelAndJoin
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+import java.nio.file.Files
+import java.nio.file.Paths
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+
+
+fun main() {
+    // Regtest environment must have Compact block filter enabled to run this
+    // Setup your environment according to the https://github.com/thunderbiscuit/podman-regtest-infinity-pro documentation.
+
+    // Add your regtest IP address
+    val ip: IpAddress = IpAddress.fromIpv4(127u, 0u, 0u, 1u)
+    val peer: Peer = Peer(ip, 18444u, false)
+    val peers: List<Peer> = listOf(peer)
+    val currentPath = Paths.get(".").toAbsolutePath().normalize()
+    val persistenceFilePath = Files.createTempDirectory(currentPath, "tempDirPrefix_")
+
+    val wallet = getNewWallet(ActiveWalletScriptType.P2WPKH, Network.REGTEST)
+    val address = wallet.revealNextAddress(KeychainKind.EXTERNAL)
+
+    // Fund this address. Send coins from your regtest to this address
+    println("Receiving address. Send funds to this address: ${address.address}")
+
+    // Wait 70 seconds for funds to arrive before syncing
+    println("Waiting 70 seconds for funds to arrive here ${address.address} before kyoto (compact block filter syncing) ...")
+    Thread.sleep(10000)
+
+    // Create CBF node and client
+    runBlocking {
+        val lightClient = CbfBuilder()
+            .peers(peers)
+            .connections(1u)
+            .scanType(ScanType.New)
+            .dataDir(persistenceFilePath.toString())
+            .build(wallet)
+        val client = lightClient.client
+        val node = lightClient.node
+
+        val logJob = launchLogCollector(this, client::nextLog, "LOG")
+        val logWarningJob = launchLogCollector(this, client::nextWarning, "WARNING")
+        val logInfoJob = launchLogCollector(this, client::nextInfo, "INFO")
+
+        //Start CBF node
+        node.run()
+        println("Node running")
+
+        //Update wallet
+        val update: Update = client.update()
+        wallet.applyUpdate(update)
+        println("Wallet balance: ${wallet.balance().total.toSat()} sats")
+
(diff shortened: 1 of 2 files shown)


bdk-jvm/examples/src/main/kotlin/Kyoto.kt
```diff
@@ -0,0 +1,64 @@
...
+    runBlocking {
+        val lightClient = CbfBuilder()
+            .peers(peers)
+            .connections(1u)
+            .scanType(ScanType.New)
+            .dataDir(persistenceFilePath.toString())
+            .build(wallet)
+        val client = lightClient.client
+        val node = lightClient.node
+        val logJob = launch {
```
rustaceanrob: Maybe for this example we will also want to print the info and warnings in separate tasks with `client.nextInfo()` and `client.nextWarning()`.

Btw is there a prettier logger in Kotlin than the standard `println`?

bdk-jvm/examples/src/main/kotlin/Kyoto.kt
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+        //Start CBF node
+        node.run()
+        println("Node running")
+
+        //Update wallet
+        val update: Update = client.update()
+        wallet.applyUpdate(update)
+        println("Wallet balance: ${wallet.balance().total.toSat()}")
```
rustaceanrob: Needs an annotation of `satoshis` or `sats`

bdk-jvm/examples/src/main/kotlin/Kyoto.kt
```diff
@@ -0,0 +1,64 @@
...
+    val persistenceFilePath = Files.createTempDirectory(currentPath, "tempDirPrefix_")
+
+    val wallet = getNewWallet(ActiveWalletScriptType.P2WPKH, Network.REGTEST)
+    val address = wallet.revealNextAddress(KeychainKind.EXTERNAL)
+
+    // Fund this address. Send coins from your regtest to this address
+    println("Receiving address. Send funds to this address: ${address.address}")
+
+    // Wait 70 seconds for funds to arrive before syncing
+    println("Waiting 70 seconds for funds to arrive here ${address.address} before kyoto (compact block filter syncing) ...")
```
rustaceanrob, line 32: If the wallet has a non-zero amount at start up, maybe we can skip the funding?

bdk-jvm/examples/src/main/kotlin/Kyoto.kt
```diff
@@ -0,0 +1,64 @@
...
+fun main() {
+    // Regtest environment must have Compact block filter enabled to run this
+    // Setup your environment according to the https://github.com/thunderbiscuit/podman-regtest-infinity-pro documentation.
+
+    // Add your regtest IP address
+    val ip: IpAddress = IpAddress.fromIpv4(127u, 0u, 0u, 1u)
+    val peer: Peer = Peer(ip, 18444u, false)
+    val peers: List<Peer> = listOf(peer)
+    val currentPath = Paths.get(".").toAbsolutePath().normalize()
+    val persistenceFilePath = Files.createTempDirectory(currentPath, "tempDirPrefix_")
```
rustaceanrob, line 23: Since this is an example, maybe we can preserve the data in between syncs

bdk-jvm/examples/src/main/kotlin/Kyoto.kt
ItoroD: Great idea. I will add those.

Yes, there are prettier loggers.

bdk-jvm/examples/src/main/kotlin/Kyoto.kt
ItoroD: Not sure its sufficient enough to just say if non-zero amount, as they might want to send more than what they have. Maybe checking if amount is sufficient enough for example amount then we can skip.

bdk-jvm/examples/src/main/kotlin/Kyoto.kt
ItoroD: I see what you mean. I think that totally makes sense for a Kyoto example. Will be great to see the progressive update to your wallet on every sync. 

But in this example, each run is getting a new wallet and generating descriptors for those wallets for you (The wallet is non-persistent). User is not adding descriptors of their own. So the idea of them seeing updates between syncs on same wallet will need a little tweaking of this example. 

That said I am thinking of maybe creating another Kyoto example that allows persisting, adding your own wallet descriptors, persisting your wallet data, allows/shows user how to play around by sending funds to see new sync updates. Will like to know your thoughts as I think your idea makes a lot of sense

thunderbiscuit:
I agree with Rob's review comments, except maybe for the persistence of the data (at least in this particular example).

It looks good to me overall, my only thought is that this is in conflict with #824.

If we merge 824 first, it doesn't make sense to have this anymore, and if we merge this one first, it will get wiped by 824 in any case a few days later. I would suggest closing this PR and opening the same PR on [bdk-jvm](https://github.com/bitcoindevkit/bdk-jvm).

bdk-jvm/examples/src/main/kotlin/Kyoto.kt
rustaceanrob: Not sure how difficult this would be in Kotlin, but you could potentially prompt the user with `y/n` if they would like to add funds.

ItoroD: > Not sure how difficult this would be in Kotlin, but you could potentially prompt the user with `y/n` if they would like to add funds.

Pretty simple. For this example its always a new wallet gotten so balance will always be 0 at start. 

I will consider adding that in a more interactive example.

ItoroD:
Closing this. I will reopen in bdk-jvm

PR #829, opened 2025-08-12
Set `lookahead` with `Wallet::load`

rustaceanrob:
### Description

Found by @andreasgriffin, the user may want to specify a `lookahead` when loading a wallet from storage, but we missed this after adding it to `new`. AFICT for each language supported this would not be a breaking change.

### Notes to the reviewers

Fixed a clippy lint as well

### Changelog notice

- Set `lookahead` on `Wallet::load`

#### All Submissions:

### DIFF
--- a/bdk-ffi/src/wallet.rs
+++ b/bdk-ffi/src/wallet.rs
@@ -445,7 +445,7 @@ impl Wallet {
 }
 
 impl Wallet {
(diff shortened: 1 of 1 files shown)


reez (approved):
ACK bc3613a2dba2fb21181622cf1908a2928a36dba1

default lookahead looks good https://docs.rs/bdk_chain/0.23.1/bdk_chain/indexer/keychain_txout/constant.DEFAULT_LOOKAHEAD.html

PR #840, opened 2025-08-28
feat: Create trigger workflow trigger for bdk-jvm

ItoroD:
### Description

This would allow bdk-jvm workflow file to be triggered when a merge happens. 

### Notes to the reviewers

- A repo_access token needs to be created by someone with access to both bdk-ffi and bdk-jvm. I only have access (as collaborator) on bdk-ffi but not on bdk-jvm.
- In this pr we are only concerned about changes to the bdk-ffi directory. No need to run test of specific bindings if push was not done in this folder.
- The dispatcher requires url for the repo where i needed to add the repo name and owner I believe the below is correct?

`https://api.github.com/repos/{owner}/{repo}/dispatches \`

where name and owner are

`https://api.github.com/repos/bdk/bdk-jvm/dispatches \`

See docs
- [Here](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#create-a-repository-dispatch-event)

#### All Submissions:

### DIFF
--- a/.github/workflows/trigger-external-bindings.yml
+++ b/.github/workflows/trigger-external-bindings.yml
@@ -1,0 +1,27 @@
+name: Trigger External Bindings Workflow
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+    paths:
+      - 'bdk-ffi/**'
+
+permissions: {}
+
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+    steps:    
+    #Trigger bdk-jvm-test in bdk-jvm repository
+      - name: Trigger bdk-jvm-test in bdk-jvm repository
+        if: ${{ github.ref_name == 'master' }}
+        run: |
+          response=$(curl -X POST \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: token ${{ secrets.BDK_JVM_ACCESS_TOKEN }}" \
+            https://api.github.com/repos/bitcoindevkit/bdk-jvm/dispatches \
+            -d '{"event_type":"trigger-bdk-jvm-tests"}')
+          echo "Response: $response"
+          echo "HTTP Status Code: $(echo $response | jq -r '.status')"
+          echo "Response Body: $(echo $response | jq -r '.body')"


.github/workflows/trigger-external-bindings.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-24.04
+    steps:    
+    #Trigger bdk-jvm-test in bdk-jvm repository
+      - name: Trigger bdk-jvm-test in bdk-jvm repository
+        if: ${{ (github.ref_name == 'main' || github.ref_type == 'tag') }}
+        run: |
+          response=$(curl -X POST \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: token ${{ secrets.REPO_ACCESS_TOKEN }}" \
+            https://api.github.com/repos/bdk/bdk-jvm/dispatches \
```
thunderbiscuit: I'm playing around with this and one issue I found was that this line is incorrect. The url would be

```sh
https://api.github.com/repos/bitcoindevkit/bdk-jvm/dispatches
```

thunderbiscuit:
I'm playing around with this and one issue I found was that this line 31 is incorrect. The url would be

```sh
https://api.github.com/repos/bitcoindevkit/bdk-jvm/dispatches
```

thunderbiscuit:
Our default branch is also called `master`, so this line should be fixed:

```yaml
if: ${{ (github.ref_name == 'main' || github.ref_type == 'tag') }}
```

To use the name of the branch.

ItoroD:
> Our default branch is also called `master`, so this line should be fixed:
> 
> ```yaml
> if: ${{ (github.ref_name == 'main' || github.ref_type == 'tag') }}
> ```
> 
> To use the name of the branch.

Thats right! Do you think we should also run on tags? or master is sufficient ? I was thinking default branch might just be enough

thunderbiscuit:
I think default branch is enough too.

ItoroD:
> BDK_JVM_ACCESS_TOKEN

That makes sense

.github/workflows/trigger-external-bindings.yml
```diff
@@ -0,0 +1,29 @@
...
+          response=$(curl -X POST \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: token ${{ secrets.BDK_JVM_ACCESS_TOKEN }}" \
+            https://api.github.com/repos/bitcoindevkit/bdk-jvm/dispatches \
+            -d '{"event_type":"trigger-bdk-jvm-tests"}')
+          echo "Response: $response"
+          echo "HTTP Status Code: $(echo $response | jq -r '.status')"
+          echo "Response Body: $(echo $response | jq -r '.body')"
+        env:
+          REPO_ACCESS_TOKEN: ${{ secrets.BDK_JVM_ACCESS_TOKEN }}
```
thunderbiscuit: This line defines an environment variable called REPO_ACCESS_TOKEN, but then line 22 doesn't use it directly, it also tries to get the data from the secrets. Is that on purpose? Should one of these lines be removed?

ItoroD: it should be taken out yes. I used that environment variable in my test before. Sorry!

PR #846, opened 2025-09-03
feat: Add Input for PSBT

ItoroD:
### Description
Since signing will soon be done with the PSBT type, I think its a good time to expose some of the other PSBT fields like `Input`. 

This will allow easier access to the input in a PSBT. It might be specifically important because the sign method of PSBT can do interesting stuff like returning a map of input index and keys used to sign it.

I would like to take a pause here and see how this is looking on the bindings side before I go any further.

### Changelog notice
- Add input type for psbt

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -513,7 +513,7 @@ pub struct TapScriptEntry {
 #[derive(Clone, Debug, uniffi::Record, Hash, Eq, PartialEq)]
 pub struct TapKeyOrigin {
     /// leaf hashes as hex strings
-    pub leaf_hashes: Vec<String>,
+    pub tap_leaf_hashes: Vec<String>,
     /// key source
     pub key_source: KeySource,
 }
@@ -564,7 +564,7 @@ pub struct TapScriptSigKey {
     pub xonly_pubkey: String,
     /// Taproot-tagged hash with tag "TapLeaf".
     /// This is used for computing tapscript script spend hash.
-    pub leaf_hash: String,
+    pub tap_leaf_hash: String,
 }
 
 /// A key-value map for an input of the corresponding index in the unsigned transaction.
@@ -692,7 +692,7 @@ impl From<&BdkInput> for Input {
                 .map(|(k, v)| {
                     let key = TapScriptSigKey {
                         xonly_pubkey: k.0.to_string(),
-                        leaf_hash: k.1.to_string(),
+                        tap_leaf_hash: k.1.to_string(),
                     };
                     (key, v.to_vec())
                 })
@@ -720,7 +720,7 @@ impl From<&BdkInput> for Input {
                 .map(|(k, v)| {
                     let key = k.to_string();
                     let value = TapKeyOrigin {
-                        leaf_hashes: v.0.iter().map(|h| h.to_string()).collect(),
+                        tap_leaf_hashes: v.0.iter().map(|h| h.to_string()).collect(),
                         key_source: KeySource {
                             fingerprint: v.1 .0.to_string(),
                             path: v.1 .1.to_string(),


bdk-ffi/src/bitcoin.rs
```diff
@@ -500,6 +502,249 @@ impl Display for Transaction {
...
+    pub final_script_sig: Option<Arc<Script>>,
+
+    /// The finalized, fully-constructed scriptWitness with signatures and any
+    /// other scripts necessary for this input to pass validation.
+    pub final_script_witness: Option<Vec<Vec<u8>>>,
+    /// RIPEMD160 hash to preimage map.
+    pub ripemd160_preimages: Vec<PreimagePair>,
+    /// SHA256 hash to preimage map.
+    pub sha256_preimages: Vec<PreimagePair>,
+    /// HSAH160 hash to preimage map.
```
reez: nit: Should be "HASH160"

bdk-ffi/src/bitcoin.rs
```diff
@@ -500,6 +502,249 @@ impl Display for Transaction {
...
+    pub hash160_preimages: Vec<PreimagePair>,
+    /// HAS256 hash to preimage map.
```
reez: nit: Should be "HASH256"

bdk-ffi/src/bitcoin.rs
```diff
@@ -500,6 +502,249 @@ impl Display for Transaction {
...
+    /// master key fingerprint
+    pub fingerprint: String,
+    /// derivation path segments
+    pub path: String,
+}
+
+#[derive(Clone, Debug, uniffi::Record)]
+pub struct KeyValuePair {
+    /// raw key bytes (for proprietary / unknown keys)
+    pub key: Vec<u8>,
```
reez: are we missing `type_value`?

bdk-ffi/src/bitcoin.rs
```diff
@@ -500,6 +502,249 @@ impl Display for Transaction {
...
+                    path: v.1 .1.to_string(),
+                })
+                .collect::<Vec<TapKeyOrigin>>(),
+            tap_internal_key: input.tap_internal_key.as_ref().map(|k| k.to_string()),
+            tap_merkle_root: input.tap_merkle_root.as_ref().map(|k| k.to_string()),
+            proprietary: input
+                .proprietary
+                .iter()
+                .map(|(k, v)| KeyValuePair {
+                    key: k.to_key().key.clone(),
```
reez: are we missing `prefix` & `subtype`?

bdk-ffi/src/bitcoin.rs
```diff
@@ -500,6 +502,249 @@ impl Display for Transaction {
...
+                    fingerprint: v.1 .0.to_string(),
+                    path: v.1 .1.to_string(),
+                })
+                .collect::<Vec<TapKeyOrigin>>(),
+            tap_internal_key: input.tap_internal_key.as_ref().map(|k| k.to_string()),
+            tap_merkle_root: input.tap_merkle_root.as_ref().map(|k| k.to_string()),
+            proprietary: input
+                .proprietary
+                .iter()
+                .map(|(k, v)| KeyValuePair {
```
reez: why KVP instead of `ProprietaryKeyValuePair`?

bdk-ffi/src/bitcoin.rs
ItoroD: Yes, that missing. I will fix that tomorrow.

Thanks!

bdk-ffi/src/bitcoin.rs
ItoroD: You are correct using KVP is the reason I missed prefix & subtype in your previous [comment](https://github.com/bitcoindevkit/bdk-ffi/pull/846#discussion_r2373218669). I will fix this. Thanks for the find.

bdk-ffi/src/bitcoin.rs
```diff
@@ -500,6 +502,267 @@ impl Display for Transaction {
     }
 }
 
+#[derive(Clone, Debug, uniffi::Record)]
+pub struct PartialSig {
+    pub pubkey: String,
+    pub signature: Vec<u8>,
+}
```
thunderbiscuit: Since the [rust-bitcoin](https://docs.rs/bitcoin/latest/src/bitcoin/psbt/map/input.rs.html#82) type for this field is a map, I think we could use the uniffi equivalent here instead of a vector of structs. Let me know if you've tried this and it doesn't work for some reason!

bdk-ffi/src/bitcoin.rs
```diff
@@ -500,6 +502,267 @@ impl Display for Transaction {
...
+
+#[derive(Clone, Debug, uniffi::Record)]
+pub struct Bip32Derivation {
+    /// hex-encoded public key (serialized)
+    pub pubkey: String,
+    /// master key fingerprint
+    pub fingerprint: String,
+    /// derivation path segments
+    pub path: String,
+}
```
thunderbiscuit: Same as above. We might be able to use a map instead of a Vec of this struct.

bdk-ffi/src/bitcoin.rs
```diff
@@ -500,6 +502,267 @@ impl Display for Transaction {
...
+    /// other scripts necessary for this input to pass validation.
+    pub final_script_witness: Option<Vec<Vec<u8>>>,
+    /// RIPEMD160 hash to preimage map.
+    pub ripemd160_preimages: Vec<PreimagePair>,
+    /// SHA256 hash to preimage map.
+    pub sha256_preimages: Vec<PreimagePair>,
+    /// HASH160 hash to preimage map.
+    pub hash160_preimages: Vec<PreimagePair>,
+    /// HASH256 hash to preimage map.
+    pub hash256_preimages: Vec<PreimagePair>,
```
thunderbiscuit: Same for these.

ItoroD:
> It's looking good!

Thanks

> I think a few of the Vec could be simply HashMap<K, V> as in the rust-bitcoin codebase, unless there is something I'm missing.
>
> It's a lot of conversion between simple types, so I'm wondering if stealing one or two of the rust-bitcoin tests on this type would be good as well? I haven't looked at their tests so don't take that as a requirement. I don't know if they actually do a lot of tests just on this Input type.

They do have a few test that concern input. Though they are not direct tests. The intention of some of the test are to validate correctness of an action performed, like signing. See [here ](https://github.com/rust-bitcoin/rust-bitcoin/blob/b620233a5d9076a35ad36dc8e39ca26114caf31e/bitcoin/tests/psbt-sign-taproot.rs#L143) I am sure I can squeeze out a few good test like this from there.  Will be a great idea to have them!

An additional thing I will have to perform is ensure whatever Psbt sample I create/use has values for the fields I will be checking/testing. Might take a while to do that for all fields. (Simulating different Psbt states, signed, unsigned, partially signed, finalized ). I could start with a few.

bdk-ffi/src/bitcoin.rs
ItoroD: Yes, you are correct. I struggled with trying to use BTreeMap in a record this made me think maps are not just one of the Uniffi compatible type. HashMap is definitely closer and it works. I just checked. I did 

`pub bip32_derivation: HashMap<String, KeySource>` 

instead of 

`pub bip32_derivation: Vec<Bip32Derivation>`

where keysource is 


```
#[derive(Clone, Debug, uniffi::Record)]
pub struct KeySource {
    pub fingerprint: String,
    pub path: String,
}
```

bdk-ffi/src/bitcoin.rs
ItoroD: Thanks for pointing these out. BTreeMap not working threw me off. But I see [now](https://mozilla.github.io/uniffi-rs/0.27/udl/builtin_types.html)

bdk-ffi/src/bitcoin.rs
ItoroD: I was thinking about this again, the only thing about collecting these into Hashmap is that we loose the ordering that the BTreeMap gave from the actual rust-bitcoin types. Should we care about the ordering ?

thunderbiscuit: Yeah I wondered that too. But looking at the fields, I do think the keys identify everything you need, and so the ordering doesn't matter. If anything devs would need an extra line of lookup to find the associated data instead of relying on the ordering, but even then I don't know if that's even how people use the BTreeMap for these fields. My hunch would be to use the HashMap.

ItoroD: Alright, I will update. Thanks.

ItoroD:
Just an update on this. I will be pushing new changes today with some tests as well

bdk-ffi/src/bitcoin.rs
```diff
@@ -500,6 +503,244 @@ impl Display for Transaction {
     }
 }
 
+#[derive(Clone, Debug, uniffi::Record)]
+pub struct PartialSig {
```
thunderbiscuit: Sorry I think this type can now be removed.

bdk-ffi/src/bitcoin.rs
```diff
@@ -500,6 +503,244 @@ impl Display for Transaction {
...
+    /// HASH256 hash to preimage map.
+    pub hash256_preimages: HashMap<String, Vec<u8>>,
+    /// Serialized taproot signature with sighash type for key spend.
+    pub tap_key_sig: Option<Vec<u8>>,
+    /// Map of `<xonlypubkey>|<leafhash>` with signature.
+    pub tap_script_sigs: Vec<TapScriptSig>,
+    /// Map of Control blocks to Script version pair.
+    pub tap_scripts: Vec<TapScriptEntry>,
+    /// Map of tap root x only keys to origin info and leaf hashes contained in it.
+    pub tap_key_origins: Vec<TapKeyOrigin>,
```
thunderbiscuit: These 3 types can be made HashMaps as well.

bdk-ffi/src/bitcoin.rs
```diff
@@ -500,6 +503,244 @@ impl Display for Transaction {
...
+    /// Taproot Internal key.
+    pub tap_internal_key: Option<String>,
+    /// Taproot Merkle root.
+    pub tap_merkle_root: Option<String>,
+    /// Proprietary key-value pairs for this input.
+    pub proprietary: Vec<ProprietaryKeyValuePair>,
+    /// Unknown key-value pairs for this input.
+    pub unknown: Vec<KeyValuePair>,
```
thunderbiscuit: And these last two as well.

bdk-ffi/src/bitcoin.rs
ItoroD: yh no worries. I had actually done these locally but i though you wanted them excluded intentionally because of the nature of the key.

tap_script_sigs is a concatenation of xonlypubkey and leafhash
`<xonlypubkey>|<leafhash>`

thunderbiscuit: Yeah I'd keep the tuple (XOnlyPublicKey, TapLeafHash) as a struct similar to what you have above, but use that as my key on the HashMap, akin to the BTreeMap they have in Rust.

ItoroD: got it

thunderbiscuit (approved):
Tested ACK d24fe5266b5675bb8e13aa2417a10dd9abef7f25

Issue #834, opened 2025-08-18
Update to bdk_wallet 2.1.0


reez:
**Describe the enhancement**  

Update to bdk_wallet 2.1.0

https://github.com/bitcoindevkit/bdk_wallet/releases/tag/wallet-2.1.0

**Additional context**

Test out things like multipath descriptors

reez:
Was gonna open up an issue about testing multipath descriptors, but looks like I've already got it here.

thunderbiscuit:
This is now in, with the multipath constructor to boot!

PR #847, opened 2025-09-03
Expose `Wallet::create_from_two_path_descriptor` constructor

thunderbiscuit:
This was requested in #834. 
cc @reez.
See [Wallet::create_from_two_path_descriptor](https://docs.rs/bdk_wallet/2.1.0/bdk_wallet/struct.Wallet.html#method.create_from_two_path_descriptor).

### Note to Reviewers

One of the odd things about this API is that you cannot use private extended keys in your multipath descriptor to create wallets. In other words you can only use this API for with xpubs.

This looked so odd to me that I had to go down the rabbit hole and find out that indeed that's the case in Rust too. I will open a small documentation PR mentionning this in the API docs because that's not obvious at first glance.

In any case I added tests for both cases (pass and fail). Here I just took the descriptor used in the Rust tests, but I think we should add to our Constants file a number of recovery phrases with xpub and xprv keys that could then be used for this tests. I just didn't want to muddy the PR with new extra test fluff.

### Changelog notice

```md
Added
    - Add `Wallet::create_from_two_path_descriptor` constructor [#847]

[#847]: https://github.com/bitcoindevkit/bdk-ffi/pull/847
```

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/CreatingWalletTest.kt
+++ b/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/CreatingWalletTest.kt
@@ -42,6 +42,32 @@ class CreatingWalletTest {
         )
     }
 
+    // Create a wallet with a public multipath descriptor.
+    @Test
+    fun createWalletWithMultipathDescriptor() {
+        val multipathDescriptor = Descriptor(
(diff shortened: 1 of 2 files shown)


reez (approved):
ACK b433b37d2930334a4e11d943eccca53bf0f3e7c1

I'll test this at some point in the iOS app hopefully, not a blocker to a 2.1 release for us, but def a nice-to-have.

thunderbiscuit:
I'll open the PR and new test and see how exactly we want to write it down. I'll merge this fix here as well once it's approved in bdk_wallet.

PR #850, opened 2025-09-04
Migrate Network type to proc macros (and remove the last of the UDL)!

### LINKED ISSUE #806: Remove `.udl`
**Describe the enhancement**  

We have migrated away from `.udl` to proc macros almost completely, so it would be nice to finish the "completely" part and remove the "almost".

**Problem**

We don't seem to be able to totally migrate away right now thought because of `Network`.

I don't think we can remove this at the moment, because the `bitcoin::Network` enum is marked `#[non_exhaustive]` in the bitcoin crate, making it incompatible with UniFFI's proc macro system.

Non-exhaustive enums can have new variants added in future versions but UniFFI's `#[derive(uniffi::Enum)]` requires exhaustive pattern matching for safe code gen (`[NonExhaustive, Remote]` attributes in `.udl` are a workaround that allows UniFFI to compile while acknowledging the limitation).

So the only option I can think of to expose it via proc macros would be we'd need to create our own local exhaustive enum and manually convert between it and `bitcoin::Network` (not saying that's what we'd want to do at all though even if it "works"!).

If you have any other ideas (or if I'm misunderstanding what I described above!) let me know.

thunderbiscuit:
This one is just a crowd pleaser. It's the one you pull out at the 3rd encore. REMOVING THE DARN UDL YESSSS BOYSSSSSS.

There are small changes to the scaffolding calls once you remove all of the udl file. Nothing major.

Fixes #806.

### Changelog notice

No changelog notice.

#### All Submissions:


reez (approved):
ACK 7fcb1d7a414959430b52896707127460607f7407

Its such a crowd pleaser I'm too eager to review and ACK before I was even asked for one

PR #851, opened 2025-09-05
Update Rust compiler to 1.85.1

### LINKED ISSUE #828: Update Rust version to 1.85.0
We currently build with Rust 1.84.1, which is about to be below the bdk crates' [MSRV of 1.85.0](https://github.com/bitcoindevkit/bdk_wallet/issues/289).

1.89.0 will [come out as stable tomorrow (August 7)](https://releases.rs/), so this would be a good time to bump.

thunderbiscuit:
This fixes #828.

A few things are noteworthy for reviewers here. First, I tried the bump just like we always do: updating the CI files and the build scripts. You need to install the new toolchain to then build, and it can be done like this:

```shell
rustup toolchain install 1.85.1
rustup default 1.85.1
```

But after taking another look at the uniffi repo and the way Mozilla does it, I noticed they were pinning the Rust compiler version in a file called `rust-toolchain.toml`, which I had seen before in many Rust projects, including in bdk (which uses the older, `rust-version` file, but it's the same thing they just haven't updated yet). Upon digging into this and trying to see if it could be useful for us, it became clear that it's the wayyyyyyy cleaner way to do things.

Calling commands like cargo or rustup in a directory which contains this file will automatically tell cargo/rustup to override whatever version of the Rust compiler they consider default and use the pinned version instead. This means we don't need to specify it explicitly in either the CI runs (and keep it in sync!), or in build scripts. It basically ensures everyone builds using the same version of the Rust compiler, but it's defined in one single place. WIN.

This is what I have implemented here.


reez (approved):
ACK 5f80de82daf7654b48fbd8869afbb3841a0bd6c4

That's awesome info you tracked down!

PR #853, opened 2025-09-08
Require explicit wildcard derivation for descriptor key types

### LINKED ISSUE #324: DescriptorSecretKey.extend() method requires a wildcard be applied to the root extended key
Our current implementation of the `.extend()` method carries over the wildcard choice made initially by the `DescriptorSecretKey` type. This means that if the DescriptorSecretKey was created without a wildcard, there is no way to extend it with a wildcard.

Example:
```kotlin
val rootDescriptorSecretKey = DescriptorSecretKey.fromString(
    "tprv8ghw3FWfWTeLCEXcr8f8Q8Lz4QPCELYv3jhBXjAm7XagA6R5hreeWLTJeLBfMj7Ni6Q3PdV1o8NbvNBHE59W97EkRJSU4JkvTQjaNUmQubE"
)
val keysAtNewPath = rootDescriptorSecretKey.extend(DerivationPath("m/1212"))
val descriptorTest1 = Descriptor(
    descriptor = "wpkh(${keysAtNewPath.asString()})",
    network = Network.TESTNET
)
println("Issue #324 descriptor is ${descriptorTest1.asStringPrivate()}")
// Issue #324 descriptor is wpkh(tprv8ghw3FWfWTeLCEXcr8f8Q8Lz4QPCELYv3jhBXjAm7XagA6R5hreeWLTJeLBfMj7Ni6Q3PdV1o8NbvNBHE59W97EkRJSU4JkvTQjaNUmQubE/1212)#8qetj2j9
```

We cannot apply the wildcard to the `DerivationPath` type, and so the only way to have a wilcard for this path is to create the root key with a wild card in the first place:
```kotlin
val rootDescriptorSecretKey = DescriptorSecretKey.fromString(
    "tprv8ghw3FWfWTeLCEXcr8f8Q8Lz4QPCELYv3jhBXjAm7XagA6R5hreeWLTJeLBfMj7Ni6Q3PdV1o8NbvNBHE59W97EkRJSU4JkvTQjaNUmQubE/*"
)
val keysAtNewPath = rootDescriptorSecretKey.extend(DerivationPath("m/1212"))
val descriptorTest1 = Descriptor(
    descriptor = "wpkh(${keysAtNewPath.asString()})",
    network = Network.TESTNET
)
println("Issue #324 descriptor is ${descriptorTest1.asStringPrivate()}")

// Issue #324 descriptor is wpkh(tprv8ghw3FWfWTeLCEXcr8f8Q8Lz4QPCELYv3jhBXjAm7XagA6R5hreeWLTJeLBfMj7Ni6Q3PdV1o8NbvNBHE59W97EkRJSU4JkvTQjaNUmQubE/1212/*)#lgncmkah
```

This little quirk of the API comes from the fact that in the FFI we extend the DescriptorSecretKey "directly", whereas the `.extend()` method in rust-bitcoin belongs to the `DerivationPath` type. This little bug (not really a bug but just a weird behaviour?) doesn't have an easy answer, because it might involve reshaping our API and the way we extend keys with custom paths.

Opening this issue for discussion.

thunderbiscuit:
This is a small issue that's been on my mind for a very long time. Fixes #324.

Our `DescriptorSecretKey` and `DescriptorPublicKey` types, upon creation, build extended keys that have a wildcard added by default. For example, creating a DescriptorSecretKey currently gives you something like this:

```rust
tprv8ZgxMBicQKsPdWuqM1t1CDRvQtQuBPyfL6GbhQwtxDKgUAVPbxmj71pRA8raTqLrec5LyTs5TqCxdABcZr77bt2KyWA5bizJHnC4g4ysm4h/*
```

This is a bit weird because... well you wouldn't often add a wildcard right there as your first derivation step. None of the bips do this. Rather, you often add wildcards once you've fully derived your key at the index you want it at. Moreover, you could not use our API as is to create _unwildcarded_ DescriptorSecretKeys or DescriptorPublicKeys. This is an odd limitation of the code as it stands, because in reality those are totally possible; we just didn't allow it with the current code. I suspect we didn't get a lot of feedback on this simply because it's not a part of the API that is used much.

This PR addresses that, and creates extended keys the way you'd expect them:

```rust
let mnemonic = Mnemonic::from_string("awesome awesome awesome awesome awesome awesome awesome awesome awesome awesome awesome awesome".to_string()).unwrap();
let key = DescriptorSecretKey::new(Network::Testnet, &mnemonic, None);
// key = "tprv8ZgxMBicQKsPdWAHbugK2tjtVtRjKGixYVZUdL7xLHMgXZS6BFbFi1UDb1CHT25Z5PU1F9j7wGxwUiRhqz9E3nZRztikGUV6HoRDYcqPhM4"
```

From there you can extend them however you want, including adding a wildcard to it if you'd like, using a new method on the type, `add_wildcard()`.

### Changelog notice

```md
## Changed
    - The `DescriptorSecretKey::new` constructor does not produce an extended key with an automatic wildcard anymore [#853]

## Added
    - The `DescriptorPublicKey::new` constructor [#853]
    - New `DescriptorPublicKey::add_wildcard` method, which adds a wildcard to the derivation path of the descriptor [#853]
    - New `DescriptorSecretKey::add_wildcard` method, which adds a wildcard to the derivation path of the descriptor [#853]

[#853]: https://github.com/bitcoindevkit/bdk-ffi/pull/853
```

#### All Submissions:

#### New Features:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,22 +6,31 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased](https://github.com/bitcoindevkit/bdk-ffi/compare/v2.3.0...HEAD)
 
+## Breaking
+
+- The `DescriptorSecretKey::new` constructor does not produce an extended key with an automatic wildcard anymore [#853]
+- `Descriptor` and `DescriptorSecretKey` constructors now require a `NetworkKind` [#986]
+
 ### Added
 
-- Expose miniscript `has_wildcard` and `sanity_check` methods on `Descriptor` type #945
-- Expose `new_wsh_sortedmulti` and `new_pk` methods on `Descriptor` type #949
-- Expose `new_sh_sortedmulti`, `new_sh_wsh_sortedmulti`, `new_pkh`, `new_wpkh`, `new_sh_wpkh` methods on `Descriptor` type #973
-- Expose wallet event helpers `Wallet::apply_unconfirmed_txs_events` and `Wallet::apply_evicted_txs_events` #971
-- Expose wallet scan/sync helpers `Wallet::start_full_scan_at`, `Wallet::start_sync_with_revealed_spks_at`, and `Wallet::checkpoints` #971
-- Expose locked outpoint wallet APIs and locked outpoint persistence in `ChangeSet` #971
-- Expose transaction builder controls for sighash, transaction ordering, and adding foreign UTXOs with explicit sequence values #971
-- Expose `Persister::get_pre_v1_wallet_keychains` for pre-v1 SQLite wallet migration support #971
-- Expose `NetworkKind` type #986
+- Expose miniscript `has_wildcard` and `sanity_check` methods on `Descriptor` type [#945]
+- Expose `new_wsh_sortedmulti` and `new_pk` methods on `Descriptor` type [#949]
+- Expose `new_sh_sortedmulti`, `new_sh_wsh_sortedmulti`, `new_pkh`, `new_wpkh`, `new_sh_wpkh` methods on `Descriptor` type [#973]
+- Expose wallet event helpers `Wallet::apply_unconfirmed_txs_events` and `Wallet::apply_evicted_txs_events` [#971]
+- Expose wallet scan/sync helpers `Wallet::start_full_scan_at`, `Wallet::start_sync_with_revealed_spks_at`, and `Wallet::checkpoints` [#971]
+- Expose locked outpoint wallet APIs and locked outpoint persistence in `ChangeSet` [#971]
+- Expose transaction builder controls for sighash, transaction ordering, and adding foreign UTXOs with explicit sequence values [#971]
+- Expose `Persister::get_pre_v1_wallet_keychains` for pre-v1 SQLite wallet migration support [#971]
+- Expose `NetworkKind` type [#986]
+- New `WildcardType` enum [#853]
+- New `DescriptorPublicKey::add_wildcard` method, which adds an unhardened wildcard to the derivation path of the descriptor [#853]
+- New `DescriptorSecretKey::add_wildcard(wildcard_type: WildcardType)` method, which adds a wildcard to the derivation path of the descriptor [#853]
 
-### Breaking
-
-- `Descriptor` and `DescriptorSecretKey` constructors now require a `NetworkKind` #986
-
+[#853]: https://github.com/bitcoindevkit/bdk-ffi/pull/853
+[#853]: https://github.com/bitcoindevkit/bdk-ffi/pull/945
+[#853]: https://github.com/bitcoindevkit/bdk-ffi/pull/949
+[#986]: https://github.com/bitcoindevkit/bdk-ffi/pull/971
+[#986]: https://github.com/bitcoindevkit/bdk-ffi/pull/973
 [#986]: https://github.com/bitcoindevkit/bdk-ffi/pull/986
 
 ## [v2.3.0]
--- a/bdk-ffi/src/keys.rs
+++ b/bdk-ffi/src/keys.rs
@@ -16,6 +16,7 @@ use bdk_wallet::keys::{
 use bdk_wallet::miniscript::descriptor::{DescriptorXKey, Wildcard};
 use bdk_wallet::miniscript::BareCtx;
 
+use crate::types::WildcardType;
 use std::convert::TryFrom;
 use std::fmt::Display;
 use std::str::FromStr;
@@ -132,14 +133,14 @@ impl Display for DerivationPath {
 impl_from_core_type!(BdkDerivationPath, DerivationPath);
 impl_into_core_type!(DerivationPath, BdkDerivationPath);
 
-/// A descriptor containing secret data.
+/// The descriptor secret key, either a single private key or an xprv.
 #[derive(Debug, uniffi::Object)]
 #[uniffi::export(Debug, Display)]
 pub struct DescriptorSecretKey(pub(crate) BdkDescriptorSecretKey);
 
 #[uniffi::export]
 impl DescriptorSecretKey {
-    /// Construct a secret descriptor using a mnemonic.
+    /// Construct a secret descriptor key using a mnemonic.
     #[uniffi::constructor]
     pub fn new(network_kind: NetworkKind, mnemonic: &Mnemonic, password: Option<String>) -> Self {
         let mnemonic = mnemonic.0.clone();
@@ -148,7 +149,7 @@ impl DescriptorSecretKey {
             origin: None,
             xkey: xkey.into_xprv(network_kind).unwrap(),
             derivation_path: BdkDerivationPath::master(),
-            wildcard: Wildcard::Unhardened,
+            wildcard: Wildcard::None,
         });
         Self(descriptor_secret_key)
     }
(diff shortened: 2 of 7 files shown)


bdk-ffi/src/keys.rs
```diff
@@ -83,23 +83,23 @@ impl DerivationPath {
...
     #[uniffi::constructor]
     pub fn new(network: Network, mnemonic: &Mnemonic, password: Option<String>) -> Self {
         let mnemonic = mnemonic.0.clone();
         let xkey: ExtendedKey = (mnemonic, password).into_extended_key().unwrap();
         let descriptor_secret_key = BdkDescriptorSecretKey::XPrv(DescriptorXKey {
             origin: None,
             xkey: xkey.into_xprv(network).unwrap(),
-            derivation_path: BdkDerivationPath::master(),
-            wildcard: Wildcard::Unhardened,
+            derivation_path: BdkDerivationPath::default(),
```
ItoroD: Just curious. Any reason why we are updating to default() here?

bdk-ffi/src/keys.rs
```diff
@@ -83,23 +83,23 @@ impl DerivationPath {
...
 #[derive(Debug, uniffi::Object)]
 #[uniffi::export(Debug, Display)]
 pub struct DescriptorSecretKey(pub(crate) BdkDescriptorSecretKey);
 
 #[uniffi::export]
 impl DescriptorSecretKey {
-    /// Construct a secret descriptor using a mnemonic.
+    /// Construct a secret descriptor key using a mnemonic.
     #[uniffi::constructor]
     pub fn new(network: Network, mnemonic: &Mnemonic, password: Option<String>) -> Self {
```
ItoroD: I would have said since "wildcardness" is being set in the new() method why don't we also just pass a parameter that determines whether wildcard should be added on create or not. But I think you are right, there is no real reason why wildcard would be placed immediately/at end of the `tprv...../*`. This will make sense to be added after the extension has been added as like a step 3 like in your `fn test_add_wildcard()` test below.

From running the test on here, I believe it should work seamlessly too on the language bindings side (android, jvm...)

bdk-ffi/src/keys.rs
```diff
@@ -199,6 +217,21 @@ pub struct DescriptorPublicKey(pub(crate) BdkDescriptorPublicKey);
...
+        let mnemonic = mnemonic.0.clone();
+        let xkey: ExtendedKey = (mnemonic, password).into_extended_key().unwrap();
+        let descriptor_public_key = BdkDescriptorPublicKey::XPub(DescriptorXKey {
+            origin: None,
+            xkey: xkey.into_xpub(network, &secp),
+            derivation_path: BdkDerivationPath::default(),
+            wildcard: Wildcard::None,
+        });
+        Self(descriptor_public_key)
+    }
```
reez: what do you think about this instead:

```
    pub fn new(network: Network, mnemonic: &Mnemonic, password: Option<String>) -> Self {
        let public_key = DescriptorSecretKey::new(network, mnemonic, password).as_public();
        Self(public_key.0.clone())
    }
```

bdk-ffi/src/keys.rs
```diff
@@ -256,6 +289,24 @@ impl DescriptorPublicKey {
         }
     }
 
+    pub fn add_wildcard(&self) -> Result<Arc<Self>, DescriptorKeyError> {
```
reez, line 356: after this line we could do:

```
                if matches!(descriptor_x_key.wildcard, Wildcard::Unhardened) {
                    return Ok(Arc::new(Self(self.0.clone())));
                }
```                
                
but its not totally necessary and nothings wrong with the code as is; but if we already have a wildcard we can just return self at that point I think

(same logic can apply after line 167 too if needed)

bdk-ffi/src/keys.rs
thunderbiscuit: 👍

bdk-ffi/src/keys.rs
thunderbiscuit: Nope I don't know why I added it. The default is indeed `BdkDerivationPath::master()`, so I'll keep it here.

bdk-ffi/src/keys.rs
thunderbiscuit: Good point, and this is making me realize that I don't really know why we'd have a constructor for a public key that uses a mnemonic (private key data really). If you want a `DescriptorPublicKey` from a mnemonic, it's cleaner to keep just one way of doing it: create the `DescriptorSecretKey` and then call `as_public()` on it. Removing now!

thunderbiscuit (commented):
Just responding to comments.

bdk-ffi/src/keys.rs
thunderbiscuit: I tried this just now, but in fact you can't shortcut the method that early, because the `descriptor_x_key` is an enum with 3 variants, so you have to check which one it is beforehand, basically sending you down the `match` path on the line below, and at that point there isn't much gain I think.

bdk-ffi/src/keys.rs
```diff
@@ -208,6 +208,24 @@ impl DescriptorSecretKey {
...
+        let descriptor_secret_key = &self.0;
+        match descriptor_secret_key {
+            BdkDescriptorSecretKey::Single(_) => Err(DescriptorKeyError::InvalidKeyType),
+            BdkDescriptorSecretKey::XPrv(descriptor_x_key) => {
+                let descriptor_secret_key_with_wildcard =
+                    BdkDescriptorSecretKey::XPrv(DescriptorXKey {
+                        origin: descriptor_x_key.origin.clone(),
+                        xkey: descriptor_x_key.xkey,
+                        derivation_path: descriptor_x_key.derivation_path.clone(),
+                        wildcard: Wildcard::Unhardened,
```
reez: What should `add_wildcard()` do when the key already has a hardened wildcard, like `/*h` or `/*'`? What we have here always replaces the wildcard with `Wildcard::Unhardened`, would that change the existing derivation semantics? Should we preserve the existing wildcard when one is already present? Let me know if Im thinking about any of this wrong though

bdk-ffi/src/keys.rs
reez: yep that make sense, good call

bdk-ffi/src/keys.rs
thunderbiscuit: Oh good catch. I never really work with hardened wildcards so for some reason this possibility didn't come to mind immediatly but yes, I should instead check whether there is a wildcard, and if there is just keep it as is. Thanks!

thunderbiscuit:
Ok a few things to note here:

1. I added an argument to the `DescriptorSecretKey::add_wildcard` method. It's a boolean that allows you to choose a hardened wildcard if you want. This is (I think?) less used, so I put a default value of `false` on it, but we could also remove the default if you guys prefer.
2. The behavior of the method is explained in comments, and basically goes as follows: if you ask for a wildcard that is the same as the one already on the DescriptorPrivateKey, nothing happens and you get the same key back. If your key didn't have a wildcard, then that wildcard gets added, and if your key already had a wildcard but it was of the _opposite_ type, we return an error.
3. This is a little different in the case of the `DescriptorPublicKey`, because using hardened wildcards there does not make sense, and no keys can be derived from it. In this case I also throw an error.

reez:
> Ok a few things to note here:
> 
> 1. I added an argument to the `DescriptorSecretKey::add_wildcard` method. It's a boolean that allows you to choose a hardened wildcard if you want. This is (I think?) less used, so I put a default value of `false` on it, but we could also remove the default if you guys prefer.
> 2. The behavior of the method is explained in comments, and basically goes as follows: if you ask for a wildcard that is the same as the one already on the DescriptorPrivateKey, nothing happens and you get the same key back. If your key didn't have a wildcard, then that wildcard gets added, and if your key already had a wildcard but it was of the _opposite_ type, we return an error.
> 3. This is a little different in the case of the `DescriptorPublicKey`, because using hardened wildcards there does not make sense, and no keys can be derived from it. In this case I also throw an error.

small API thought/question: use an enum instead of a bool for the hardened/unhardened choice?

Something like `WildcardType::Unhardened` / `WildcardType::Hardened`, would it read even better or not?

```
descriptor_secret_key.add_wildcard(WildcardType::Hardened)
descriptor_secret_key.add_wildcard(WildcardType::Unhardened)
```

vs 

```
descriptor_secret_key.add_wildcard(true)
descriptor_secret_key.add_wildcard(false)
```

I'm looking at https://docs.rs/miniscript/12.3.5/src/miniscript/descriptor/key.rs.html#181-190 too as inspiration (even though I dont think we'd want to include `None`?).

Not a blocker at all though.

My arguments against it are if argument name is visible, bool is pretty readable (but so many languages have so many different ways of handling this).

thunderbiscuit:
I like the idea actually (of adding the `WildcardType` enum). I'll try to make it clear in the code that this required a bit of cooking on our side and is not the direct Rust API, but in this case there isn't an easy way to just port it one for one, and I think this is a good way to do it.

thunderbiscuit:
@reez I added a new type as per your idea, `WildcardType`. It's one of our own custom types but I think in this instance it makes sense. Let me know what you think!

bdk-ffi/src/types.rs
```diff
@@ -1462,3 +1463,21 @@ impl From<bdk_wallet::TxDetails> for TxDetails {
         }
     }
 }
+
+/// Wildcard types for descriptors
+#[derive(uniffi::Enum, PartialEq)]
+pub enum WildcardType {
+    /// Unhardened wildcard, e.g. *
+    Unhardened,
+    /// Unhardened wildcard, e.g. *h
```
reez: nit: should we go with “Hardened wildcard” here? Looks like latest miniscript uses that wording too (I think they fixed in 13.0.0, and even though we are on 12.3.6 I think it still makes sense to use “Hardened wildcard”): 
https://docs.rs/miniscript/latest/src/miniscript/descriptor/key.rs.html#193-201

CHANGELOG.md
```diff
@@ -6,22 +6,33 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+- Expose transaction builder controls for sighash, transaction ordering, and adding foreign UTXOs with explicit sequence values [#971]
+- Expose `Persister::get_pre_v1_wallet_keychains` for pre-v1 SQLite wallet migration support [#971]
+- Expose `NetworkKind` type [#986]
 
 ### Breaking
 
-- `Descriptor` and `DescriptorSecretKey` constructors now require a `NetworkKind` #986
+- `Descriptor` and `DescriptorSecretKey` constructors now require a `NetworkKind` [#986]
+- New `DescriptorPublicKey::add_wildcard` method, which adds a wildcard to the derivation path of the descriptor [#853]
+- New `DescriptorSecretKey::add_wildcard(hardened: bool)` method, which adds a wildcard to the derivation path of the descriptor [#853]
```
reez: nit: mention WildcardType instead of the old `hardened: bool`?

reez (commented):
Overall this looks good to me, I just left two small nits

bdk-ffi/src/types.rs
thunderbiscuit: Yes good catch.

CHANGELOG.md
thunderbiscuit: Totally. Fixed it now.

reez (approved):
ACK 31ce47ceee4f1b76dd99ad67ea6e4be00a29e370

PR #859, opened 2025-09-11
New stuff on the `FeeRate` type

thunderbiscuit:
This is adding 3 things to the `FeeRate` type.

1. API docs
2. Display trait
3. New methods

The first and 3rd ones are simple and totally cool. The display trait need a bit of discussion maybe.

Note that the new methods are cool/useful because if you have the weight or vbytes for a transaction, you can give this to the `FeeRate` type and get an `Amount` that the transaction will pay in fee. I need to play with some of the advanced parts of the API a bit more, but I think you could basically predict the number of sats a transaction will pay before signing, which is pretty useful.

### The display trait

This one is me sort of freehanding what I think is the better visual display of the feerate, and I understand it's a bit of a subjective thing. Happy to hear what other people think. The [rust-bitcoin FeeRate type](https://docs.rs/bitcoin/latest/bitcoin/struct.FeeRate.html) defines actually [2 "display" versions](https://docs.rs/bitcoin-units/0.1.2/src/bitcoin_units/fee_rate.rs.html#110-119) (this is a fairly obscure Rust feature but I do think it's pretty cool). The standard display is just the number (sat/kwu), which is... kind of weird if you're not used to it. For example 1 sat/vbyte is just `250` if you use the standard display (`println!("Fee rate: {}", feerate)`). If you use the alternative display, however (`println!("Fee rate: {:#}", feerate)`), you get "1.00 sat/vb". This is a better and nicer string, but the "X.00" is hardcoded, and internally you just get the `FeeRate::to_sat_per_vb_ceil` method. This means you have big plateaus of all the same sat/vb rate, even if at a more granual level you're sort of in between (1.09, 1.75, etc.). The decimals are always X.00. Instead, what I think makes the type interesting is its use of the `kwu`, which affords you more granularity. The downside of this is that on the normal print/display, you're not reminded of that and just get a weird and big number. So in this PR... I decided to merge the two Rust implementation and ideas, and add the rate type to the string while keeping the sat/kwu number. This is the result:

```kotlin
val feeRate = FeeRate.fromSatPerVb(3uL)
println("Fee rate: $feeRate")
// 750 sat/kwu
```

### The new methods

This one has a tiny hiccup, more in style than anything, but I want to point it out. There appears to be a bug (as far as me and my friend Claude can tell it's at the uniffi level? that prevents us from returning an `Option<Amount>` on those two methods if I use the standard way to handle this. In theory, these should be very straightforward lines:

```rust
pub fn fee_vb(&self, vb: u64) -> Option<Amount> {
    self.0.fee_vb(vb).map(Amount::from)
}
```

In practice, if you try to do this you'll get compilation errors at the uniffi level. I've tried a whole lot of stuff, and I even put Claude on it for a while, and it just kept running in circles and finding the same problems as me. Some names are clashing in the imports and uniffi thinks returning an `Amount` is the same as a `BdkAmount`, but of course it's not. In the end I can make it work by separating the one liner into a few lines with full type declaration.

### Changelog notice

```md
## Added
    - Add `FeeRate.toString()` method [#859]
    - Add `FeeRate::fee_vb` and `FeeRate::fee_wu` methods [#859]

[#859]: https://github.com/bitcoindevkit/bdk-ffi/pull/859
```

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -30,6 +30,7 @@ use bdk_wallet::bitcoin::Transaction as BdkTransaction;
 use bdk_wallet::bitcoin::TxIn as BdkTxIn;
 use bdk_wallet::bitcoin::TxOut as BdkTxOut;
 use bdk_wallet::bitcoin::Txid as BitcoinTxid;
+use bdk_wallet::bitcoin::Weight;
 use bdk_wallet::bitcoin::Wtxid as BitcoinWtxid;
 use bdk_wallet::miniscript::psbt::PsbtExt;
 use bdk_wallet::serde_json;


bdk-ffi/src/bitcoin.rs
```diff
@@ -134,22 +137,56 @@ impl FeeRate {
...
+    ///
+    /// This is equivalent to converting vb to weight using Weight::from_vb and then calling Self::fee_wu(weight).
+    pub fn fee_vb(&self, vb: u64) -> Option<Arc<Amount>> {
+        let rust_amount: BdkAmount = self.0.fee_vb(vb)?;
+        let amount: Amount = rust_amount.into();
+        Some(Arc::new(amount))
+
+        // The whole code above should be replaceable by the following line:
+        // self.0.fee_vb(vb).map(Arc::new(Amount::from))
+        // But in practice you get uniffi compilation errors on it. Not sure what is going on with it,
```
reez, line 175: how weird, good comment for the future when hopefully it resolves itself

bdk-ffi/src/bitcoin.rs
```diff
@@ -121,10 +122,12 @@ impl HashableOutPoint {
 /// This is an integer type representing fee rate in sat/kwu. It provides protection against mixing
 /// up the types as well as basic formatting features.
 #[derive(Clone, Debug, uniffi::Object)]
+#[uniffi::export(Display)]
```
reez, line 129: nice

bdk-ffi/src/bitcoin.rs
```diff
@@ -134,22 +137,56 @@ impl FeeRate {
...
+
+        // The whole code above should be replaceable by the following line:
+        // self.0.fee_vb(vb).map(Arc::new(Amount::from))
+        // But in practice you get uniffi compilation errors on it. Not sure what is going on with it,
+        // but the code we use works just as well.
+    }
+
+    /// Calculates fee by multiplying this fee rate by weight, in weight units, returning `None` if overflow occurred.
+    //
+    // This is equivalent to Self::checked_mul_by_weight().
```
reez: `///` instead of `//` for these two lines

bdk-ffi/src/bitcoin.rs
```diff
@@ -134,22 +137,56 @@ impl FeeRate {
...
+        let weight: Weight = Weight::from_wu(wu);
+        let rust_amount: BdkAmount = self.0.fee_wu(weight)?;
+        let amount: Amount = rust_amount.into();
+        Some(Arc::new(amount))
+    }
+}
+
+impl Display for FeeRate {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{} sat/kwu", self.0)
```
reez: what do you think of using "{:#}" because I think it forwards both the number and the unit decision to the FeeRate type from bitcoin-units (instead of hardcoding sat/kwu)? For now I think we get the same output with both, but maybe in the future if something weird changes it'll keep things aligned with the upstream.

So the idea is replacing `write!(f, "{} sat/kwu", self.0)` with `write!(f, "{:#}", self.0)`

thunderbiscuit: Ok going with this approach!

thunderbiscuit:
Sorry for the delay. Ready for review now @reez.

reez (approved):
ACK 8c62796bdb3c448d8d2a52d7e6fc9e77e1c05a0a

PR #861, opened 2025-09-17
Move Rust tests into their own separate module

thunderbiscuit:
This PR moves all tests currently in source files into their own `tests/` directory. This keeps the source files clean, and makes it hopefully more inviting to write and maintain tests for the API.

Just doing this exercise outlined for me a bunch of low hanging fruits that could be tested right away, so I take that as a win.

One of the new clippy warning for 1.85.0 is that you define the `Default` trait for types that have a `new()` constructor without arguments. I can do that, but it's not useful to have in our codebase because those never get used, and are not translatable to the language bindings. It's therefore adding noise/lines that don't get used anywhere, only adding confusion as to why we implemented the trait IMO. The last commit on this PR disables that rules for the 2 types for which this applies, hence allowing the clippy workflow to pass.


bdk-ffi/src/lib.rs
```diff
@@ -1,15 +1,15 @@
...
+pub mod electrum;
+pub mod error;
+pub mod esplora;
+pub mod keys;
+pub mod kyoto;
+pub mod macros;
+pub mod store;
+pub mod tx_builder;
+pub mod types;
+pub mod wallet;
```
reez: Is there a way to keep the modules private and instead expose just what the new integration tests need with a cfg(test) re-export or similar

thunderbiscuit: I can expand on this on a live call just to make sure it's clear, but in short, the issue was that previously our _entire_ codebase was private haha! This is a little odd for a library. We basically had _no_ visible types. The way the `tests` directory works is it imports your library just like it would any other dependency. This is apparently so it can really test your public API and not your internal functions users don't have access to. 

In doing this, I realized none of our code was available! This sounds odd, but it works because we don't have any users of bdk-ffi as a library, and uniffi is sort of our only "user". On top of that, uniffi completely disregards the public/private visibility modifiers on the Rust code. It takes what you export and makes it available to bindings code.

This setup is the reason why I had to bring back the "rlib" type in the crate types: it's only there for the tests directory. Once that's done, the library actually exposed none of its API. Me adding those modules as public makes them accessible for the tests, and because they are all at their heart "public-facing" modules, AKA modules that are intended for users, I put them all public here.

Note that not all of those are _actually_ required for this PR; only the modules that are being tested. But adding a lot of cfg test stuff everywhere I think would more confuse everything, when in fact it's all meant for public consuption (even thought we don't have actual users). Does that make sense? Happy to chat about this more over the dev call too.

reez: I think the question things are hingeing on for me is if we are definitely wanting everything to be public/stable; because I think switching these to pub mod makes every existing pub item under those modules part of the crate’s public API (which I think would mean we couldn't rename types or reorganize modules without a breaking change, etc?)... could totally be wrong in how I'm conceiving of it, and I'm not necessarily for/against, mores it's just the main thing I'm thinking thru.

thunderbiscuit: I'm able to sidestep the issue by adding the tests in a new `tests` module! This is apparently common, so I'm ok with this solution if you are.

reez: yeah at first glance I think this sidesteps that issue nicely! good solution.

let me know if this pr has any more changes coming or not, when I know its finalized again I'll do another close review again

thunderbiscuit: Nothing more in this pr. Ready for final review.

PR #864, opened 2025-09-24
Bump `bdk_kyoto` version

rustaceanrob:
Work in progress draft PR. Copied from my commit message:

The CBF crate, renamed on crates.io from `kyoto` to `bip157`, includes some performance improvements and dependency reduction.

Starting with a relatively simple change, the configuration of a DNS resolver has been removed, as DNS is now carried out by the operating system. In testing this has resolved problems for DNS queries while using a VPN or on public WiFi (with firewalls).

Now the CBF node does no disk I/O over `sqlite`. This removes a number of warings and errors related to persistance. All chain data is stored within the `Wallet` structure and subsequently persisted to `bdk_chain`'s extension for `sqlite`. This creates a single source-of-truth for persistance between sessions, which has been a huge pain point in the past. The logic of `CheckPoint::insert` handles competing forks. The builder has no error cases at the FFI level with this change.

The `ScanType` has been reduced to two options, `Scan` or `Recovery`. The `Recovery` variant contains a few notable checkpoints to start from. It also contains a field that the user may set to peek scripts while scanning filters. This should be approximately the number of scripts revealed by the wallet being recovered. The guidance for developers here would be to set conservatively. One scheme is setting the peek to 1_000, 10_000, 100_000, depending on profiles: normal, power user, enterprise.

The `Info` variants have been reduced to a single `Progress` containing all info related to the sync.

Broadcasting a transaction now waits for the confirmation of gossip, as in at least one remote node asked for the transaction be shared with them.

`Node::run` is now mutable. This implies `CbfNode.run()` can only be called once. I think that is perferctly reasonable as any other type of behavior is faulty.

`bitcoin-address-book` is a simple crate I created to abstract peer selection, which `bip157` depends on.

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/src/kyoto.rs
+++ b/bdk-ffi/src/kyoto.rs
@@ -1,19 +1,19 @@
-use bdk_kyoto::builder::NodeBuilder as BDKCbfBuilder;
-use bdk_kyoto::builder::NodeBuilderExt;
-use bdk_kyoto::kyoto::lookup_host;
-use bdk_kyoto::kyoto::tokio;
-use bdk_kyoto::kyoto::AddrV2;
-use bdk_kyoto::kyoto::ScriptBuf;
-use bdk_kyoto::kyoto::ServiceFlags;
+use bdk_kyoto::bip157::lookup_host;
+use bdk_kyoto::bip157::tokio;
+use bdk_kyoto::bip157::AddrV2;
+use bdk_kyoto::bip157::Network;
+use bdk_kyoto::bip157::Node;
+use bdk_kyoto::bip157::ServiceFlags;
+use bdk_kyoto::builder::Builder as BDKCbfBuilder;
+use bdk_kyoto::builder::BuilderExt;
+use bdk_kyoto::HeaderCheckpoint;
 use bdk_kyoto::LightClient as BDKLightClient;
-use bdk_kyoto::NodeDefault;
 use bdk_kyoto::Receiver;
 use bdk_kyoto::RejectReason;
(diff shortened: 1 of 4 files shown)


bdk-ffi/src/kyoto.rs
```diff
@@ -352,7 +349,7 @@ impl CbfClient {
             .enable_all()
             .build()
             .unwrap()
-            .block_on(lookup_host(hostname, self.dns_resolver));
+            .block_on(lookup_host(hostname));
```
reez, line 352: if blocking here should we remove the `async` for the function?

I know it was already like this but it just caught my eye here on review

rustaceanrob: Yes we should, good call.

thunderbiscuit (approved):
ACK e428b03daca69ed11e600472a2e63f66c1cfddfa.

ItoroD:
@rustaceanrob For `ScanType`, since New is removed I believe `Sync` will just handle for when the wallet is newly setup as well. As in, although I do not technically have a last checkpoint because I am new wallet, it still downloads as if starting point was the last check point?

ItoroD:
> It will actually start from genesis. This was partially due to developers not using the option in practice, and the checkpoints being a trust-vector and maintenance burden. For wallets that are known to be new, they may recover from taproot.
> 
> Another strategy is they can ping mempool.space for a recent checkpoint and add it to the wallet.

Alright, thanks for the explanation.

PR #871, opened 2025-10-03
Bump bdk_wallet to 2.2.0

thunderbiscuit:
This is a fairly simple bump. Note that we should merge #864 first.

### Changelog notice

```md
## Changed
    - Update bdk_wallet to 2.2.0 [#871]

[#871]: https://github.com/bitcoindevkit/bdk-ffi/pull/871
```

#### All Submissions:

### DIFF
--- a/bdk-ffi/src/error.rs
+++ b/bdk-ffi/src/error.rs
@@ -22,6 +22,7 @@ use bdk_wallet::error::CreateTxError as BdkCreateTxError;
 use bdk_wallet::keys::bip39::Error as BdkBip39Error;
 use bdk_wallet::miniscript::descriptor::DescriptorKeyParseError as BdkDescriptorKeyParseError;
 use bdk_wallet::miniscript::psbt::Error as BdkPsbtFinalizeError;
+#[allow(deprecated)]
 use bdk_wallet::signer::SignerError as BdkSignerError;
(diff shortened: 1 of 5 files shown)


thunderbiscuit:
Mozilla has also updated their `thiserror` dependency to 2.X, so I'm also taking the opportunity to bump our dependency here.

bdk-ffi/src/error.rs
```diff
@@ -1509,6 +1510,7 @@ impl From<BdkPsbtFinalizeError> for PsbtFinalizeError {
     }
 }
 
+#[allow(deprecated)]
```
reez, line 1507: interesting, hadn't seen/noticed these before

reez (approved):
ACK 4fdb5ed5ca06c0dc5d5b4180e9cf1553cc60189d

Sounds good waiting for kyoto pr merge

bdk-ffi/src/error.rs
thunderbiscuit: Oh yeah I forgot to add a comment about those! The warnings were added to 2.2 as part of the deprecation of the signing capabilities of the wallet. With 3.0 the wallet will only handle public keys.

reez: pretty cool!

PR #874, opened 2025-10-07
Update readme for 2.2.0 release

thunderbiscuit:
This is a big cleanup of the root readme.

I think it covers it all but let me know what you think. I tried to simplify it and keep the details in the more specific readmes.


reez (approved):
ACK 27ada0d2fa8ac5b0ab164a0ff5424dc49f357b72

Nice and needed 👍

PR #883, opened 2025-10-14
deps: uniffi 0.30.0

reez:
### Description

Testing uniffi 0.30.0

Two changes:
0. bumping to 0.30.0
1. export Ord trait so swift and kotlin now treat them as comparable value types

### Notes to the reviewers

Tested out this other thing along the way but didn't work
2. ~(breaking! but potentially awesome) convert HashableOutPoint from an object to a record ["Enums and Records can have methods"](https://github.com/mozilla/uniffi-rs/blob/main/CHANGELOG.md#%EF%B8%8F-breaking-changes-for-external-bindings-authors-%EF%B8%8F-1)~ compiles but exporting our own methods on a record fails in bindgen step

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -983,37 +983,37 @@ impl From<TxOut> for BdkTxOut {
 }
 
 /// A bitcoin Block hash
-#[derive(Debug, Clone, Copy, PartialEq, Eq, std::hash::Hash, uniffi::Object)]
-#[uniffi::export(Display, Eq, Hash)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, std::hash::Hash, uniffi::Object)]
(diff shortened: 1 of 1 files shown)


bdk-ffi/src/bitcoin.rs
```diff
@@ -98,21 +98,23 @@ pub enum Network {
 ///
 /// Due to limitations in generating the foreign language bindings, we cannot use [`OutPoint`] as a
 /// key for hash maps.
-#[derive(Debug, PartialEq, Eq, std::hash::Hash, uniffi::Object)]
-#[uniffi::export(Debug, Eq, Hash)]
-pub struct HashableOutPoint(pub(crate) OutPoint);
+#[derive(Debug, Clone, PartialEq, Eq, std::hash::Hash, uniffi::Record)]
```
reez: note: once we switch to a record I dont think we have to hang trait exports on the struct itself anymore

Issue #875, opened 2025-10-07
Changelog for Next Release (2.2)


thunderbiscuit:
Building the changelog for new releases when they come out is always a bit of a pain. Instead of going back into all PRs that were merged since the last release (and having to figure out which ones were on which release), I would like to see if we can instead just add them to this issue.

I'll edit this issue with changelog entries when we merge PRs, and when it's time to release we should be able to just copy/paste what we have here.

thunderbiscuit:
Done!

PR #893, opened 2025-10-28
Add DerivationPath helper methods

ItoroD:
### Notes to the reviewers
Added more useful methods for `DerivationPath` for now I added the simplest ones. More should be added later; like child, into_child.

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -3,6 +3,8 @@ use crate::error::{
     PsbtParseError, TransactionError,
 };
 use crate::error::{ParseAmountError, PsbtFinalizeError};
+use crate::keys::DerivationPath;
+
 use crate::{impl_from_core_type, impl_hash_like, impl_into_core_type};
 use std::collections::HashMap;
 
@@ -510,7 +512,7 @@ pub struct TapScriptEntry {
     pub leaf_version: u8,
 }
 
-#[derive(Clone, Debug, uniffi::Record, Hash, Eq, PartialEq)]
+#[derive(Clone, Debug, uniffi::Record)]
 pub struct TapKeyOrigin {
     /// leaf hashes as hex strings
     pub tap_leaf_hashes: Vec<String>,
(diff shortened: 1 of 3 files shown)


reez:
I dig this addition. Are you going to add `#[uniffi::export(Display)]` to `DerivationPath`?

ItoroD:
> I dig this addition. Are you going to add `#[uniffi::export(Display)]` to `DerivationPath`?

Good call.

bdk-ffi/src/bitcoin.rs
```diff
@@ -723,7 +725,7 @@ impl From<&BdkInput> for Input {
                         tap_leaf_hashes: v.0.iter().map(|h| h.to_string()).collect(),
                         key_source: KeySource {
                             fingerprint: v.1 .0.to_string(),
-                            path: v.1 .1.to_string(),
+                            path: Arc::new(v.1 .1.clone().into()),
```
thunderbiscuit: The spaces here I thought would have been caught by the cargo formatter, but they weren't. Those lines could be cleaned up as:

```rust
fingerprint: v.1.0.to_string(),
path: Arc::new(v.1.1.clone().into()),
```

thunderbiscuit: Man those deeply nested types are quite a handful 😆😆😆

bdk-ffi/src/keys.rs
```diff
@@ -64,7 +65,7 @@ impl Display for Mnemonic {
 }
 
 /// A BIP-32 derivation path.
-#[derive(uniffi::Object)]
+#[derive(Clone, Debug, Hash, Eq, PartialEq, uniffi::Object)]
```
thunderbiscuit: I don't think they hurt much, but in general I'm wary of adding trait implementations that are not needed. Note also that those are not exported to the bindings, so they'd need to be useful at the Rust layer, but I don't know that they are for now (the Clone is, but not the Eq, PartialEq, and Hash).

bdk-ffi/src/keys.rs
```diff
@@ -64,7 +65,7 @@ impl Display for Mnemonic {
...
 pub struct DerivationPath(pub(crate) BdkDerivationPath);
```
thunderbiscuit, line 70: I think you're missing the Display trait export here:

```rust
#[derive(Debug)]
#[derive(uniffi::Object)]
#[uniffi::export(Display)]
pub struct DerivationPath(pub(crate) BdkDerivationPath);
```

thunderbiscuit: You have the Display implemented below, but bindings users won't get it unless we add the `#[uniffi::export(Display)]` line.

bdk-ffi/src/keys.rs
thunderbiscuit: After playing with it locally, the Debug trait is also used (as well as the Clone). Others can be dropped on this one and others that have them.

thunderbiscuit (changes requested):
Just a few cleanup ideas.

bdk-ffi/src/keys.rs
ItoroD: Again! I thought I added that. I really do need to "Shine my eye". Maybe lost on rebase and merge resolution. Thanks

bdk-ffi/src/bitcoin.rs
ItoroD: Cargo fmt is actually the one adding those spaces

bdk-ffi/src/keys.rs
ItoroD: good call

Issue #869, opened 2025-10-02
bdk_wallet 2.3 or 3.0


reez:
This issue is to track items that we want to implement or test in the next bdk_wallet release (likely 3.0 but maybe 2.3):
- remove temp fix https://github.com/bitcoindevkit/bdk-ffi/pull/863#pullrequestreview-3296026846 if upstream bdk_esplora adds fixes like https://github.com/bitcoindevkit/bdk/pull/2053#event-20046371201

reez:
> [bitcoindevkit/bdk_wallet#388](https://github.com/bitcoindevkit/bdk_wallet/pull/388)
> 
> expose apply_unconfirmed_txs_events and apply_evicted_txs_events
> 
> event::wallet_events may be transitional until CanonicalView::diff so not sure if we should defer it or not based on how quickly that comes

Landed upstream in `bdk_wallet` and ships in `3.0.0-rc.1` so I exposed these in my rc1 branch. This only wraps the public wallet method and existing `WalletEvent` mapping in bdk-ffi without exposing lower-level event internals that may still change later.

reez:
**rc1**

- [x] `Wallet::apply_unconfirmed_txs_events`
- [x] `Wallet::apply_evicted_txs_events`
- [x] `Wallet::checkpoints()`
- [x] `Wallet::start_sync_with_revealed_spks_at`
- [x] `Wallet::start_full_scan_at`
- [x] `TxBuilder::sighash`
- [x] `BumpFeeTxBuilder::sighash`

Notes:

`TxBuilder::sighash / BumpFeeTxBuilder::sighash`
Upstream explicitly warns "Use this option very carefully" in bdk_wallet. I exposed it for parity since bdk-ffi already exposes PSBT input sighash_type and signing-related sighash controls, but I’d be fine dropping it if we want to keep the FFI surface more conservative.

reez:
Decided to just implement/expose everything, not because we will necessarily merge, but so we can give feedback upstream if needed based on shape of apis as it relates to implementing these new things in bindings...

reez:
**rc1**

- [x] `TxBuilder::add_foreign_utxo_with_sequence`
- [x] `TxBuilder::ordering / BumpFeeTxBuilder::ordering`
- [x] `ChangeSet.locked_outpoints`
- [x] `Wallet::list_locked_outpoints / list_locked_unspent / is_outpoint_locked / lock_outpoint / unlock_outpoint`

Notes:

[...]

PR #897, opened 2025-10-31
Enable loading of single-descriptor wallets

### LINKED ISSUE #860: Allow `Wallet::load` from single descriptor wallets
Right now this doesn't work as the load constructor requires 2 descriptors.

thunderbiscuit:
Fixes #860.

Note that the cleaner way to do this would be to make the change descriptor an option on the `Wallet::load` constructor, but that would be a breaking change. We can clean it up on the 3.0 release, unless the multi-keychain support makes it in, in which case all of this will be much cleaner but in a different way.

### Documentation

There are no direct equivalent on the Rust side.

- [ ] Other: 

#### All Submissions:

#### New Features:


thunderbiscuit:
Note to reviewers: I will open a new PR cleaning up the constants file and using the easier to work with [Constant.kt](https://github.com/bitcoindevkit/bdk-jvm/blob/master/lib/src/test/kotlin/org/bitcoindevkit/Constants.kt). I just didn't want to drown this PR in test suite changes. Let me know if you prefer that, I can easily add the commits to fix the tests here as well.

As with the bdk-jvm test suite, I have added a README file in the resources that explain the different wallet databases and their features. It's easier to use them for different tests when you know what's in them, and also makes it clear when you need a new one because you want to tests something not covered in the currently available ones.

reez (approved):
ACK 8cbf31ee5f2b5aa3ab47e3a814994d2b65c76f25

This looks great to me. I added a bulletpoint to https://github.com/bitcoindevkit/bdk-ffi/issues/869#issue-3478727244 about this for 3.0 hopefully that's cool

reez:
> Awesome yes that's a good place to add it to make sure we don't forget.

cool nice work on this

PR #902, opened 2025-11-04
fix(kyoto): Spawn dedicated thread for DNS

rustaceanrob:
In personal testing and also reported to me, DNS appears to be very, very slow with the current setup. This function uses a `new_current_thread` runtime executor, but what I did not realize is this might be the main application thread. Spawning a detached thread here, same as is done with the CBF node, should add an incremental speed up. The main thread can continue to do whatever work it is supposed to do without pausing the DNS queries.


rustaceanrob:
c9f11de2b289d4bee9d5f880b18e0527b100b5ee rebased

PR #908, opened 2025-11-14
feat: Add wallet event

ItoroD:
### Description

Adds wallet event to bdk-ffi types and exposes `apply_update_events`
### Notes to the reviewers
- Since [WalletEvent ](https://docs.rs/bdk_wallet/latest/bdk_wallet/event/enum.WalletEvent.html) is non-exhaustive I used wildcard with the unreachable! macro. I was thinking maybe there are better options (panic!() or unimplemented!). Not sure of most ideal for rust
- Event looks great too. I tried this out in our dev-kit wallet. See what it looks like here from 2 different even printout
[event-logs.txt](https://github.com/user-attachments/files/23547594/event-logs.txt)

### Documentation

- [WalletEvent](https://docs.rs/bdk_wallet/latest/bdk_wallet/event/enum.WalletEvent.html)

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/src/types.rs
+++ b/bdk-ffi/src/types.rs
@@ -23,6 +23,7 @@ use bdk_wallet::descriptor::policy::{
     Condition as BdkCondition, PkOrF as BdkPkOrF, Policy as BdkPolicy,
     Satisfaction as BdkSatisfaction, SatisfiableItem as BdkSatisfiableItem,
 };
+use bdk_wallet::event::WalletEvent as BdkWalletEvent;
 #[allow(deprecated)]
 use bdk_wallet::signer::{SignOptions as BdkSignOptions, TapLeavesOptions};
 use bdk_wallet::AddressInfo as BdkAddressInfo;
@@ -32,6 +33,7 @@ use bdk_wallet::Update as BdkUpdate;
 
 use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::convert::TryFrom;
+use std::fmt::Display;
 use std::sync::{Arc, Mutex};
 
 use crate::{impl_from_core_type, impl_into_core_type};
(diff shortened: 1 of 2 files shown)


bdk-ffi/src/types.rs
```diff
@@ -498,6 +499,143 @@ impl From<BdkSatisfiableItem> for SatisfiableItem {
...
+    ///
+    /// The conflicts field contains the txid and vin (in which it conflicts) of the conflicting
+    /// transactions.
+    TxReplaced {
+        /// Transaction id.
+        txid: Arc<Txid>,
+        /// Transaction.
+        tx: Arc<Transaction>,
+        /// Conflicting transactions.
+        conflicts: Vec<ConflictingTx>,
```
thunderbiscuit: I'd call this new intermediate type `Conflict`. It's not just a transaction conflict, it's more of a specific input in a specific transaction. Just a small nit. 

So you'd get a `Vec<Conflict>`, where a `Conflict` is the Txid with the input number that's conflicting.

bdk-ffi/src/types.rs
```diff
@@ -498,6 +499,143 @@ impl From<BdkSatisfiableItem> for SatisfiableItem {
...
+        /// Transaction.
+        tx: Arc<Transaction>,
+    },
+}
+
+/// Represents a conflicting transaction in a replacement transaction.
+#[derive(Debug, Clone, uniffi::Record)]
+pub struct ConflictingTx {
+    /// Position of the conflicting transaction.
+    pub position: u32,
```
thunderbiscuit: If you rename this type `Conflict`, your docstring could be `/// Represent a conflict in a replacement transaction.`

The name of this field I think should be `vin`, and the docstring should be something like `/// The index of the conflicting input.`

ItoroD: Thanks. Will make the necessary updates.

thunderbiscuit (approved):
ACK aabeb14c838e8e35008de23c166c5f9f8ffd0c29.

PR #920, opened 2025-12-02
feat(kyoto): Add optional new wallet sync height

rustaceanrob:
For users recovering an old wallet, it is best to used the pre-defined points like segwit or taproot; however, new users that have definitely not used their descriptor may be caught off guard if they have to wait for their wallet to sync for more than a couple seconds. This change allows for a recovery from any hash and height, with the idea being the developer can provide a recent block via a service like mempool(dot)space

@andreasgriffin 

#### All Submissions:

#### New Features:


thunderbiscuit (approved):
ACK 677a60e3667a55bd5613db9b6407f43116f3036f.

PR #924, opened 2025-12-13
add lib crate-type to bdk-ffi to be able to reuse it for other bindings like bdk-dart

kumulynja:
### Description

This PR just adds "lib" as a crate-type besides "cdylib" and "staticlib". In this way bdk-ffi can be more easily wrapped and reused for other bindings like the dart bindings in bdk-dart. Without it, more code from bdk-ffi has to be duplicated to bdk-dart and the process becomes more complex and hacky.

### Notes to the reviewers

I tested this PR in the following bdk-dart PR already: https://github.com/bitcoindevkit/bdk-dart/pull/12
It would be good to get this merged here first to be able to remove the patch to my fork of bdk-ffi there.

#### All Submissions:

### DIFF
--- a/bdk-ffi/Cargo.toml
+++ b/bdk-ffi/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2018"
 license = "MIT OR Apache-2.0"
 
 [lib]
-crate-type = ["cdylib", "staticlib"]
+crate-type = ["lib", "cdylib", "staticlib"]
 name = "bdkffi"
 
 [[bin]]


bdk-ffi/Cargo.toml
```diff
@@ -7,20 +7,30 @@ edition = "2018"
 license = "MIT OR Apache-2.0"
 
 [lib]
-crate-type = ["cdylib", "staticlib"]
+crate-type = ["lib", "cdylib", "staticlib"]
```
thunderbiscuit, line 10: This is the only line that needs to be on this PR.

kumulynja:
> I'm not against this change per se, but I'd like to understand better why the plugin for dart needs it when others don't (JVM, Android, Swift, Python, React Native).
> 
> I took some time to clean this up a while back and my understanding was that the `lib` type is required in situations where your Rust code is used as a dependency itself, whereas for us here it's the compiled artifacts that are used by the plugins.
> 
> Again I'm not against it because it doesn't impact us negatively other than build times, but I want to make sure I know why each of those crate types are required (`cdylib` is for Android/JVM/Python, `staticlib` is for iOS).

It's because Dart has a new build system that does the compiling of the artifacts/native code dynamically when a user imports the `bdk-dart` package. So as maintainers of `bdk-dart`, there is no need anymore to distribute pre-built native libs upfront. This greatly reduces the maintenance burden for `bdk-dart` since packaging/distribution/bundling of the native libs are a real pain otherwise. 

The only requirement is that the native/foreign code (Rust in the case of `bdk`) is part of the Dart project itself so it can be published together with the Dart package. Since `bdk-ffi` and `bdk-dart` are separate repositories, this requirement is not met, but we added a very small Rust project in `bdk-dart` that just exports `bdk-ffi` to circumvent this. To be able to do that, `bdk-ffi` needs to be used as a dependency in this small wrapper as you can see here: https://github.com/kumulynja/bdk-dart/tree/2c5cd9ccfd647d2695e52150a7670730b28020b7/native

Here is some more context about Native Assets in Dart and it being the new standard and way to go for bindings:

> Build hooks 🚀
> Integrating native code (like C++, Rust, or Swift) into a Dart package has often required managing complex, platform-specific build files like CMake or Gradle. With Dart 3.10, this process is now dramatically simpler.
> 
> We’re excited to announce that build hooks (formerly known as native assets) are now stable. You can use these hooks to compile native code or download native assets (like dynamic libraries) and bundle them directly with your Dart package. This powerful capability lets you reuse existing native code or libraries from within your package and eliminates the need to write separate build files such as SPM, Gradle, or CMake for different operating systems.

source: https://blog.dart.dev/announcing-dart-3-10-ea8b952b6088

bdk-ffi/Cargo.toml
kumulynja: This is solved now by rolling back the formatting of the other lines.

reez (approved):
ACK ea7be23764060ad5bdf3915dd550c6f400d9f923

Thought a good bit about this, I think because it's such a minimal change and is a big unblock for `bdk-dart` im in favor. I know it slows down build times, but that as well was minimal in my testing: dev builds ~+0.66s, release builds ~+1.03, release-smaller negligible.

The other alternatives are to do nothing or a much larger refactor, so this PR seemed like a clear winner in my mind of all the options.

Open to anyone else's feedback or if I missed anything

thunderbiscuit (approved):
ACK a3d83653c39e5a29117d0fa6b4051d3971fabb6d.

PR #928, opened 2025-12-30
feat: expose only_witness_utxo and add_foreign_utxo on TxBuilder

kumulynja:
### Description

This PR adds the `only_witness_utxo` and `add_foreign_utxo` on `TxBuilder` since they are available in `bdk`, but missing in `bdk-ffi`. They are useful for bdk bindings to be used together with for example bark (Ark implementation from Second Tech) so a unilateral exit can be done with Cpfp transactions.

### Notes to the reviewers

Since the `add_foreign_utxo` function is still experimental, I added the same comments mentioning this and the caution that has to be taken. I think it is good to expose it already though for experimentation like with bark as mentioned before.

Both functions seem to be working in [an example in Dart to unilaterally exit from bark](https://gitlab.com/ark-bitcoin/bark-ffi/-/blob/master/bindings/dart/example/custom_onchain_example.dart?ref_type=heads#L210) through a fork of both bdk-ffi and bdk-dart.

### Documentation

- [x] `bdk-ffi`

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -20,8 +20,6 @@ use bdk_wallet::bitcoin::io::Cursor;
 use bdk_wallet::bitcoin::psbt::Input as BdkInput;
 use bdk_wallet::bitcoin::psbt::Output as BdkOutput;
 use bdk_wallet::bitcoin::secp256k1::Secp256k1;
-use std::collections::HashMap;
-use std::convert::TryFrom;
 
 use bdk_wallet::bitcoin::bip32::ChildNumber as BdkChildNumber;
 use bdk_wallet::bitcoin::taproot::LeafNode as BdkLeafNode;
@@ -42,6 +40,8 @@ use bdk_wallet::bitcoin::Wtxid as BitcoinWtxid;
 use bdk_wallet::miniscript::psbt::PsbtExt;
 use bdk_wallet::serde_json;
 
+use std::collections::HashMap;
+use std::convert::TryFrom;
 use std::fmt::Display;
 use std::fs::File;
 use std::io::{BufReader, BufWriter};
@@ -692,6 +692,8 @@ pub struct Input {
     pub unknown: HashMap<Key, Vec<u8>>,
 }
 
+use crate::error::AddForeignUtxoError;
+
 impl From<&BdkInput> for Input {
     fn from(input: &BdkInput) -> Self {
         Input {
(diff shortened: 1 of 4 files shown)


bdk-ffi/src/tx_builder.rs
```diff
@@ -366,6 +371,73 @@ impl TxBuilder {
...
+    /// Returns an error if the `outpoint` doesn't exist or the `satisfaction_weight` is not provided.
+    ///
+    /// [`max_weight_to_satisfy`]: https://docs.rs/miniscript/latest/miniscript/trait.Descriptor.html#method.max_weight_to_satisfy
+    pub fn add_foreign_utxo(
+        &self,
+        outpoint: OutPoint,
+        psbt_input: Input,
+        satisfaction_weight: u64,
+    ) -> Result<Arc<Self>, AddForeignUtxoError> {
+        let bdk_outpoint: BdkOutPoint = outpoint.into();
```
reez, line 450: should we guard here if psbt_input lacks both witness_utxo and non_witness_utxo

```
if psbt_input.witness_utxo.is_none() && psbt_input.non_witness_utxo.is_none() {
  return Err(AddForeignUtxoError::MissingUtxo);
}
```

would match upstream behavior at the method boundary instead of deferring failure to finish()

https://docs.rs/bdk_wallet/latest/bdk_wallet/tx_builder/struct.TxBuilder.html#method.add_foreign_utxo

"This method returns errors… The psbt_input does not contain a witness_utxo or non_witness_utxo."

bdk-ffi/src/tx_builder.rs
```diff
@@ -366,6 +371,73 @@ impl TxBuilder {
...
+    /// This is automatically set when working with SegWit transactions. The `witness_utxo` field
+    /// is enough to construct the signature hash.
+    pub fn only_witness_utxo(&self) -> Arc<Self> {
+        Arc::new(TxBuilder {
+            only_witness_utxo: true,
+            ..self.clone()
+        })
+    }
+
+    /// Add a foreign UTXO i.e. a UTXO not owned by this wallet.
```
reez: Docs here don’t look identical to upstream add_foreign_utxo any reason for that?

bdk-ffi/src/tx_builder.rs
```diff
@@ -366,6 +371,73 @@ impl TxBuilder {
         })
     }
 
+    /// Only fill witness_utxo.
```
reez: Docs here don’t look identical to upstream only_witness_utxo any reason for that?

bdk-ffi/src/tests/tx_builder.rs
```diff
@@ -86,3 +88,459 @@ fn create_and_sync_wallet() -> Wallet {
...
+                panic!("Expected transaction with only_witness_utxo to succeed");
+            }
+        }
+    } else {
+        println!("Failed to retrieve policies, skipping transaction test");
+    }
+}
+
+#[test]
+fn test_add_foreign_utxo_missing_witness_data() {
```
reez, line 200: if we add the guard in add_foreign_utxo (per my code comment in tx_builder.rs) we should update the test for that too

bdk-ffi/src/tx_builder.rs
```diff
@@ -366,6 +371,73 @@ impl TxBuilder {
...
+    ///
+    /// Only fill-in the `psbt::Input::witness_utxo` field when spending from SegWit descriptors.
+    /// This reduces the size of the PSBT. It is valid to fill in both `witness_utxo` and `non_witness_utxo`,
+    /// but the PSBT will be larger.
+    ///
+    /// This is automatically set when working with SegWit transactions. The `witness_utxo` field
+    /// is enough to construct the signature hash.
```
thunderbiscuit: I'm not sure how true this statement is, or at least it's not in the original docs, and your default value on the field is false. I'd remove this line for now. Was there a reason you added it?

bdk-ffi/src/bitcoin.rs
```diff
@@ -809,6 +812,338 @@ impl From<&BdkInput> for Input {
     }
 }
 
+impl TryFrom<Input> for BdkInput {
```
thunderbiscuit, line 837: I don't know if implementing the `TryFrom` trait here is required since we already have the `From` trait implemented. Is there a reason to use the `psbt_input.try_into()?;` instead of just `.into()` and returning the error on the TxBuilder if the conversion fails?

bdk-ffi/src/tx_builder.rs
kumulynja: I can add it if that's the recommended/preferred way. But it means repetition of checks between the original bdk code and bdk-ffi, meaning that it are also two places to maintain if changed. While it gives the same outcome and behavior for the end user if we just let it fail from the original code. Let me know if you would prefer it though and I'll add it.

bdk-ffi/src/bitcoin.rs
kumulynja: I only found: `From<&BdkInput> for Input`, not the other way around. That's why I did this. Let me know if I missed the `From` to use or if there is a better way to do this.

thunderbiscuit: No that's my bad I didn't realize you were going the other way sorry. All good!

bdk-ffi/src/tx_builder.rs
thunderbiscuit: I think it would make sense to match the bdk workflow, but in this case we also know that:
1. This TxBuilder workflow will be completely replaced by the new `Wallet::create_psbt` method in 3.0
2. I don't think there are many users of this method other than you @kumulynja
3. We can fix this later on in a patch or minor release since it would be a non-breaking change

I'm ok with this for now.

thunderbiscuit (approved):
ACK 83b203087cf193140b0e4265a124d2df2350a542

bdk-ffi/src/tx_builder.rs
reez: is the decision on this to:
- not address if our next release is 3.0
- track & address if our next release is 2.x

thunderbiscuit: I think yes that would make sense. What's your take on it?

PR #933, opened 2026-01-08
Use FeeRate type in TxDetails fee_rate field

### LINKED ISSUE #932: Return `FeeRate` type in TxDetail instead of `FeeRate::to_sat_per_vb_ceil()` (f32)
See [code that performs this here](https://github.com/bitcoindevkit/bdk-ffi/blob/a3d83653c39e5a29117d0fa6b4051d3971fabb6d/bdk-ffi/src/types.rs#L1327).

This currently means users cannot compute an accurate to any number of decimal points the FeeRate when returned by the `TxDetails` struct.

thunderbiscuit:
Fixes #932.

The current implementation doesn't allow to calculate more precise versions of the feerate and always gives back the ceiling version of the feerate integer in sat/vb. This lets users do the calculations they need on the type.

It's too bad but this is actually a breaking change, so can only ship with 3.0.

### Documentation

https://docs.rs/bdk_wallet/2.3.0/bdk_wallet/struct.TxDetails.html

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-ffi/Cargo.lock
+++ b/bdk-ffi/Cargo.lock
@@ -16,15 +16,15 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.13"
+version = "1.0.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5192cca8006f1fd4f7237516f40fa183bb07f8fbdfedaa0036de5ea9b0b45e78"
+checksum = "940b3a0ca603d1eade50a4846a2afffd5ef57a9feac2c0e2ec2e14f9ead76000"
 
 [[package]]
 name = "anyhow"
-version = "1.0.101"
+version = "1.0.102"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


busayo-OD:
> @busayo-OD
> 
> > Do you expect consumers to perform the sat/kwu → sat/vB conversion themselves, or would it make sense to expose a helper that returns sat/vB as a float to make fractional fee rates directly accessible?
> 
> In theory I do like this and wish that rust-bitcoin had more utilities to work with sats/vb, but the design of the type was chosen because they wanted to circumvent dealing with floating-point arithmetic completely (pushing that complexity onto the user instead of within the `FeeRate` type). In this library we only expose what's available in the Rust equivalent libraries we support, and so in this case I don't want to add a custom wrapper.
> 
> @ItoroD thanks for the review! Just FYI I think those 3.0 PRs will sit here for a while before we start merging breaking changes on `master`. But it's good to know they're reviewed and ready to go!

Makes sense to keep this aligned with the Rust APIs and avoid floating-point handling. Thanks for the explanation.

PR #935, opened 2026-01-12
feat: Add derivation path child and extend methods

ItoroD:
### Description
Added more methods for the derivation path type. Below are 4 convenient methods I have chosen to leave out and why.  

`into_child` : This does the same thing `child` does except that one consumes self and the other clones. This will not make a difference on bindings level.

`children_from` ,`normal_children` and `hardened_children`
1. The methods below return iterators and if we try to collect them in a vec like we often do for other methods, application crashes as it tries to collect for all possible indexes of a u32. 
2. When dealing with potentially large or infinite sequences (like derivation paths), returning a pre-allocated Vec is inefficient and forces the caller to guess how many items they need upfront.

The above 3 can still be performed by the using using the `child` method. Which is more idiomatic. (as long as the state of param is maintained properly)

### Documentation

- [ ] Other: 

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DerivationPathTest.kt
+++ b/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DerivationPathTest.kt
@@ -1,0 +1,131 @@
+package org.bitcoindevkit
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import kotlin.test.assertFalse
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.junit.runner.RunWith
+import kotlin.test.assertFailsWith
+
+@RunWith(AndroidJUnit4::class)
+class DerivationPathTest {
+    @Test
+    fun derivationPathFromString() {
+        val pathString = "m/84h/0h/0h/0/0"
+        val expectedPath = "84'/0'/0'/0/0"
+        val derivationPath = DerivationPath(pathString)
+        assertEquals(
+            expected = expectedPath,
+            actual = derivationPath.toString()
+        )
+    }
+
+    @Test
+    fun derivationPathtoString() {
+        val pathString = "m/44h/1h/0h/1/5"
+        val expectedPath = "44'/1'/0'/1/5"
+        val derivationPath = DerivationPath(pathString)
+        assertEquals(
+            expected = expectedPath,
+            actual = derivationPath.toString()
+        )
+    }
+
+    @Test
+    fun correctlyIdentifiesMaster() {
+        val pathString = "m"
+        val derivationPath = DerivationPath(pathString)
+        assertTrue(derivationPath.isMaster())
+
+        val nonMasterPathString = "m/0h/1/2"
+        val nonMasterDerivationPath = DerivationPath(nonMasterPathString)
+        assertFalse(nonMasterDerivationPath.isMaster())
+
+        val masterPathString = DerivationPath.master()
+        assertTrue(masterPathString.isMaster())
+    }
+
+    @Test
+    fun invalidDerivationPath(){
+        val invalidPathString = "invalid/path/string"
+        assertFailsWith<Bip32Exception.InvalidChildNumberFormat> { DerivationPath(invalidPathString) }
+    }
+
+    @Test
+    fun createChildDerivationPath() {
+        val parentPathString = "m/44h/0h"
+        val childIndex = 5u
+        val expectedNormalChildPath = "44'/0'/5"
+        val parentDerivationPath = DerivationPath(parentPathString)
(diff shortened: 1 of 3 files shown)


ItoroD:
> This is looking good!
> 
> Because it's a simple type and methods I'd love to see a few tests that would play around with the `DerivationPath` type. Would you mind adding a few, at least on the Android tests?

Sure! I could add those. I did add some tests to the rust side before now. But its good to see it on bindings side too!

bdk-ffi/src/keys.rs
```diff
@@ -100,6 +101,25 @@ impl DerivationPath {
     pub fn is_empty(&self) -> bool {
         self.0.is_empty()
     }
+
+    /// Create a new DerivationPath that is a child of this one.
+    pub fn child(&self, child_number: ChildNumber) -> Arc<Self> {
```
reez: Should this validate ChildNumber before converting? ChildNumber exposes raw u32, so callers can pass index >= 2^31 and we’ll build a non‑BIP32 path. Upstream has from_normal_idx/from_hardened_idx maybe child should return Result and use those checked constructors

bdk-ffi/src/types.rs
```diff
@@ -1331,3 +1332,36 @@ impl From<bdk_wallet::TxDetails> for TxDetails {
...
+impl From<BdkChildNumber> for ChildNumber {
+    fn from(value: BdkChildNumber) -> Self {
+        match value {
+            BdkChildNumber::Normal { index } => ChildNumber::Normal { index },
+            BdkChildNumber::Hardened { index } => ChildNumber::Hardened { index },
+        }
+    }
+}
+
+impl From<ChildNumber> for BdkChildNumber {
```
reez: Should this be TryFrom using the checked constructors to enforce the BIP32 index range? Right now any u32 passes through

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DerivationPathTest.kt
```diff
@@ -0,0 +1,115 @@
...
+        val parentPathString = "m/44h/0h"
+        val childIndex = 5u
+        val expectedNormalChildPath = "44'/0'/5"
+        val parentDerivationPath = DerivationPath(parentPathString)
+        val childDerivationPath = parentDerivationPath.child(ChildNumber.Normal(childIndex))
+        assertEquals(
+            expected = expectedNormalChildPath,
+            actual = childDerivationPath.toString()
+        )
+
```
reez, line 67: I think if you set childIndex to 2147483648u and then add `assertFailsWith<Exception> { DerivationPath(childDerivationPath.toString()) }`  it could fail because the path can be constructed but not parsed back because I think we’re bypassing the BIP32 index checks?

bdk-ffi/src/keys.rs
ItoroD: Shouldn't a caller be able to enter an index >= 2^31 ? Those are hardened path. The child method will treat it as such.

EDIT - SMH. I see now

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DerivationPathTest.kt
ItoroD: Ahh! that is a dangerous catch! Thanks! I will address this tomorrow

bdk-ffi/src/keys.rs
reez: When I was looking at https://github.com/rust-bitcoin/rust-bitcoin/blob/bitcoin-0.32.8/bitcoin/src/bip32.rs#L124-L162 I didnt think so, I was reading it as both Hardened and Normal required under that

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DerivationPathTest.kt
reez: sounds good, overall this pr is looking good though 👍

ItoroD: Made some updates. Added new test in relation to these changes too `checkInvalidNormalChildNumberFails` and `checkInvalidHardenedChildNumberFails`

bdk-ffi/src/types.rs
```diff
@@ -1331,3 +1332,42 @@ impl From<bdk_wallet::TxDetails> for TxDetails {
         }
     }
 }
+
+/// A child number in a derivation path
+#[derive(Copy, Clone, uniffi::Enum)]
+pub enum ChildNumber {
```
thunderbiscuit: Small nit: this type comes from the rust-bitcoin library, and so belongs in the `bitcoin.rs` file.

PR #943, opened 2026-01-28
test: Add test for descriptor types.

ItoroD:
### Description

This test was taken from our old Offline test that @thunderbiscuit created [here](https://github.com/bitcoindevkit/bdk-ffi/pull/688/changes#diff-9170994d90422aa9fb1547ba2a781a594c7963d63ae5b46dc4e7d1957a3f4f30R20). I only updated with a few additions like Check for `sortedmulti`. I was taking a look at issue #687 and figured these are good tests to bring back. These tests have shown some more descriptor types we can do in the bindings. I am particular curious about `tr` and the range of complex descriptors we are able to perform currently and those we cannot. Adding these test is a great way to start.

### Documentation

- [ ] Other: 

#### All Submissions:

### DIFF
--- a/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DescriptorTest.kt
+++ b/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DescriptorTest.kt
@@ -5,6 +5,7 @@ import androidx.test.ext.junit.runners.AndroidJUnit4
 import org.junit.Assert.assertFalse
 import org.junit.Assert.assertTrue
 import org.junit.runner.RunWith
+import kotlin.test.assertEquals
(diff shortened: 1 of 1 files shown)


ItoroD:
> nice addition. looks like its a lot of coverage, if we want full coverage we may want to add wsh & sh_wsh?

Thanks for the review. I have added those now.

I am hoping we would be able to update these tests sometime later to use the actual helper methods from miniscript as in #949

thunderbiscuit:
This one needs a quick rebase and is good to go. At some point I'd be great to maybe add to the test all the descriptors that are _not_ supported, just to make it more explicit and easier to answer questions like [this one](https://github.com/bitcoindevkit/bdk-python/issues/31).

PR #949, opened 2026-02-17
feat: Add `new_wsh_sortedmulti` and `new_pk`

ItoroD:
### Description
This Pr is based on my comment from #687 [here](https://github.com/bitcoindevkit/bdk-ffi/issues/687#issuecomment-3877896847). 

### Notes to the reviewers
In bdk-wallet we can do the following to create a 1 of 2 multisig descriptor. Thanks so miniscript helper functions
```rust
let multisig_descriptor = Descriptor::new_wsh_sortedmulti(1, vec![key.clone(), key2.clone()]).unwrap();
```

In bindings: kotlin, we do:
```kotlin
val multisigDescriptor = Descriptor("wsh(multi(1,$key,$key2))", Network.REGTEST)
```
Current we have to concatenate a string with the keys and script type to parse the descriptor.

Much easier and safer to be able to do:
```kotlin
val multisigDescriptor = Descriptor.newWshSortedmulti(2u, pkList);
```

### Documentation

- new_wsh_sortedmulti - https://docs.rs/miniscript/12.3.5/miniscript/descriptor/enum.Descriptor.html#method.new_sh_wsh_sortedmulti

- new_pk - https://docs.rs/miniscript/12.3.5/miniscript/descriptor/enum.Descriptor.html#method.new_pk
  
- [x] Other: 

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DescriptorTest.kt
+++ b/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DescriptorTest.kt
@@ -41,6 +41,37 @@ class DescriptorTest {
         Descriptor("tr($MAINNET_EXTENDED_PRIVKEY/$BIP86_MAINNET_RECEIVE_PATH/0)", Network.BITCOIN)
     }
 
+    // Create a multisig descriptor.
+    @Test
+    fun createMultisigDescriptor() {
+        val pkList = listOf(
+            "xpub661MyMwAqRbcFW31YEwpkMuc5THy2PSt5bDMsktWQcFF8syAmRUapSCGu8ED9W6oDMSgv6Zz8idoc4a6mr8BDzTJY47LJhkJ8UB7WEGuduB",
+            "xpub661MyMwAqRbcFW31YEwpkMuc5THy2PSt5bDMsktWQcFF8syAmRUapSCGu8ED9W6oDMSgv6Zz8idoc4a6mr8BDzTJY47LJhkJ8UB7WEGuduB",
+            "xpub661MyMwAqRbcFW31YEwpkMuc5THy2PSt5bDMsktWQcFF8syAmRUapSCGu8ED9W6oDMSgv6Zz8idoc4a6mr8BDzTJY47LJhkJ8UB7WEGuduB"
+        )
+        val wshDescriptor = Descriptor.newWshSortedmulti(2u, pkList);
+
+        assertEquals(wshDescriptor.descType(), DescriptorType.WSH_SORTED_MULTI)
+    }
+
+    @Test
+    fun createInvalidMutlisigDescriptor() {
+        val invalidPkList = listOf(
+            "invalid_xpub",
+            "1111111111111",
+            "***"
+        )
+        assertFails {
+            Descriptor.newWshSortedmulti(2u, invalidPkList);
+        }
+    }
+
+    @Test
+    fun createSingleKeyDescriptor() {
+        val newPkDescriptor = Descriptor.newPk("xpub661MyMwAqRbcFW31YEwpkMuc5THy2PSt5bDMsktWQcFF8syAmRUapSCGu8ED9W6oDMSgv6Zz8idoc4a6mr8BDzTJY47LJhkJ8UB7WEGuduB")
+        assertEquals(newPkDescriptor.descType(), DescriptorType.BARE)
+    }
+
     // Cannot create addr() descriptor.
     @Test
     fun cannotCreateAddrDescriptor() {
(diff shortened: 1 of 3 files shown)


ItoroD:
> this is really cool @ItoroD

Thanks!

If this get approved, I am hoping to bring in the other similar miniscript methods that take either a list of PKs or single Pk. There are other ones that take more [complex ](https://docs.rs/miniscript/12.3.5/miniscript/descriptor/enum.Descriptor.html#method.new_bare)params that I can focus my attention on later once these simple ones are in. 

I did just these 2 in this pr just as a sort of concept and starter for the rest.

bdk-ffi/src/descriptor.rs
```diff
@@ -305,6 +305,52 @@ impl Descriptor {
...
     }
 
+    /// Create a new wsh sorted multi descriptor
+    /// Errors when miniscript exceeds resource limits under p2sh context
+    #[uniffi::constructor]
+    pub fn new_wsh_sortedmulti(k: u32, pks: Vec<String>) -> Result<Self, DescriptorError> {
+        let bdk_pks: Vec<BdkDescriptorPublicKey> = pks
+            .iter()
+            .map(|pk| {
+                BdkDescriptorPublicKey::from_str(pk).map_err(|e| DescriptorError::Miniscript {
```
reez: Should we avoid reparsing raw strings into BdkDescriptorPublicKey here? Right now malformed keys become DescriptorError::Miniscript while the rest of the API exposes key parsing through DescriptorPublicKey and key oriented error variants. Taking DescriptorPublicKey directly, or adding a dedicated conversion for key parse failures would keep this constructor consistent with the existing bindings surface I think

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DescriptorTest.kt
```diff
@@ -40,6 +40,32 @@ class DescriptorTest {
...
+    // Create a multisig descriptor.
+    @Test
+    fun createMultisigDescriptor() {
+        val pkList = listOf(
+            "xpub661MyMwAqRbcFW31YEwpkMuc5THy2PSt5bDMsktWQcFF8syAmRUapSCGu8ED9W6oDMSgv6Zz8idoc4a6mr8BDzTJY47LJhkJ8UB7WEGuduB",
+            "xpub661MyMwAqRbcFW31YEwpkMuc5THy2PSt5bDMsktWQcFF8syAmRUapSCGu8ED9W6oDMSgv6Zz8idoc4a6mr8BDzTJY47LJhkJ8UB7WEGuduB",
+            "xpub661MyMwAqRbcFW31YEwpkMuc5THy2PSt5bDMsktWQcFF8syAmRUapSCGu8ED9W6oDMSgv6Zz8idoc4a6mr8BDzTJY47LJhkJ8UB7WEGuduB"
+        )
+
+        Descriptor.newWshSortedmulti(2u, pkList);
```
reez: Should we assert the resulting descriptor string or descType() here? I think this only checks that construction succeeds so it I dont know if it would catch a wrapper bug that returns the wrong descriptor variant

ItoroD: Yes good call. Will update

bdk-ffi/src/descriptor.rs
ItoroD: Taking DescriptorPublicKey directly will mean the user will parse his string first before passing it to our new methods? Isn't it better if the method handles both? 

With regards to the error, currently you will get a `DescriptorKeyException` if you pass a malformed key. Also, some of these constructors are not going to be taking pk alone, they will take actual miniscript like [here](https://docs.rs/miniscript/12.3.5/miniscript/descriptor/enum.Descriptor.html#method.new_sh)

ItoroD: I just ran the android test with a malformed key and indeed its a `DescriptorException$Miniscript` error that shows up. I should be able to change it to give a DescriptorKeyException like it does when you parse with DescriptorPublicKey.

About whether to take strings as param or DescriptorPublicKey, I was only thinking about the way we have been passing things into the descriptor type so far (with strings). Hoping it will not be too disruptive it use DescriptorPublicKey.fromString first before they they pass it to this constructor.

reez: Yeah your right keeping String params here is good w me, that makes sense.

I’d be interested to hear what you find if you look into the error mapping. If it’s an easy fix, that would be great, but if it's not then I am def good with merging the pr as is.

Good work on this overall, its has some tricky bits for sure.

ItoroD: I have looked at the Error variant we are using again and I come to the conclusion that `DescriptorError::Key` might be the best to use that case. It maps to [KeyError ](https://docs.rs/bdk_wallet/2.3.0/bdk_wallet/keys/enum.KeyError.html)in bdk-wallet and it seems to have a wide range of error variants.

Issue #941, opened 2026-01-21
Publish Rust API Docs for bdk-ffi


thunderbiscuit:
I propose we build and publish the Rust API docs website for this library. I have [published a test of this on my fork of bdk-ffi](https://thunderbiscuit.github.io/bdk-ffi/bdkffi/index.html). At first glance it might appear silly to do this since we don't have any Rust users and only users at the bindings layer, but let me outline some of the pros I see:

1. The Rust API docs are clean! Since they are generated directly from our code, they do not contain any of the extra uniffi stuff that pollutes the other API docs website and that Mozilla doesn't make private. See the difference between the [Rust docs](https://thunderbiscuit.github.io/bdk-ffi/bdkffi/index.html) and the [Kotlin docs](https://bitcoindevkit.org/android/). The Kotlin docs, while easier to read and navigate for Kotlin devs, contains all sorts of "not supposed to use" types that just pollute the docs but that we can't remove easily.
2. For a number of languages, producing API docs is hard and not at the top of the priority list. For example I've been working with 2 React Native teams on their migration from the old bdk-rn to the new 2.2.0 bdk-rn, and they have no access to API docs at all, and must instead rely on IDE hints and autocomplete. Have a ready-to-go website I can share links to would have been great, even if it's not in their native web-wrapping UI they're familiar with.
3. Even for me and maintainers on the team, I think having access to this would be useful. The API has changed so much over time, it can be hard to answer quickly a question from a user who is not on the latest version. Having access to this as a reference would be fantastic.
4. This will force us to clean up our docs! Building the docs locally throws a ton of warnings regarding broken links. It's a bonus that building this will get us to clean the docs a bit.

reez:
These are live now great job @thunderbiscuit ! 

closing issue

thunderbiscuit:
Yep! Loving them I use them all the time 😆. For anyone looking, they're hosted here: https://bitcoindevkit.github.io/bdk-ffi/bdkffi/index.html

PR #951, opened 2026-02-17
Enable building Rust API docs

thunderbiscuit:
I'm not sure of the best way to do this, because we need the modules to be public for the docs to build correctly, but so far they've been private (you can't use the lib from another Rust project).

This is open mainly for discussion purposes. I could also just build locally and publish. See #941 

#### All Submissions:

### DIFF
--- a/.github/ISSUE_TEMPLATE/release.md
+++ b/.github/ISSUE_TEMPLATE/release.md
@@ -51,10 +51,15 @@ git push upstream v0.6.0
 ```
 14. - [ ] Trigger manual releases for both libraries (for Swift, go to the [bdk-swift](https://github.com/bitcoindevkit/bdk-swift) repository and trigger the workflow using `master`. Simply add the version number and tag name in the text fields when running the workflow manually. Note that the version number must not contain the `v`, i.e. `0.26.0`, but the tag will have it, i.e. `v0.26.0`). For Android, trigger the release workflow using the tag (not a branch).
 15. - [ ] Make sure the released libraries work and contain the artifacts you would expect.
-16. - [ ] Aggregate all the changelog notices from the PRs and add them to the changelog file. PR that.
-17. - [ ] Bump the version on master from `1.1.0-SNAPSHOT` to `1.2.0-SNAPSHOT` (Android) and `1.1.0-alpha.0` to `1.2.0-alpha.0` (Rust).
-18. - [ ] Apply changes to the release issue template if needed.
-19. - [ ] Make release on GitHub (generate auto release notes between the previous tag and the new one).
-20. - [ ] Build API docs for Android locally and PR the website to the bitcoindevkit.org repo.
-21. - [ ] Post in the announcement channel.
-22. - [ ] Tweet about the new release!
+16. - [ ] Build the Rust API docs and publish them to the repo's GitHub Pages.
+```shell
+cd bdk-ffi/
+just docs
+```
+17. - [ ] Aggregate all the changelog notices from the PRs and add them to the changelog file. PR that.
+18. - [ ] Bump the version on master from `1.1.0-SNAPSHOT` to `1.2.0-SNAPSHOT` (Android) and `1.1.0-alpha.0` to `1.2.0-alpha.0` (Rust).
+19. - [ ] Apply changes to the release issue template if needed.
+20. - [ ] Make release on GitHub (generate auto release notes between the previous tag and the new one).
+21. - [ ] Build API docs for Android locally and PR the website to the bitcoindevkit.org repo.
+22. - [ ] Post in the announcement channel.
+23. - [ ] Tweet about the new release!
--- a/bdk-ffi/justfile
+++ b/bdk-ffi/justfile
@@ -13,3 +13,7 @@ check:
 
 test:
   cargo test
+
+# Build API documentation for gh-pages or local browsing
+docs:
+  bash scripts/build-docs.sh && bash scripts/deploy-docs.sh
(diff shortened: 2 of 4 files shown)


bdk-ffi/src/lib.rs
```diff
@@ -1,15 +1,15 @@
...
-mod error;
-mod esplora;
-mod keys;
-mod kyoto;
-mod macros;
-mod store;
-mod tx_builder;
-mod types;
-mod wallet;
+pub mod bitcoin;
```
reez: Could we switch the docs recipe to cargo doc --no-deps --document-private-items instead of making every module pub? That still gives us browsable module docs, but avoids committing bdkffi::{bitcoin,error,...} to a public Rust API.

thunderbiscuit: I'll reply in the main thread!

.github/ISSUE_TEMPLATE/release.md
```diff
@@ -51,10 +51,16 @@ git push upstream v0.6.0
...
 15. - [ ] Make sure the released libraries work and contain the artifacts you would expect.
-16. - [ ] Aggregate all the changelog notices from the PRs and add them to the changelog file. PR that.
-17. - [ ] Bump the version on master from `1.1.0-SNAPSHOT` to `1.2.0-SNAPSHOT` (Android) and `1.1.0-alpha.0` to `1.2.0-alpha.0` (Rust).
-18. - [ ] Apply changes to the release issue template if needed.
-19. - [ ] Make release on GitHub (generate auto release notes between the previous tag and the new one).
-20. - [ ] Build API docs for Android locally and PR the website to the bitcoindevkit.org repo.
-21. - [ ] Post in the announcement channel.
-22. - [ ] Tweet about the new release!
+16. - [ ] Build the Rust API docs and publish them to the repo's GitHub Pages.
```
reez, line 54: 👍

.github/ISSUE_TEMPLATE/release.md
```diff
@@ -51,10 +51,16 @@ git push upstream v0.6.0
...
+```shell
+cd bdk-ffi/
+# Temporarily make all modules public in the src/lib.rs file
+source scripts/deploy-docs.sh
```
reez: nit (non-blocking): could use bash here instead. the nit part is I think source runs in releasers current shell so the scripts cd and any shell state changes leak into the session where bash wouldnt

bdk-ffi/justfile
```diff
@@ -13,3 +13,9 @@ check:
 
 test:
   cargo test
+
+# Build API documentation for gh-pages or local browsing
+docs:
+  cargo doc --no-deps
```
reez: I may be misunderstanding the target output here, but here's my understanding: the current `cargo doc --no-deps` avoids exposing private APIs which matches what we talked about and makes sense. But it also won’t produce module pages like bdkffi::bitcoin, bdkffi::wallet, etc. I dont think I have a strong opinion either way, but are the reduced crate docs ok with you or were you still thinking we wanted to have those wrapper module pages available somehow? 

Maybe some of my confusion is seeing release.md `#Temporarily make all modules public in the src/lib.rs file` comment still there and im not sure if it should/shouldnt be.

.github/ISSUE_TEMPLATE/release.md
```diff
@@ -51,10 +51,16 @@ git push upstream v0.6.0
...
-17. - [ ] Bump the version on master from `1.1.0-SNAPSHOT` to `1.2.0-SNAPSHOT` (Android) and `1.1.0-alpha.0` to `1.2.0-alpha.0` (Rust).
-18. - [ ] Apply changes to the release issue template if needed.
-19. - [ ] Make release on GitHub (generate auto release notes between the previous tag and the new one).
-20. - [ ] Build API docs for Android locally and PR the website to the bitcoindevkit.org repo.
-21. - [ ] Post in the announcement channel.
-22. - [ ] Tweet about the new release!
+16. - [ ] Build the Rust API docs and publish them to the repo's GitHub Pages.
+```shell
+cd bdk-ffi/
+# Temporarily make all modules public in the src/lib.rs file
```
reez: Not sure if this comment still holds or not is all, see comment in `justfile` where I might be confused

.github/ISSUE_TEMPLATE/release.md
thunderbiscuit: Ah yes ok good catch.

bdk-ffi/justfile
reez: (Or maybe also I'm just playing catch up getting my head around all the nuances, I know you're ahead of me on that and playing around w generating these)

thunderbiscuit: Mmmm I might be misunderstanding what the --no-deps is doing let me try locally. I thought it meant not producing the docs for our dependencies, like bdk_esplora, anyhow, request, etc., and only producing the docs for bdk-ffi types.

.github/ISSUE_TEMPLATE/release.md
thunderbiscuit: If you do not make the modules public, the resulting docs website is empty (since the whole library is private from Rust's perspective), so as far as I understand it you have to do that in order to build the correct site.

bdk-ffi/justfile
thunderbiscuit: Oh I think I understand. Do you mean running this command on its own without making the modules public will not really do what we need (since the website is empty)? I maybe should add the comment here as well so people know to make the modules public before running the `just docs` command.

PR #954, opened 2026-02-18
Post release fixes

thunderbiscuit:
These are the usual post-release commits:

- Bumping the Android lib to `2.4.0-SNAPSHOT` and the Rust lib to `2.4.0-alpha.0`
- Adding the 2.3.0 entry to the changelog

Notice that I didn't add the full links in the markdown file for the changelog, as I _think_ GitHub connects those for us anyway. I'll add them if I am wrong and they're needed, but it's just so much bloat on the file that I figured I would try without.


reez (approved):
ACK 36ea2f462424a523051f17123677ef05dbbb0a3b

PR #965, opened 2026-03-05
Add small Kyoto integration test

thunderbiscuit:
Update: this is now simply adding an integration test for Kyoto, the kind that would have caught our little bug in the 2.3.0 release of the client. It uses the CI regtest environment brought in with #967.

~~Trying to connect the dots from bdk_kyoto all the way to the the bindings. bitcoindevkit/bdk-kyoto#146~~

### DIFF
--- a/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/CbfSyncTest.kt
+++ b/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/CbfSyncTest.kt
@@ -1,0 +1,80 @@
+package org.bitcoindevkit
+
+import kotlin.test.Test
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.time.withTimeout
+import kotlinx.coroutines.withTimeout
+import org.junit.runner.RunWith
+import org.kotlinbitcointools.regtesttoolbox.regenv.RegEnv
+import java.io.File
+import kotlin.time.Duration.Companion.seconds
+
+@RunWith(AndroidJUnit4::class)
+class CbfSyncTest {
+    private val conn: Persister = Persister.newInMemory()
+
+    private val kyotoDataDir: String by lazy {
+        val context = InstrumentationRegistry.getInstrumentation().targetContext
+        val dir = File(context.filesDir, "kyoto_test_data")
+        dir.mkdirs()
+        dir.absolutePath
+    }
+
+    @Test
+    fun syncWithKyotoClient() {
+        runBlocking {
+            val regtestEnv = RegEnv.connectTo(
+                host = "10.0.2.2",
+                walletName = "faucet",
+                username = "regtest",
+                password = "password",
+            )
+
+            val wallet: Wallet = Wallet.createSingle(
+                descriptor = BIP86_DESCRIPTOR,
+                network = Network.REGTEST,
+                persister = conn
+            )
+            val newAddress = wallet.revealNextAddress(KeychainKind.EXTERNAL).address
+
+            regtestEnv.send(newAddress.toString(), 0.12345678, 2.0)
+            regtestEnv.mine(2)
+
+            val ip: IpAddress = IpAddress.fromIpv4(10u, 0u, 2u, 2u)
+            val peer1: Peer = Peer(ip, 18444u, false)
+            val peers: List<Peer> = listOf(peer1)
+            val (client, node) = CbfBuilder()
+                .peers(peers)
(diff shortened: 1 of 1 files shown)


bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/CbfSyncTest.kt
```diff
@@ -0,0 +1,79 @@
...
+        dir.absolutePath
+    }
+
+    @Test
+    fun syncWithKyotoClient() {
+        runBlocking {
+            // println("Hello, Compact Block Filters!")
+            // val regtestEnv = RegEnv.connectTo(walletName = "faucet", username = "regtest", password = "password")
+
+            val wallet: Wallet = Wallet.createSingle(
```
reez, line 36: Should this use a regtest descriptor here? wondering because from what I see BIP86_DESCRIPTOR is created with Network.TESTNET but the wallet is initialized with Network.REGTEST

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/CbfSyncTest.kt
```diff
@@ -0,0 +1,79 @@
...
+                    while (true) println("Info: ${client.nextInfo()}")
+                } catch (e: CbfException) { println("Info channel closed: $e") }
+            }
+
+            val update: Update = client.update()
+            delay(8.seconds)
+            wallet.applyUpdate(update)
+
+            val balance = wallet.balance().total.toSat()
+            assert(balance > 0uL)
```
reez, line 73: Should this use a real test assertion and reenable the funding/mine steps, I think the wallet is never funded and the final assert(...) may be skipped so the test can pass without validating Kyoto sync? Maybe im wrong though

reez (commented):
Looks good overall and consistent w bdk kyoto, a couple questions comments added

thunderbiscuit:
I'll update and rebase this once #967 is in, leveraging the regtest on CI.

reez:
> I'll update and rebase this once #967 is in, leveraging the regtest on CI.

ack'd 967

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/CbfSyncTest.kt
thunderbiscuit: Descriptors created using the Network.TESTNET, SIGNET, and REGTEST variants of the enum will all look the same (produce the same keys). The _addresses_ they produce will be different between regtest and the others, but the descriptors are valid for all 3. 

I guess that's one reason maybe to migrate to the new `NetworkKind` in places where it can be used (I think the descriptors are an example of that), so as to make it easier/more explicit what's what. Any opinions on this? I could open a small PR just exposing that to see if we like it.

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/CbfSyncTest.kt
thunderbiscuit: In this case the podman regtest infinity pro (!) provides an [always-funded faucet wallet](https://github.com/thunderbiscuit/podman-regtest-infinity-pro/blob/49051f188de91be8c12065e93fa5026c8784fc69/start-services.sh#L19-L21). That's why it works without further tooling or mining of blocks!

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/CbfSyncTest.kt
```diff
@@ -0,0 +1,77 @@
...
+                    while (true) println("Warning: ${client.nextWarning()}")
+                } catch (e: CbfException) { println("Warning channel closed: $e") }
+            }
+            val infoJob = launch {
+                try {
+                    while (true) println("Info: ${client.nextInfo()}")
+                } catch (e: CbfException) { println("Info channel closed: $e") }
+            }
+
+            val update: Update = client.update()
```
reez: Could this hang in CI if `client.update()` never returns?

rustaceanrob: There are a number of errors and timeouts that should prevent this, but if possible it would be prudent to wrap this in something around a 60 second timeout. If a sync takes longer than that we should investigate why.

thunderbiscuit: Yep that's an easy fix. New line is

```kotlin
val update: Update = withTimeout(60.seconds) { client.update() }
```

reez (approved):
ACK 73ff393f0a9fab830a7e65ee60e4b24936f46460

PR #967, opened 2026-03-06
Add test for Esplora client on Regtest 

thunderbiscuit:
This PR brings in the [regtest-toolbox](https://github.com/kotlin-bitcoin-tools/regtest-toolbox) library as a test dependency (not shipped with the release artifacts) in order to control a local regtest node and allow for testing clients in our CI workflows. It also brings in the [Regtest Infinity Pro](https://regtestinfinitypro.dev/) as a service that gets fired up before the CI runs the tests.

We are currently using this setup in bdk-jvm with success.

Using this would allow us to test clients, and better see/verify changes in the API on client PRs (for example in #965).

### Notes to the reviewers

Opened as a discussion piece.

### Documentation

It does not affect the public API.

#### All Submissions:

#### New Features:


ItoroD (approved):
ACK https://github.com/bitcoindevkit/bdk-ffi/commit/9ad254e7ef9668eefd7a303815156ac04e5d107e

Loving the toolbox

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/EsploraClientTest.kt
```diff
@@ -0,0 +1,51 @@
...
+                network = Network.REGTEST,
+                persister = conn
+            )
+            val newAddress = wallet.revealNextAddress(KeychainKind.EXTERNAL).address
+
+            regtestEnv.send(newAddress.toString(), 0.12345678, 2.0)
+            regtestEnv.mine(2)
+
+            //Wait 8 second for mining to complete and for esplora to index the new blocks before scanning
+            delay(8.seconds)
```
reez: has 8 seconds proven to always be enough time for this?

ItoroD: So far yes. Check [this ](https://github.com/kotlin-bitcoin-tools/regtest-toolbox/issues/3) out too.

thunderbiscuit: Yes the 8 second delay is really not a clean way to do this (what if the CI is much slower than local, or the opposite? One side or another you're waiting for longer than you should and it's just overall hacky). 

We used this approach to get something working and opened an issue on regtest-toolbox thinking maybe a solution could be added there, but I think @ItoroD's idea of polling the Electrum server until it sees the transaction or block is probably the best way to do this test-side.

I'll add the polling loop here and test locally between runs to see the general feel for it (maybe looping on 1/10th of a second is too much? Maybe not). But in any case I'll use that approach here, and we should clean up the JVM tests as well.

reez: yeah the polling loop sounds cleaner, I like it

thunderbiscuit: I ended up with this:

```kotlin
// Wait for the Esplora client to see the transaction. Try 5x per second for 20 seconds.
repeat(100) {
    if (esploraClient.getTx(txid) != null) return@repeat
    delay(200.milliseconds)
}
```

Which usually takes between 10 and 30 loops to complete on my machine.

ItoroD: LGTM!

Since this is a closed test setup it is reasonable to call 5 times per second without it being considered spamming.

reez: This is definitely a good direction, but I dont think return@repeat breaks the loop? I think it might only return from the current iteration. Maybe we switch this to a while loop or helper that exits early and fails explicitly on timeout?

ItoroD: `return@runBlocking` breaks out of the loop.

reez: `return@runBlocking` is nice but I think it would exit the whole test body here not just the polling loop?

ItoroD: Ahh that's what I used in my fork test I think the reason I didn't notice was the test still said success although I have unknowingly existed the program. Blinded by success 😁. Really good catch @reez 

This is what I meant to do [here](https://github.com/ItoroD/bdk-jvm/pull/2/changes#diff-ab6c5a1fbbcc9ca8aa0abb0ca881117278ad4342ce8ae64ca7fc6a2d63c98a2bR34-R39) return run not runBlocking

thunderbiscuit: Good catch @reez. I updated with:

```kotlin
for (i in 0..99) {
    if (esploraClient.getTx(txid) != null) break
    delay(200.milliseconds)
}
```

which accomplishes the stated goal of only running the required number of iterations.

thunderbiscuit:
Ready for final review.

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/EsploraClientTest.kt
reez: LGTM now 👍

reez (approved):
ACK 0b8fafc5317763b3d3917ca75ab6ea066a10bd4a

PR #971, opened 2026-03-11
3.0

### LINKED ISSUE #869: bdk_wallet 2.3 or 3.0

reez:
### Description

Exposing 3.0 items.

### Documentation

- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### Changelog

```
Added
- Expose wallet event helpers `Wallet::apply_unconfirmed_txs_events` and `Wallet::apply_evicted_txs_events` [#971]
- Expose wallet scan/sync helpers `Wallet::start_full_scan_at`, `Wallet::start_sync_with_revealed_spks_at`, and `Wallet::checkpoints` [#971]
- Expose locked outpoint wallet APIs and locked outpoint persistence in `ChangeSet` [#971]
- Expose transaction builder controls for sighash, transaction ordering, and adding foreign UTXOs with explicit sequence values [#971]
- Expose `Persister::get_pre_v1_wallet_keychains` for pre-v1 SQLite wallet migration support [#971]

[#971]: https://github.com/bitcoindevkit/bdk-ffi/pull/971
```


reez:
updated to official 3.0

rustaceanrob:
`bdk_kyoto = 0.16.0` is out and includes the `3.0.0` release

reez:
> `bdk_kyoto = 0.16.0` is out and includes the `3.0.0` release

thanks I updated to 0.16.0 and made one small change `update_rx: Mutex<UpdateSubscriber<bdk_kyoto::wallets::Single>>` necessary along with it in https://github.com/bitcoindevkit/bdk-ffi/pull/971/changes/27be17fcae5b8ab5883a578c2e9f65c4bb89da1b

bdk-ffi/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "bdk-ffi"
-version = "2.4.0-alpha.0"
+version = "3.0.0"
```
thunderbiscuit: This one here I think should read `3.0.0-alpha.0` since this would merge on master.

bdk-ffi/src/descriptor.rs
```diff
@@ -38,7 +38,8 @@ impl Descriptor {
     #[uniffi::constructor]
     pub fn new(descriptor: String, network: Network) -> Result<Self, DescriptorError> {
         let secp = Secp256k1::new();
-        let (extended_descriptor, key_map) = descriptor.into_wallet_descriptor(&secp, network)?;
+        let (extended_descriptor, key_map) =
+            descriptor.into_wallet_descriptor(&secp, network.into())?;
```
thunderbiscuit, line 42: I still need to get to grips with the `NetworkKind` type. Are there situations where we'd like to use that instead of the `Network` for example? Not sure.

bdk-ffi/src/store.rs
```diff
@@ -21,6 +25,17 @@ pub(crate) enum PersistenceType {
     Sql(Mutex<BdkConnection>),
 }
 
+/// Metadata describing a keychain in a pre-v1 BDK SQLite wallet database.
+#[derive(Debug, Clone, uniffi::Record)]
+pub struct PreV1WalletKeychain {
+    /// The wallet keychain.
+    pub keychain: bdk_wallet::KeychainKind,
```
thunderbiscuit, line 33: Can we use our own `KeychainKind` from the `types` module here? I think they're functionally equivalent because it's using remote enums in uniffi, but could still be a tiny cleanup unless I am missing something.

bdk-ffi/src/store.rs
```diff
@@ -54,6 +69,32 @@ impl Persister {
             inner: PersistenceType::Custom(persistence).into(),
         }
     }
+
+    /// Retrieve keychain metadata from a pre-v1 BDK SQLite wallet database.
+    pub fn get_pre_v1_wallet_keychains(
```
thunderbiscuit, line 76: I'd love to see a test for this one in the Android or Swift parts of the codebase if possible.

bdk-ffi/src/store.rs
```diff
@@ -21,6 +25,17 @@ pub(crate) enum PersistenceType {
     Sql(Mutex<BdkConnection>),
 }
 
+/// Metadata describing a keychain in a pre-v1 BDK SQLite wallet database.
+#[derive(Debug, Clone, uniffi::Record)]
+pub struct PreV1WalletKeychain {
```
thunderbiscuit, line 31: I was looking at the Rust code for this type and I see that the API docs have changed since you added it I think. See [here](https://docs.rs/bdk_wallet/3.0.0/src/bdk_wallet/wallet/migration.rs.html#27-35).

thunderbiscuit (commented):
I compiled and ran Rust tests locally today. Just started looking at this and left a few comments. I haven't gone through all files yet (tx_builder, types, wallet left to do). I'll try and migrate the Android app to this potential alpha to see how the migration goes! Looks like it should be pretty smooth.

bdk-ffi/Cargo.toml
reez: Ahhh, I guess I did no `-alpha` just because in my head I was thinking we only bumped to `-alpha` after the release work was merged to master and the release was cut, but I might have been remembering that wrong. 

No strong opinion from me, just wanted to double check w ya, but im good with whatever you comment back on this for me to change it to.

bdk-ffi/src/descriptor.rs
reez: I don’t have any concrete ffi facing cases that come to mind, since NetworkKind only has Main and Test... I’ve been thinking of it as mostly an internal/upstream type for apis that only care about that distinction, where bdk-ffi consumers would usually want the full Network to preserve the actual chain. But I need to keep thinking about it more too so im glad you commented on it here

bdk-ffi/src/store.rs
reez: I think the small uniffi wrinkle i found here was `crate::types::KeychainKind` isn’t currently usable from this module because the backing alias in types.rs is private so switching this line wasn't a easy one line cleanup, and since the rest of the code was also using `bdk_wallet::KeychainKind` directly I decided to leave it as is in this PR.

But its definitely something we could do, it just seemed like it might end up needing to be a separate/bigger refactor

bdk-ffi/src/store.rs
reez: Yeah, that makes sense. I looked around the code and I think we already have Swift/Android persistence tests but I don’t think we had a pre-v1 SQLite database  in the repo, so adding a bindings test for this I think would have meant adding that first. And then also I wasn't totally sure if we were going to expose this in 3.0 or not.

But im with you on if we do want to expose this for sure in our ffi 3.0 (or whenever we expose it in a release) that it would be quite nice to have a test for this, and prob worth putting in time to figure out how to test it well

bdk-ffi/src/store.rs
reez: this is deep in the weeds, I love it, good catch!

updated in 6cca30e35c6f7eca02d50722fc2f432db6e0a679

bdk-ffi/Cargo.toml
reez: (go w thunders suggestion here)

reez: updated w commit 061266b46eb5a49a1ad396b7ad41c430319ac5d4

reez:
@thunderbiscuit I rebased onto current master and dropped commits superseded by #978

reez:
rebased

bdk-ffi/src/tests/tx_builder.rs
```diff
@@ -198,6 +199,61 @@ fn test_only_witness_utxo_with_finish() {
...
+}
+
+#[test]
+fn test_sighash_sets_psbt_input_sighash_type() {
+    let wallet = create_and_sync_wallet();
+    let utxos = wallet.list_unspent();
+
+    if utxos.is_empty() {
+        println!("No UTXOs available, skipping sighash PSBT assertion");
+        return;
```
j-kon: Nice addition covering the new `sighash` API.

One small test concern: this can pass without actually exercising the new behavior if the synced wallet has no UTXOs, because the test returns before checking `input.sighash_type`.

Would it be better to make the funded-wallet precondition explicit, or structure this so the test cannot silently skip the PSBT assertion?

reez: yeah, updated the test now

reez:
rebased again

j-kon (approved):

bdk-ffi/src/wallet.rs
```diff
@@ -341,6 +341,29 @@ impl Wallet {
...
 
+    /// Apply relevant unconfirmed transactions to the wallet and returns events.
+    ///
+    /// See [`apply_unconfirmed_txs`] for more information.
+    ///
+    /// See [`apply_update_events`] for more information on the returned [`WalletEvent`]s.
+    ///
+    /// [`apply_unconfirmed_txs`]: Self::apply_unconfirmed_txs
+    /// [`apply_update_events`]: Self::apply_update_events
+    pub fn apply_unconfirmed_txs_events(
```
thunderbiscuit, line 352: I'm looking at this and the sister method `apply_unconfirmed_txs()`, and I'm wondering: how would you come across an `UnconfirmedTx`? We only use them as inputs to these two methods, and so someone would need to create one to them apply it? Wondering what the workflow is

bdk-ffi/src/store.rs
```diff
@@ -54,6 +71,32 @@ impl Persister {
             inner: PersistenceType::Custom(persistence).into(),
         }
     }
+
+    /// Retrieve keychain metadata from a pre-v1 BDK SQLite wallet database.
+    pub fn get_pre_v1_wallet_keychains(
+        &self,
+    ) -> Result<Vec<PreV1WalletKeychain>, PreV1MigrationError> {
```
thunderbiscuit, line 78: The method is on the Persister type, but in the Rust library it's just a [standalone function](https://docs.rs/bdk_wallet/latest/bdk_wallet/migration/fn.get_pre_v1_wallet_keychains.html). I'm wondering if there is an advantage to this approach of making it into a associated fuction or some other reason you had in mind.

bdk-ffi/src/store.rs
thunderbiscuit: Ok this is so confusing to me 😆😆😆.

You're right most of the codebase uses the Rust type directly! But it's also... a trick!? I was confused by this ability to just use the Rust enum without any other fanfare (maybe a new feature from 0.30 I didn't know about?), especially that when you look at the generated bindings Kotlin code, the enum used in types.rs is the same enum used in say descriptors.rs, when in fact looking at the code those are two completely different types (one is custom made by us, one is the bdk_wallet library's type).

And so I was like "ok let's just cut it completely then!", but if you try to do that (remove the KeychainKind from types.rs), you actually get errors at the uniffi level, saying that KeychainKind is undefined. But wait it's just using the one from bdk_wallet... mmmmm

So it turns out that it doesn't. All files rely on our `KeychainKind` in types.rs, and I think this is related to how the visibility modifiers in Rust are disregarded by uniffi. You're correct it's not public, and yet the sort of type alias or whatever is used wherever you use KeychainKind.

I was able to clean this up and reconcile this is by simply making the type public and cleaning up the misleading imports in files that use the type. You're right it's sort of a sidequest, so I'll open a separate PR for it.

thunderbiscuit (commented):
Still working through it but here are some thoughts!

bdk-ffi/src/wallet.rs
reez: I think this one is mostly for custom mempool/backend flows, not really the normal esplora/electrum `sync -> apply_update` path. (`UnconfirmedTx` is basically just the ffi version of upstream’s `(tx, last_seen)` tuple.) So the workflow in my mind would be something like: your app/backend sees a raw mempool tx, parses it into a `Transaction`, pairs it with the unix timestamp for when it saw it, and passes that in here.

bdk-ffi/src/store.rs
reez: yeah I was thinking of Persister as the FFI stand in for the rusqlite::Connection here. Since upstream takes &mut Connection (and bdk-ffi hides that behind Persister) putting it there let callers run the migration helper against the same sqlite persister they already use for loading/persisting.

BUT I see the downside that it shows up on custom persisters too and then needs the SqliteOnly error. I don’t feel super strongly about it. What ideas are you thiniking about that are better, im def open to them

reez:
https://github.com/bitcoindevkit/bdk-ffi/pull/971#discussion_r3163408752

@thunderbiscuit 

Yeah agree, ack'd it

bdk-ffi/src/store.rs
thunderbiscuit: Resolved with #985.

bdk-ffi/src/descriptor.rs
thunderbiscuit: I'm slowly coming around to this `NetworkKind` type. I think the gist of the issue with `Network` in some of the APIs is that it lures you into thinking the network produces distinctive outputs, when in fact there are only two potential outcomes (test networks and the main network). This is the case for the `Descriptor` type for example. At the same time, there _are_ instances where the specific network matters, like in creating a `Wallet` or deriving addresses from a certain descriptor.

I'll open a PR to switch it and we can discuss further there if we want to bring this in. See #986

reez:
just rebased

bdk-ffi/src/store.rs
thunderbiscuit: I have created a db from a 0.X wallet we can use to test.

```kotlin
    @Test
    fun preV1Migration() {
        val db = Persister.newSqlite(preV1Wallet)
        val preV1Keychains: List<PreV1WalletKeychain> = db.getPreV1WalletKeychains()

        assertEquals(
            expected = 1,
            preV1Keychains.size
        )
        assertEquals(
            expected = KeychainKind.EXTERNAL,
            actual = preV1Keychains.first().keychain
        )
        assertEquals(
            expected = 7u,
            actual = preV1Keychains.first().lastDerivationIndex
        )
        assertEquals(
            expected = "xh44xwsp",
            actual = preV1Keychains.first().checksum
        )
    }
```

reez: pulled your https://github.com/thunderbiscuit/bdk-ffi/commit/7aa3e6a990a7ad9c5ff7ff3d65dbb84ac697d213 into https://github.com/bitcoindevkit/bdk-ffi/pull/971/changes/dbb19b347b04058d62311d02bab19537e799900a

bdk-ffi/src/store.rs
```diff
@@ -21,6 +25,19 @@ pub(crate) enum PersistenceType {
     Sql(Mutex<BdkConnection>),
 }
 
+/// `PreV1WalletKeychain` represents a structure that holds the keychain details
+/// and metadata required for managing a wallet's keys.
+#[derive(Debug, Clone, uniffi::Record)]
+pub struct PreV1WalletKeychain {
+    /// The name of the wallet keychains, "External" or "Internal".
+    pub keychain: bdk_wallet::KeychainKind,
```
thunderbiscuit: This field can now use the `KeychainKind` type directly, which comes from the `crate::types::KeychainKind` import.

bdk-ffi/src/store.rs
thunderbiscuit: I see the test is now added.

thunderbiscuit (changes requested):
Ok we're almost here. I was coming to give my ACK and found one last little thing. After that, ready to merge IMO.

thunderbiscuit:
Oh and I think adding the changelog entry to this PR would be great just because it has a lot.

bdk-ffi/src/store.rs
reez: updated

reez:
> Oh and I think adding the changelog entry to this PR would be great just because it has a lot.

added

thunderbiscuit (approved):
ACK eb3696b0cd0c650206ba8fdac0a91be2ec980675.

Issue #769, opened 2025-05-27
Rethinking our test suite


thunderbiscuit:
This issue will keep track of a potentially big refactor of our tests.

Here are some thoughts, discussed over a few dev calls:

- Building for all architectures for a given language has not proven to be useful, but has proven to be slow! We should probably run our tests for a single architecture and a single language version (for Python).
- Having tests that hit live endpoints on the Signet network is just not a good idea. We should probably have a beefier Regtest setup and run the tests there.
- A lot of our current tests are better thought of as examples. We should consider compiling those but not actually running them. This would catch API breaks and whatnot without having the flakiness of the live tests, while also leaving full live examples devs can run locally if they need to, or use as starting points for building on the API.

thunderbiscuit:
A more concrete todo list of tasks for this would look something like this:

- [ ] Remove all tests that hit live signet endpoints
- [ ] Make use of a persisted wallet that has some balance to test some of the transaction building APIs
- [ ] Build + test only for some of the architectures we support in each language
- [ ] Build (but not run) _some_ examples pulled from the live tests (requires adapting our actions)
- [ ] Add an action that can build and run those examples as a manual trigger using `workflow_dispatch`

Further:
- [ ] Look into adding a Regtest setup to bring back some "live" tests but on Regtest

ItoroD:
Wondering what the deciding factor for which architecture to run on will be

ItoroD:
> Sounds good to me as a first step!
> 
> I also think in general that the "live tests" don't all need to be ported/transformed into examples (too redundant), and so it makes more sense to rather start fresh on the examples side. We also don't need to duplicate the work/examples that are in the bookofbdk ([examples here](https://github.com/bitcoindevkit/book-of-bdk/tree/master/examples)), so it's a tricky balance.

Alright, sounds good. Each language folder should have their own example folder then rather than global example folder with language subfolders. Will help for pipeline actions.

> 
> One small issue I have with the approach however is that it would be good to at least keep the option to run some live tests, so we can at least run them locally or trigger them on command in GitHub, just for sanity checks. With the `@ignore`, we might not be able to run them at all? Let me know if there is a way to do this cleanly.

[...]

PR #974, opened 2026-03-24
test: add offline wallet smoke tests

j-kon:
### Description

Add small Rust-side wallet smoke tests covering realistic wallet usage flows.

The tests focus on:
- creating a wallet
- revealing the next address
- checking initial balance
- verifying derivation index progression
- validating simple `create_single` behavior

### Notes to the reviewers

I updated the original wallet tests based on review feedback:
- removed the incorrect `create_single` external/internal keychain comparison
- dropped the `_offline` suffix from test names
- reworked the tests to read more like usage flows rather than isolated assertions
- explicitly verified derivation index behavior before and after the first reveal

Scope is limited to Rust wallet tests.

### Documentation

  - [`Wallet`](https://docs.rs/bdk_wallet/latest/bdk_wallet/struct.Wallet.html)
- [x] [`uniffi`](https://github.com/mozilla/uniffi-rs)
- [x] Other:
  - [`bdk-ffi` `Wallet::create_single` docs](https://github.com/bitcoindevkit/bdk-ffi/blob/master/bdk-ffi/src/wallet.rs)
  - [`bdk-jvm` bindings-style test example](https://github.com/bitcoindevkit/bdk-jvm/blob/master/lib/src/test/kotlin/org/bitcoindevkit/LoadWalletTest.kt)
  - [`bdk-ffi` issue: rethinking our test suite](https://github.com/bitcoindevkit/bdk-ffi/issues/769)

#### All Submissions:

#### New Features:

#### Bugfixes:

cc @thunderbiscuit and @reez

### DIFF
--- a/bdk-ffi/src/tests/wallet.rs
+++ b/bdk-ffi/src/tests/wallet.rs
@@ -1,0 +1,107 @@
+use crate::bitcoin::Network;
+use crate::descriptor::Descriptor;
+use crate::store::Persister;
+use crate::wallet::Wallet;
+
+use bdk_wallet::KeychainKind;
+
+use std::sync::Arc;
+
+const EXTERNAL_DESCRIPTOR: &str = "wpkh(tprv8ZgxMBicQKsPf2qfrEygW6fdYseJDDrVnDv26PH5BHdvSuG6ecCbHqLVof9yZcMoM31z9ur3tTYbSnr1WBqbGX97CbXcmp5H6qeMpyvx35B/84h/1h/1h/0/*)";
+const INTERNAL_DESCRIPTOR: &str = "wpkh(tprv8ZgxMBicQKsPf2qfrEygW6fdYseJDDrVnDv26PH5BHdvSuG6ecCbHqLVof9yZcMoM31z9ur3tTYbSnr1WBqbGX97CbXcmp5H6qeMpyvx35B/84h/1h/1h/1/*)";
+const TWO_PATH_DESCRIPTOR: &str = "wpkh([9a6a2580/84'/1'/0']tpubDDnGNapGEY6AZAdQbfRJgMg9fvz8pUBrLwvyvUqEgcUfgzM6zc2eVK4vY9x9L5FJWdX8WumXuLEDV5zDZnTfbn87vLe9XceCFwTu9so9Kks/<0;1>/*)";
+const EXPECTED_FIRST_ADDRESS: &str = "tb1qhjys9wxlfykmte7ftryptx975uqgd6kcm6a7z4";
+
+fn external_descriptor() -> Arc<Descriptor> {
+    Arc::new(Descriptor::new(EXTERNAL_DESCRIPTOR.to_string(), Network::Signet).unwrap())
+}
+
+fn internal_descriptor() -> Arc<Descriptor> {
+    Arc::new(Descriptor::new(INTERNAL_DESCRIPTOR.to_string(), Network::Signet).unwrap())
+}
+
+fn two_path_descriptor() -> Arc<Descriptor> {
+    Arc::new(Descriptor::new(TWO_PATH_DESCRIPTOR.to_string(), Network::Signet).unwrap())
+}
+
+fn build_wallet() -> Wallet {
+    Wallet::new(
+        external_descriptor(),
+        internal_descriptor(),
+        Network::Signet,
+        Arc::new(Persister::new_in_memory().unwrap()),
+        25,
+    )
+    .unwrap()
+}
+
+#[test]
+fn test_create_wallet() {
+    let wallet = build_wallet();
+
+    assert_eq!(wallet.network(), Network::Signet);
+    assert_eq!(wallet.balance().total.to_sat(), 0u64);
+    assert!(wallet.list_unspent().is_empty());
+    assert_eq!(wallet.derivation_index(KeychainKind::External), None);
+    assert_eq!(wallet.next_derivation_index(KeychainKind::External), 0);
+}
+
+#[test]
+fn test_reveal_next_address() {
+    let wallet = build_wallet();
+
+    let address_info = wallet.reveal_next_address(KeychainKind::External);
+
+    assert_eq!(address_info.index, 0);
+    assert_eq!(address_info.keychain, KeychainKind::External);
+    assert_eq!(address_info.address.to_string(), EXPECTED_FIRST_ADDRESS);
+}
+
+#[test]
+fn test_create_single_wallet() {
+    let wallet = Wallet::create_single(
+        external_descriptor(),
+        Network::Signet,
+        Arc::new(Persister::new_in_memory().unwrap()),
+        25,
+    )
+    .unwrap();
+
+    assert_eq!(wallet.derivation_index(KeychainKind::External), None);
+
+    let address_info = wallet.reveal_next_address(KeychainKind::External);
+
+    assert_eq!(address_info.index, 0);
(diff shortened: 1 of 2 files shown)


bdk-ffi/src/tests/wallet.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    assert_eq!(address_info.index, 0);
+    assert_eq!(address_info.keychain, KeychainKind::External);
+    assert_eq!(address_info.address.to_string(), EXPECTED_FIRST_ADDRESS);
+    assert_eq!(wallet.next_derivation_index(KeychainKind::External), 1);
+}
+
+#[test]
+fn test_single_descriptor_wallet_external_and_internal_peek_match() {
+    let wallet = build_single_descriptor_wallet();
```
ItoroD: Hi @j-kon! Thanks for the test.

Since you are doing a `create_single` here maybe we do not want to use a second KeychainKind for comparison. See doc for [create_single](https://github.com/bitcoindevkit/bdk-ffi/blob/a5918b018d14511055717a63ccb47f45b2b38457/bdk-ffi/src/wallet.rs#L83).

bdk-ffi/src/tests/wallet.rs
```diff
@@ -0,0 +1,71 @@
...
+
+#[test]
+fn test_new_wallet_starts_with_zero_balance() {
+    let wallet = build_wallet();
+
+    assert_eq!(wallet.balance().total.to_sat(), 0u64);
+}
+
+#[test]
+fn test_reveal_next_address_offline() {
```
ItoroD: Small nit:

Maybe we drop the offline suffix from some of the function names?

ItoroD (changes requested):
Thanks for the test! We have also being leaning more towards adding some tests on the bindings side which are a little more comprehensive. See example [here](https://github.com/bitcoindevkit/bdk-jvm/blob/master/lib/src/test/kotlin/org/bitcoindevkit/LoadWalletTest.kt). The addition of a [regtest ](https://github.com/bitcoindevkit/bdk-jvm/pull/46/changes) setup into our CI for the various bindings also allows us simulate some "live" test. Currently we have that for bdk-jvm and android, hopefully others will have have that too. Here is an [issue ](https://github.com/bitcoindevkit/bdk-ffi/issues/769) we previously had just for context.

ItoroD:
@j-kon will also be good if you could check the checklist in PR description. And a signed commit as well

j-kon:
Thanks for the detailed feedback, really helpful.

I see your point about `create_single` and the KeychainKind usage, I’ll adjust the test to better reflect single-descriptor behavior.

Also makes sense on leaning toward bindings-side tests. I’ll review the bdk-jvm example you shared and align with that direction.

I’ll clean up the naming, checklist, and signed commit as well.

j-kon:
Thanks for the feedback.

I updated the wallet tests to align more closely with the direction you suggested.

Changes made:
- removed the incorrect `create_single` external/internal keychain comparison
- dropped the `_offline` suffix from the test names
- reworked the tests into simple wallet usage flows
- explicitly verified derivation index behavior before and after the first reveal
- kept the scope limited to `bdk-ffi/src/tests/wallet.rs`

Verification:
- `cargo fmt`
- `cargo test wallet`
- `cargo clippy --all-targets --all-features`

The PR now includes these updates.

bdk-ffi/src/tests/wallet.rs
```diff
@@ -33,39 +31,53 @@ fn build_single_descriptor_wallet() -> Wallet {
...
 
-#[test]
-fn test_reveal_next_address_offline() {
-    let wallet = build_wallet();
     let address_info = wallet.reveal_next_address(KeychainKind::External);
 
     assert_eq!(address_info.index, 0);
     assert_eq!(address_info.keychain, KeychainKind::External);
     assert_eq!(address_info.address.to_string(), EXPECTED_FIRST_ADDRESS);
+    assert_eq!(wallet.derivation_index(KeychainKind::External), Some(0));
```
ItoroD: Maybe the asserts should only have to do with address_info? I think line 52 to 54 are good for this test. line 55 and 56 already show up in the other 2 tests its best to leave them out of here IMO

ItoroD (commented):
Thanks for the updates. For completeness I think there is another create method for wallet `create_from_two_path_descriptor` maybe we can add that too. Also, I think your commit is still not signed.

j-kon:
Thanks for the follow-up.

I pushed another small update to keep the `create_single` test focused on the returned `address_info`, and added a separate `create_from_two_path_descriptor` usage-flow test so the constructor coverage stays split by wallet shape.

Verification:
- `cargo fmt --check`
- `cargo test wallet`

Scope is still limited to `bdk-ffi/src/tests/wallet.rs`.

j-kon:
I force-pushed a signed commit for the latest change set, and GitHub is now recognizing the current PR head commit as verified.

bdk-ffi/src/tests/wallet.rs
ItoroD: @j-kon I believe the lines I mentioned here were in test_reveal_next_address not create_single test. You seem to have changed create_single instead.

ItoroD:
> I force-pushed a signed commit for the latest change set, and GitHub is now recognizing the current PR head commit as verified.

Cool. Maybe you could also squash your commits. Typically, commit messages do not contain the details of requested changes by the reviewer. You want the commit message describing what you are adding just like the message of your first commit. You might find it easier amending your first commit and doing a force push when reviews require you to make a change, instead of sending different commits with the different changes.

j-kon:
> > I force-pushed a signed commit for the latest change set, and GitHub is now recognizing the current PR head commit as verified.
> 
> Cool. Maybe you could also squash your commits. Typically, commit messages do not contain the details of requested changes by the reviewer. You want the commit message describing what you are adding just like the message of your first commit. You might find it easier amending your first commit and doing a force push when reviews require you to make a change, instead of sending different commits with the different changes.

Thanks, that makes sense.

I understand. I’ll squash the PR history down to a single signed commit and keep the commit message focused on the actual addition rather than the review iterations.

j-kon:
Thanks, I squashed the PR down to a single signed commit and kept the commit message focused on the actual change:

`test: add offline wallet smoke tests`

ItoroD:
@j-kon FYI 

I will look at this today. Sorry for the delay

j-kon:
Thanks for catching that you were right, I had applied that change to the wrong test.

I’ve updated it so `test_reveal_next_address` now only checks the returned `address_info`, and I moved the derivation-state assertions back to `test_create_single_wallet`.

I force-pushed the update, and the PR is still a single signed commit.

ItoroD (approved):
ACK ba7a31fd60a1df5d02a7c9ba42a0dd1226ba341f

reez (approved):
ACK ba7a31fd60a1df5d02a7c9ba42a0dd1226ba341f

thunderbiscuit (approved):
ACK ba7a31fd60a1df5d02a7c9ba42a0dd1226ba341f.

Would you mind rebasing this on master? Then it's ready to merge!

j-kon:
I have rebased it.

thunderbiscuit:
@j-kon in this case you merged the changes on master rather than rebased. Once thing you can do now is squash the merge and your commit into one, so as to remove the merge commit.

j-kon:
Rebased and force-pushed. The PR is back to a single signed commit on top of current `master`, with the merge commit removed.

PR #978, opened 2026-04-16
chore: bump bdk_wallet 3.0.0 and bdk_kyoto 0.16.0

reez:
### Description

Bumping to 3.0 (and compatibility changes)

### Notes to the reviewers

The smallest first unit of work from https://github.com/bitcoindevkit/bdk-ffi/pull/971 ... after which all other 3.0 related PR's can be added to.

### Documentation

- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/types.rs
+++ b/bdk-ffi/src/types.rs
@@ -23,13 +23,14 @@ use bdk_wallet::descriptor::policy::{
     Condition as BdkCondition, PkOrF as BdkPkOrF, Policy as BdkPolicy,
     Satisfaction as BdkSatisfaction, SatisfiableItem as BdkSatisfiableItem,
 };
-use bdk_wallet::event::WalletEvent as BdkWalletEvent;
+use bdk_wallet::locked_outpoints::ChangeSet as BdkLockedOutpointsChangeSet;
 #[allow(deprecated)]
 use bdk_wallet::signer::{SignOptions as BdkSignOptions, TapLeavesOptions};
 use bdk_wallet::AddressInfo as BdkAddressInfo;
 use bdk_wallet::Balance as BdkBalance;
 use bdk_wallet::LocalOutput as BdkLocalOutput;
 use bdk_wallet::Update as BdkUpdate;
+use bdk_wallet::WalletEvent as BdkWalletEvent;
 
 use std::collections::{BTreeMap, BTreeSet, HashMap};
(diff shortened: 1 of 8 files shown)


bdk-ffi/src/types.rs
```diff
@@ -1307,13 +1310,21 @@ impl From<ChangeSet> for bdk_wallet::ChangeSet {
...
+            .map(|(outpoint, is_locked)| (outpoint.outpoint().into(), is_locked))
+            .collect();
         Self {
             descriptor,
             change_descriptor,
             network,
             local_chain,
             tx_graph,
             indexer,
+            locked_outpoints: bdk_wallet::locked_outpoints::ChangeSet {
```
thunderbiscuit: The other fields (tx_graph, local_chain) have their own `From` trait implemented, making the code here look seamless (`value.indexer.into()`, `value.tx_graph.into()`). I was wondering why locked_outpoint was different and I think it doesn't need to be; if you extract the logic from 1313-1317 into the From trait we should be able to just call value.locked_outpoints.into() like we do for the other fields.

thunderbiscuit: This will also I think clean up the call site here on line 1325.

bdk-ffi/src/types.rs
```diff
@@ -1336,13 +1347,20 @@ impl From<bdk_wallet::ChangeSet> for ChangeSet {
         let local_chain = value.local_chain.into();
         let tx_graph = value.tx_graph.into();
         let indexer = value.indexer.into();
+        let locked_outpoints = value
```
thunderbiscuit: Similar comment to above, but the conversion would happen the other way.

thunderbiscuit (changes requested):
Just one little thing I think could use a small cleanup (and it's more of a look/consistency comment rather than logic issue so could be fixed later if needed), otherwise ready to go.

reez:
> Just one little thing I think could use a small cleanup (and it's more of a look/consistency comment rather than logic issue so could be fixed later if needed), otherwise ready to go.

Good call on those. Will update

reez:
Updated based on feedback ✅

thunderbiscuit (approved):
ACK 7deab3dab84049d52c89474aa1d65a46f80aa5b8. A small step for reez, a big step for BDK!

PR #979, opened 2026-04-20
feat(kyoto): Add `peer_info` method for IPv4/6 peer connections

rustaceanrob:
This is a useful method for wallets that can display where nodes are located geographically, which is a neat UI feature. Other connection types could potentially be supported in the future, but for now this returns the current list of IPv4/6 connections, which is the overwhelming majority of nodes.

### DIFF
--- a/bdk-ffi/src/kyoto.rs
+++ b/bdk-ffi/src/kyoto.rs
@@ -359,6 +359,25 @@ impl CbfClient {
             .map(|ip| Arc::new(IpAddress { inner: ip }))
             .collect()
     }
+
+    /// Get the list of current connections.
+    pub async fn peer_info(&self) -> Result<Vec<Arc<IpAddress>>, CbfError> {
+        let peers = self
+            .sender
+            .peer_info()
+            .await
+            .map_err(|_| CbfError::NodeStopped)?;
+        Ok(peers
+            .into_iter()
+            .filter_map(|(ip, _)| match ip {
+                AddrV2::Ipv4(ip) => Some(IpAddr::V4(ip)),
+                AddrV2::Ipv6(ip) => Some(IpAddr::V6(ip)),
+                _ => None,
+            })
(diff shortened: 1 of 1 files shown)


bdk-ffi/src/kyoto.rs
```diff
@@ -359,6 +359,25 @@ impl CbfClient {
...
+    /// Get the list of current connections.
+    pub async fn peer_info(&self) -> Result<Vec<Arc<IpAddress>>, CbfError> {
+        let peers = self
+            .sender
+            .peer_info()
+            .await
+            .map_err(|_| CbfError::NodeStopped)?;
+        Ok(peers
+            .into_iter()
+            .filter_map(|(ip, _)| match ip {
```
reez, line 372: What's the reason we’re only exposing `AddrV2::Ipv4/Ipv6` here and not `TorV3` as well (I think bdk-ffi and upstream have support, I def could be missing something obvious though)?

rustaceanrob: Although there is support via Socks5, the Tor addresses are advertised and represented as a `[u8; 32]` public key in the P2P network, which would require a newtype in FFI as far as I understand. To derive the onion address from a public key we also need Sha3-256 and Base32 encoding, see this part the [Socks5 proxy](https://github.com/2140-dev/kyoto/blob/master/src/network/socks.rs#L32). Rather than changing `IpAddress`, I figure the extra complication isn't worth it. I would think the vast majority of connections will be IPv4/6 even if the connection is over Tor.

reez: Ah yeah, makes sense to me

reez (approved):
ACK 18b2cc0c1cbaba6088df771eebe7fce4d5884b8b

rustaceanrob:
1d4180ec4aeb8e5bd6c27def280e3b49f5a68592 rebased

Issue #818, opened 2025-07-29
Add examples module to all bindings


ItoroD:
It will be great for users of the library or anyone at all to see how some of the APIs can be used in our various language bindings. This has already been started for the JVM bindings [here](https://github.com/bitcoindevkit/bdk-ffi/pull/793). A similar thing can be done for other bindings.

Note
---
- Our various bindings are being moved to their own repository, so it is best to add the examples module in the new repositories

staru09:
Is this open to work?

thunderbiscuit:
Yes it is @staru09! Would you be interested in the Swift examples or the Android ones (or both)?

thunderbiscuit:
Yep if you'd like to refactor the Python ones that'd be a great place to start. You can find the Python code [here at bdk-python](https://github.com/bitcoindevkit/bdk-python).

You can start with [the examples we have in bdk-jvm](https://github.com/bitcoindevkit/bdk-jvm/tree/master/examples/src/main/kotlin) if you'd like.

RowDaBoat:
Hi, new guy here. Has anyone taken up the Swift examples yet?
I've been playing around and got a `main.swift` mirroring the `WalletSetupBip32` example from `bdk-jvm`.

Should I do a PR including the other examples too?

reez:
> Hi, new guy here. Has anyone taken up the Swift examples yet? I've been playing around and got a `main.swift` mirroring the `WalletSetupBip32` example from `bdk-jvm`.
> 
> Should I do a PR including the other examples too?

sounds great! if you want to do one PR with the Swift mirror of the https://github.com/bitcoindevkit/bdk-ffi/pull/793/changes bdk-jvm that would be a nice PR

RowDaBoat:
Awesome. I started from there, then I saw it got some minor updates, so I took the latest version.
I'll be sending the PR tomorrow to `bdk-swift`.

reez:
> Awesome. I started from there, then I saw it got some minor updates, so I took the latest version. I'll be sending the PR tomorrow to `bdk-swift`.

If it's a PR for Swift, send the PR to bdk-ffi since we still have bdk-swift inside of bdk-ffi. bdk-jvm we spun out into a separate repo in the last year so that's why the examples for those will stay in a separate repo from bdk-ffi

Thanks for working on this!

RowDaBoat:
Hello again, I've also translated the `MultisigTransaction` example to `swift` (and will probably do the rest too).
[...]

PR #981, opened 2026-04-22
Add a Swift version of WalletSetupBip32, MultsigTransaction and MultisigTaprootTransaction examples

RowDaBoat:
### Description

Showing users how the APIs can be used across languages is valuable. This PR mirrors the WalletSetupBip32 JVM example in Swift as requested in [issue 818](https://github.com/bitcoindevkit/bdk-ffi/issues/818).

### Notes to the reviewers

This Swift example follows the Kotlin example as closely as possible.

#### All Submissions:

* [x] I ran `swift build` and `swift run` before committing

#### New Features:

* [x] I've added the new feature to the CI build

### DIFF
--- a/.github/workflows/test-swift.yaml
+++ b/.github/workflows/test-swift.yaml
@@ -42,3 +42,4 @@ jobs:
           swift test --filter Offline
           swift build --target WalletSetupBip32
           swift build --target MultisigTransaction
+          swift build --target MultisigTaprootTransaction
--- a/.github/workflows/test-swift.yaml
+++ b/.github/workflows/test-swift.yaml
@@ -41,3 +41,4 @@ jobs:
         run: |
           swift test --filter Offline
           swift build --target WalletSetupBip32
+          swift build --target MultisigTransaction
--- a/.github/workflows/test-swift.yaml
+++ b/.github/workflows/test-swift.yaml
@@ -38,4 +38,6 @@ jobs:
         run: bash ./build-xcframework.sh
 
       - name: "Run Swift tests"
-        run: swift test --filter Offline
+        run: |
+          swift test --filter Offline
+          swift build --target WalletSetupBip32
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ Info.plist
 bdkffi.xcframework
 Sources/
 xcuserdata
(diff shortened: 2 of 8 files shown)


j-kon (changes requested):
Thanks for adding this Swift example. I found one user-facing issue that should be fixed before this lands.

`bdk-swift/README.md` documents `swift run WalletSetupBip32Example`, but the executable added in `bdk-swift/Package.swift` is named `WalletSetupBip32`, and the CI change builds that same target. As written, the README command will fail because there is no executable product with that name.

I think the README command should be changed to `swift run WalletSetupBip32`.

Aside from that, this looks aligned with issue #818 and mirrors the JVM example closely.

j-kon (approved):
Thanks for the update. The README command now matches the Swift executable target, so my previous concern is resolved ACK.

j-kon (approved):
ACK

RowDaBoat:
I've added swift versions of `MultisigTransaction` and `MultisigTaprootTransaction`.

reez:
Swift code itself looks good to me, so I will comment on that

This would need feedback from the other folks who opened/commented on #818

thunderbiscuit:
Sorry I've been going all out on the 3.0.0 release, and this was lower priority.

My feeling about this (I'm sorry this might appear counter to what is outlined in #818, and I'll have to update that issue probably), is that examples like this add a significant amount of code (about 700 lines in this case), and that code must have a clear owner(s), otherwise it's just bloat that no one is really interested in maintaining.

In the case of Android I am this person, and on the JVM side @ItoroD kind of is (correct me if I'm wrong Itoro). Here I don't see one directly, and after discussing it with the other guys over dev calls I don't think we have enough interest in this to warrant merging.

This is not saying your code isn't good or we don't like the contribution, but just that I want to balance the scope of what we maintain with the resources we have to maintain it.

RowDaBoat:
That's alright, it's absolutely reasonable.

~~Would having at least _some_ kind of short sample on Swift's README.md (or any other for that matter) make sense? If so, I can craft it on a new PR, if not that's ok anyways.~~
EDIT: just read Steve about contributing with documentation.

In any case, just close this one.

PR #982, opened 2026-04-24
test: remove noisy print statements

j-kon:
## Summary

Remove unconditional debug/print output from a small set of Rust test files where the assertions already provide enough signal on failure.

## Scope

This change is intentionally limited to test cleanup in:

- `bdk-ffi/src/tests/descriptor.rs`
- `bdk-ffi/src/tests/keys.rs`
- `bdk-ffi/src/tests/bitcoin.rs`

I left `tx_builder.rs` unchanged in this pass because those prints are mixed into live-network and failure-diagnostic paths, so I wanted to keep the scope conservative.

## What changed

- removed `println!` calls from descriptor template/address-data/psbt-related tests where they were only adding noise on success
- removed `dbg!` calls from key tests
- kept test behavior the same
- replaced one print-driven sanity check in `test_to_address_data` with explicit variant assertions

## Verification

- `cargo fmt --check`
- `cargo test descriptor`
- `cargo test psbt`
- `cargo test test_to_address_data`
- `cargo test test_from_str_inner`

### DIFF
--- a/bdk-ffi/src/tests/bitcoin.rs
+++ b/bdk-ffi/src/tests/bitcoin.rs
@@ -1,4 +1,4 @@
-use crate::bitcoin::{Address, Key, Network, ProprietaryKey, Psbt};
+use crate::bitcoin::{Address, AddressData, Key, Network, ProprietaryKey, Psbt};
 use bdk_electrum::bdk_core::bitcoin::hex::DisplayHex;
 
 #[test]
@@ -335,7 +335,6 @@ fn test_to_address_data() {
     )
     .unwrap();
     let p2pkh_data = p2pkh.to_address_data();
-    println!("P2PKH data: {:#?}", p2pkh_data);
 
     // P2SH address
     let p2sh = Address::new(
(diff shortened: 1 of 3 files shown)


bdk-ffi/src/tests/bitcoin.rs
```diff
@@ -374,18 +373,16 @@ fn test_psbt_spend_utxo() {
 fn test_psbt_input_length() {
     let psbt = sample_psbt2();
     let psbt_inputs = psbt.input();
-    println!("Psbt Input: {:?}", psbt_inputs);
 
     let unknown = &psbt_inputs[0].unknown;
-    println!("Unknown field in Psbt Input: {:?}", unknown);
     assert_eq!(psbt_inputs.len(), 1);
+    assert_eq!(unknown.len(), 1);
```
thunderbiscuit: I'm not sure why we're testing the length of the `unknown` field here but let's remove this line while we're at it!

thunderbiscuit (changes requested):
One tiny fix I'd like to see and this is ready to go! Thanks for working on this.

j-kon:
Thanks, I removed that extra `unknown.len()` assertion from `test_psbt_input_length` and force-pushed the update.

I re-ran:
- `cargo fmt --check`
- `cargo test test_psbt_input_length`
- `cargo test psbt`

thunderbiscuit (approved):
ACK f8f8bfd45c934445b97381d44a0af06a10e74feb.

PR #994, opened 2026-05-12
Update `bdk_kyoto` to `0.17.0`

rustaceanrob:
With the release of `bitcoin = 0.32.9` there was a backport that unexplainably broke the P2P layer of Kyoto. This fixes the `bitcoin` version at `0.32.8` in the meantime while I investigate the behavior.

PR #996, opened 2026-05-12
Update `bdk_kyoto` to `0.17.0` on `3.0.0` release branch

rustaceanrob:
Cherry pick of #994 onto `3.0.0`

### DIFF
--- a/bdk-ffi/src/kyoto.rs
+++ b/bdk-ffi/src/kyoto.rs
@@ -1,3 +1,4 @@
+use bdk_kyoto::bip157;
 use bdk_kyoto::bip157::lookup_host;
 use bdk_kyoto::bip157::tokio;
 use bdk_kyoto::bip157::AddrV2;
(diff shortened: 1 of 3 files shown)


bdk-ffi/src/kyoto.rs
```diff
@@ -302,9 +303,9 @@ impl CbfClient {
 
     /// Broadcast a transaction to the network, erroring if the node has stopped running.
     pub async fn broadcast(&self, transaction: &Transaction) -> Result<Arc<Wtxid>, CbfError> {
```
thunderbiscuit, line 305: If the method is `submit_package` on the client, we might want to call it that here as well. I know we've had these issues with the Esplora and Electrum clients before where the naming is not standardized, and sticking to the clients API naming is what we have done in the past.

thunderbiscuit: In any case this is 3.0 so it's the right time to break the API if we want to. I'm not leaning one way or another in this case; @rustaceanrob what you do you think?

rustaceanrob: If we change the name I would prefer to roll out the full CPFP API around it, as in we implement a one-parent-one-child type and allow it to be submitted here. I don't think I would have that ready in time for `3.0.0` nor do I think its very pressing for FFI. `broadcast_tx` is a better description at the moment IMO

thunderbiscuit: All right let's do that then.

thunderbiscuit (approved):
ACK 918efa7d45dbc4edd2e80b3a91f6e7be3a776c79.

PR #997, opened 2026-05-12
ci permissions

reez:
### Notes to the reviewers

Related https://github.com/bitcoindevkit/bdk/issues/1637

We should also check our downstream libs (python, jvm, dart, rn)

### Documentation

- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/.github/workflows/audit.yaml
+++ b/.github/workflows/audit.yaml
@@ -13,6 +13,8 @@ on:
   schedule:
     - cron: '0 0 * * 0' # Once per week
 
+permissions: {}
+
 jobs:
   security-audit:
     name: "Security audit"
--- a/.github/workflows/rust-ci.yaml
+++ b/.github/workflows/rust-ci.yaml
@@ -31,6 +31,8 @@ jobs:
 
       - name: "Set up Rust"
         uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          cache: false
 
       - name: "Check Cargo.lock"
         run: |
@@ -60,6 +62,7 @@ jobs:
 
       - uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
+          cache: false
           components: clippy, rustfmt
 
       - name: "Check fmt"
--- a/.github/workflows/test-android.yaml
+++ b/.github/workflows/test-android.yaml
@@ -59,6 +59,8 @@ jobs:
 
       - name: "Set up Rust"
         uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          cache: false
 
       - name: "Build Android library"
         run: |
--- a/.github/workflows/test-swift.yaml
+++ b/.github/workflows/test-swift.yaml
@@ -33,6 +33,8 @@ jobs:
 
       - name: "Set up Rust"
         uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          cache: false
 
       - name: "Build Swift package"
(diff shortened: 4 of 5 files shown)


ItoroD:
Any measurement of the time we save removing the cache?

thunderbiscuit (approved):
I didn't know but the Rust caching stuff on CI is quite the rabbit hole!. 

In our case however, I think a bunch of stuff is going on:

1. When Steve opened the issue bitcoindevkit/bdk#1637 about slow caching on the bdk repo, they were using a different action back then. All the repos use the newer actions-rust-lang/setup-rust-toolchain@v1 now.
2. Apparently caching is sometimes good sometimes bad for speed? It's an extra step and so if you end up not needing the cache because your dependencies change, it's a cost to pay for nothing.
3. Looking at one of our workflows (Test Android), the cache actually doesn't hit anything! This is because the action looks at the root of the repo and doesn't find a `Cargo.toml` file, so the whole thing fails anyway haha. I don't actually know whether that's the case for all worfklows but it very well might be.
4. In other words I think until further digging and exploration, turning the cache off exactly like you have it in the PR is probably a good default for us. If someone wants to dig into it and optimize our CIs to use it (I think it's possible to do by telling the action where exactly to find the Cargo.toml but needs more testing) they can do it, and we can review then. For now I'm tempted to go ahead with what you have.
5. I agree with @ItoroD it would have been nice to see timed runs to get a feel for the time savings. But also now that I think the cache doesn't work for us I assume it's not a meaningful metric.
6. Needs rebase.

ACK b5fba8d5dd135a4d1a30877522cce3535e35a676.

reez:
> I didn't know but the Rust caching stuff on CI is quite the rabbit hole!.
> 
> In our case however, I think a bunch of stuff is going on:
> 
> 1. When Steve opened the issue [bitcoindevkit/bdk#1637](https://github.com/bitcoindevkit/bdk/issues/1637) about slow caching on the bdk repo, they were using a different action back then. All the repos use the newer actions-rust-lang/setup-rust-toolchain@v1 now.
> 2. Apparently caching is sometimes good sometimes bad for speed? It's an extra step and so if you end up not needing the cache because your dependencies change, it's a cost to pay for nothing.
> 3. Looking at one of our workflows (Test Android), the cache actually doesn't hit anything! This is because the action looks at the root of the repo and doesn't find a `Cargo.toml` file, so the whole thing fails anyway haha. I don't actually know whether that's the case for all worfklows but it very well might be.
> 4. In other words I think until further digging and exploration, turning the cache off exactly like you have it in the PR is probably a good default for us. If someone wants to dig into it and optimize our CIs to use it (I think it's possible to do by telling the action where exactly to find the Cargo.toml but needs more testing) they can do it, and we can review then. For now I'm tempted to go ahead with what you have.
> 5. I agree with @ItoroD it would have been nice to see timed runs to get a feel for the time savings. But also now that I think the cache doesn't work for us I assume it's not a meaningful metric.
> 6. Needs rebase.
> 
> ACK [b5fba8d](https://github.com/bitcoindevkit/bdk-ffi/commit/b5fba8d5dd135a4d1a30877522cce3535e35a676).

Yeah, since our Rust crate lives under bdk-ffi/ the current default cache setup is not a great baseline to benchmark against. I think disabling it is the right simple default for now and then a followup could configure rust-src-dir: bdk-ffi / cache-workspaces or something and measure that against no cache.

PR #1000, opened 2026-05-12
This move is chaotic good

thunderbiscuit:


reez (commented):
Good rename, but there are still local path references to the old directory names (CI is failing because workflows still use paths like `bdk-ffi`, `bdk-android`, and `bdk-swift` as working directories / `cd` targets)

There might be some other ones too I haven't caught yet like workflows, CODEOWNERS, build scripts, and docs? Obv some names should probably stay as package/repo names, but local filesystem paths would to move to `rust`, `kotlin`, and `swift` etc

thunderbiscuit:
If we agree this is a good idea, I propose we merge right after the 3.1 release, to give plenty of time for all downstream users to update their build workflows and CI.

Issue #983, opened 2026-04-29
Migration to uniffi `0.31.1`


thunderbiscuit:
I'd like to make sure we don't upgrade here if it will create problems for downstream libraries (bdk-dart, bdk-rn, bdk-python, bdk-jvm, etc.)

reez:
I think we are going to be good on the dart side https://github.com/Uniffi-Dart/uniffi-dart/releases/tag/v0.2.0%2Bv0.31.1

but I will still do some further testing and research

thunderbiscuit:
Done!

PR #1001, opened 2026-05-12
Bump uniffi to 0.31.1

thunderbiscuit:
Initially this PR also updated the Cargo.lock file at the same time through a simple `cargo update`, but that picks up on a ton of new stuff, which conflicts with #1015. Not sure how best to handle it. I can of course just push it all in one go here, but also it feels like we're always going to have these weird updates happening where you just bump one thing and then a million others bump at the same time because your cargo update is so coarse. Will take a look at simplifying this.

Edit: the key is to run `cargo udpate --package <crate>`, which forces the Cargo.lock to only bump the dependency you updated in your Cargo.toml file.

See #983.

### Documentation

- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/Cargo.lock
+++ b/bdk-ffi/Cargo.lock
@@ -34,9 +34,9 @@ checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
 name = "askama"
-version = "0.13.1"
+version = "0.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5d4744ed2eef2645831b441d8f5459689ade2ab27c854488fbab1fbe94fce1a7"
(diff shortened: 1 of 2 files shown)


ItoroD:
Seems safe. Should expect similar for JVM. Aiming for 3.0 or after?

thunderbiscuit:
Yeah I was thinking of doing the 3.0 on the current uniffi just to ensure we don't break anything for anyone and the bumps goes smoothly for all downstream bindings, but if we have the green light from everyone else (I think dart is the one that we still need to confirm there) then I'd bump on master so that we get it for the 3.1 release.

reez (approved):
ACK e68c53cb632263f739451c8a72cae77f9c8dbcd3

Wondering if when it gets rebased our new merged Cargo.lock thing will trigger or not

thunderbiscuit:
Success! https://github.com/bitcoindevkit/bdk-ffi/actions/runs/26182549292/job/77029253459

PR #1010, opened 2026-05-21
Electrum params

reez:
### Description

Adds optional timeout and retry args to ElectrumClient::new, matching electrum-client::ConfigBuilder 

Defaults remain unchanged when omitted: timeout = None, and retry keeps upstream’s default.

### Documentation

- [ ] [`bdk_wallet`](https://docs.rs/bdk_wallet/latest/bdk_wallet/)

https://docs.rs/electrum-client/0.24.1/electrum_client/struct.ConfigBuilder.html#method.timeout

https://docs.rs/electrum-client/0.24.1/electrum_client/struct.ConfigBuilder.html#method.retry

- [ ] [`bitcoin`](https://docs.rs/bitcoin/latest/bitcoin/index.html)
  
- [ ] [`uniffi`](https://github.com/mozilla/uniffi-rs) 
- [ ] Other: 

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [ ] I ran `cargo fmt` and `cargo clippy` before committing
* [ ] I've added exactly one `changelog:*` label
* [ ] I've linked the relevant upstream docs or specs above

#### New Features:

* [ ] I've added tests for the new feature
* [ ] I've added docs for the new feature

#### Bugfixes:

* [ ] This pull request breaks the existing API
* [ ] I've added tests to reproduce the issue which are now passing
* [ ] I'm linking the issue being fixed by this PR

PR #1011, opened 2026-05-22
Electrum params (3.0)

reez:
### Description

https://github.com/bitcoindevkit/bdk-ffi/pull/1010 for 3.0 branch

### Documentation

- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/electrum.rs
+++ b/bdk-ffi/src/electrum.rs
@@ -19,6 +19,7 @@ use bdk_wallet::bitcoin::hex::{Case, DisplayHex};
 use std::collections::BTreeMap;
 use std::convert::TryFrom;
 use std::sync::Arc;
+use std::time::Duration;
 
 /// Wrapper around an electrum_client::ElectrumApi which includes an internal in-memory transaction
 /// cache to avoid re-fetching already downloaded transactions.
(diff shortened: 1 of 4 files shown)


thunderbiscuit:
@reez could you add a commit for the bump of bdk_electrum to 0.24.0 to this PR?

bdk-ffi/src/electrum.rs
```diff
@@ -29,15 +29,25 @@ pub struct ElectrumClient(BdkBdkElectrumClient<bdk_electrum::electrum_client::Cl
 impl ElectrumClient {
     /// Creates a new bdk client from a electrum_client::ElectrumApi
     /// Optional: Set the proxy of the builder
+    /// Optional: Set the timeout of the builder
```
thunderbiscuit: The Rust docs make the mistake of not defining what the unit is on the timeout, but I don't think we should make that same mistake because ffi users have an even harder time digging into the source code that is used to generate the bindings (you can't "command+click" into a type/method and see the Rust source).

I suggest you make the correction here and we open a PR on Electrum to fix it too.

```md
// Optional: Set the timeout (in seconds) of the builder
```

https://docs.rs/electrum-client/0.24.1/src/electrum_client/config.rs.html#57-61

thunderbiscuit (changes requested):

bdk-ffi/src/electrum.rs
thunderbiscuit: Ok actually this API changed in the 0.25.0 release, and the timeout method takes a `Duration` argument rather than a `u8`.

So there is no documentation problem on the Rust side anymore, but I suggest we keep the `u8` argument (and the docs mentioning this is defined in seconds) here and you transform it into a `Duration` in the loop, something like 

```rust
Duration::from_secs(timeout)
```

reez:
bdk_electrum updated & changelog reflected

thunderbiscuit (approved):
ACK 69c29f91bf5fad9b6afdec4e6b2441f79fe901cc.

PR #1012, opened 2026-05-26
Exploration: custom sync methods

thunderbiscuit:
This is WIP, but would basically allow you to sync specific indices on either of your keychains, or say "sync me the last x number of addresses on each keychain".

### Documentation

- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/wallet.rs
+++ b/bdk-ffi/src/wallet.rs
@@ -12,6 +12,7 @@ use crate::types::{
 };
 
 use bdk_wallet::bitcoin::Network;
+use bdk_wallet::chain::spk_client::SyncRequest as BdkSyncRequest;
 #[allow(deprecated)]
 use bdk_wallet::signer::SignOptions as BdkSignOptions;
 use bdk_wallet::{PersistedWallet, Wallet as BdkWallet};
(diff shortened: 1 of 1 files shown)


bdk-ffi/src/wallet.rs
```diff
@@ -664,6 +665,71 @@ impl Wallet {
...
+                    .spk_at_index(KeychainKind::External, i)
+                    .map(|spk| ((KeychainKind::External, i), spk))
+            })
+            .chain(internal_indices.into_iter().filter_map(|i| {
+                spk_index
+                    .spk_at_index(KeychainKind::Internal, i)
+                    .map(|spk| ((KeychainKind::Internal, i), spk))
+            }));
+
+        let builder = BdkSyncRequest::builder()
```
j-kon, line 694: Could we preserve the `expected_spk_txids` behavior that BDK Wallet’s normal `start_sync_with_revealed_spks{,_at}` path adds? Those expected txids are used by sync backends to detect previously-seen mempool txs that disappear from SPK history. As written, this custom `SyncRequest` builder only sets the chain tip and selected SPKs, so these selective sync APIs can silently miss eviction/replacement updates compared with the existing sync path.

bdk-ffi/src/wallet.rs
```diff
@@ -664,6 +665,71 @@ impl Wallet {
         Arc::new(SyncRequestBuilder(Mutex::new(Some(builder))))
     }
 
+    /// Create a partial [`SyncRequest`] for a specific set of derivation indices.
```
j-kon, line 668: Since this is a new public FFI API, could we add tests that reveal a known set of external/internal addresses and assert the request contains only the selected indices / last N revealed SPKs? That would also help guard this custom builder against drifting from BDK Wallet’s normal sync-request semantics.

PR #1013, opened 2026-05-27
Test v3 migration

thunderbiscuit:
This PR adds old databases (v0.32, v1, v2) to the Android assets, and tests the migration paths for each of those into a v3 wallet.

## Notes

You will notice that there is a bit of redundancy/crossover between the new `DatabaseVersionCompatTest` and the current `PersistenceTest`. As is, the `PersistenceTest` is really using v2 databases, and so I didn't want to mess with this and combine the two ideas, even though eventually I think we should clean this up. The tests in `PersistenceTest` are not wrong, because we know (from the new tests in `DatabaseVersionCompatTest`) that the v2 to v3 conversion is clean and seamless, and so it's all good. The only think I'd like to fix is that they don't merge the ideas of testing (1) using a v2 db into a v3 wallet, and (2) loading from persistence itself (single and double descriptor wallets, as well as the v0.32 migration).

### Documentation

- [ ] Other: 

#### All Submissions:


bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DatabaseVersionCompatTest.kt
```diff
@@ -0,0 +1,214 @@
...
+        val newV3DB = Persister.newInMemory()
+
+        val wallet = Wallet(
+            descriptor = descriptor,
+            changeDescriptor = changeDescriptor,
+            network = Network.REGTEST,
+            persister = newV3DB,
+        )
+
+        wallet.revealAddressesTo(KeychainKind.EXTERNAL, 6u)
```
reez: Should this use the preV1Keychains values we just read instead of hardcoding 6u and 0u here? I think that would make the test cover the migration path more directly (read old metadata, apply indices to new v3 wallet, then persist/reload + check that next external/internal addresses are 7 and 1).

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DatabaseVersionCompatTest.kt
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    // You can correctly migrate a v0.32 Wallet into a v3 Wallet
+    @Test
+    fun migrateToV3From032() {
+        val oldDB = Persister.newSqlite(resolveAsset("db_v032.sqlite3", "old_databases"))
+        val preV1Keychains: List<PreV1WalletKeychain> = oldDB.getPreV1WalletKeychains()
+
+        assertEquals(2, preV1Keychains.size)
+        assertEquals(KeychainKind.EXTERNAL, preV1Keychains.first().keychain)
```
reez: Tiny optional thing: should we look these up by keychain instead of relying on first() / last()? I don’t think the helper guarantees row ordering is the only reason I'm mentioning

bdk-android/lib/src/androidTest/assets/README.md
```diff
@@ -31,3 +32,12 @@ The wallet database `wallet_pre_v1.sqlite3` is used for testing. This wallet:
...
 - Has revealed the first address on the internal keychain (last revealed index is 0)
 - The descriptors are BIP86 descriptors with the MNEMONIC_AWESOME mnemonic
+
+## Old Databases
+
+The `old_databases` directory contains wallets created from different versions of BDK. These wallets:
+
+- Are Regtest wallets
+- Have revealed 7 addresses (0-6) on the external keychain (next address to come up is index 7)
+- Have revealed the first address (index 0) on the external keychain (next address to come up is index 1)
```
reez: nit/question: should this say internal keychain? The bullet above covers addresses 0-6, and the tests expect the next internal address to be index 1.

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DatabaseVersionCompatTest.kt
```diff
@@ -0,0 +1,214 @@
...
+
+        // println("V3 Database Tables: $tables")
+        // V3 Database Tables: [bdk_schemas, bdk_wallet, bdk_blocks, bdk_txs, bdk_txouts, bdk_anchors, bdk_descriptor_last_revealed, bdk_wallet_locked_outpoints, bdk_descriptor_derived_spks]
+
+        assertTrue(tables.contains("bdk_descriptor_derived_spks"))
+        assertTrue(tables.contains("bdk_wallet_locked_outpoints"))
+    }
+
+    // You can take a v2 Wallet database and use it to create a v3 Wallet, and the database migrates gracefully.
+    // The v3 database adds the bdk_descriptor_derived_spks table.
```
reez: nit: should this say v3 adds bdk_wallet_locked_outpoints here? The v2 already has bdk_descriptor_derived_spks and the assertion below checks for the locked outpoints.

reez (commented):
Thanks for adding these migration items and tests! 

Your note about the overlap with PersistenceTest makes sense to me, I think keeping DatabaseVersionCompatTest separate is fine for sure.

I left one question on the v0.32 migration test path, plus a couple of non blocking nits. 

Nothing else jumped out to me overall this looks good.

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DatabaseVersionCompatTest.kt
thunderbiscuit: Yes absolutely, that's an oversight for sure (hardcoding the correct answer into my test 😆)

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DatabaseVersionCompatTest.kt
thunderbiscuit: Good question about the ordering, now I'm curious. But yes good point let me do a lookup by keychain instead, much more robust.

bdk-android/lib/src/androidTest/assets/README.md
thunderbiscuit: I love your reviews cause I know you read every darn word I write

Good catch!

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DatabaseVersionCompatTest.kt
thunderbiscuit: Yep!

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DatabaseVersionCompatTest.kt
reez: Maybe something like this is what I was digging into a bit

```
val externalKeychain = preV1Keychains.single { it.keychain == KeychainKind.EXTERNAL }
val internalKeychain = preV1Keychains.single { it.keychain == KeychainKind.INTERNAL }
```

Then the assertions could use those named values. I like single here too because I think the test would also fail if ever had a missing or duplicate keychain?

But yeah totally optional and not asking for some massive refactor of anything, I'm good w what you got overall here.

bdk-android/lib/src/androidTest/assets/README.md
reez: > I love your reviews cause I know you read every darn word I write
> 
> Good catch!

I'd say the same thing about your reviews, appreciated.

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DatabaseVersionCompatTest.kt
thunderbiscuit: Ah this `single` function is actually even nicer than what I have, because it performs the check for nullity at the same time (that I currently have to do on a different line):

```kotlin
val externalPreV1Keychain = preV1Keychains.find { it.keychain == KeychainKind.EXTERNAL }
requireNotNull(externalPreV1Keychain) { "missing external pre-v1 keychain" }
```

Switching to the `single` API.

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DatabaseVersionCompatTest.kt
reez: Yep this covers the hardcoded part. Should we also persist and then reload from newV3DB before checking the next addresses? I think that would make this prove the migrated v3 persistence state too instead of only the live wallet after calling revealAddressesTo

thunderbiscuit: Makes sense. I applied this with the following:

```kotlin
val wallet = Wallet(
    descriptor = descriptor,
    changeDescriptor = changeDescriptor,
    network = Network.REGTEST,
    persister = newV3DB,
)

wallet.revealAddressesTo(KeychainKind.EXTERNAL, externalPreV1Keychain.lastDerivationIndex)
wallet.revealAddressesTo(KeychainKind.INTERNAL, internalPreV1Keychain.lastDerivationIndex)
wallet.persist(newV3DB)

val reloadedWallet = Wallet.load(
    descriptor = descriptor,
    changeDescriptor = changeDescriptor,
    persister = newV3DB,
)

val addressInfo: AddressInfo = reloadedWallet.revealNextAddress(KeychainKind.EXTERNAL)
val changeAddressInfo: AddressInfo = reloadedWallet.revealNextAddress(KeychainKind.INTERNAL)

assertEquals(addressInfo.index, 7u)
assertEquals(changeAddressInfo.index, 1u)
```

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/DatabaseVersionCompatTest.kt
```diff
@@ -0,0 +1,226 @@
...
+
+        assertEquals(2, preV1Keychains.size)
+        assertEquals(KeychainKind.EXTERNAL, externalPreV1Keychain.keychain)
+        assertEquals(KeychainKind.INTERNAL, internalPreV1Keychain.keychain)
+        assertEquals(6u, externalPreV1Keychain.lastDerivationIndex)
+        assertEquals(0u, internalPreV1Keychain.lastDerivationIndex)
+        assertEquals("rn0zejch", externalPreV1Keychain.checksum)
+        assertEquals("j82ry8g0", internalPreV1Keychain.checksum)
+
+        val newV3DBFilePath = context.getDatabasePath("new_v3_wallet.sqlite3")
```
reez, line 199: Could we make sure this test starts from a clean new_v3_wallet.sqlite3 database before creating the persister? Wallet(...) creates a new persisted wallet, and upstream returns DataAlreadyExists if the persister already contains wallet data, so repeated runs on the same app data can fail here before testing the migration path.

thunderbiscuit: Wow good catch. I didn't realize but on local tests reusing an old database already there could definitely be an issue. Fixed now with the following:

```kotlin
// This ensures local tests always create a new DB and don't reuse an old one leftover from prior tests
if (newV3DBFilePath.exists()) newV3DBFilePath.delete()
```

reez (approved):
ACK 21f8d55abd568eccc0c7a250d3ba0a5472a9eebd

LGTM!

PR #1014, opened 2026-05-27
Bump miniscript to 12.3.7 patch

thunderbiscuit:
This is just a bump of miniscript to the latest patch 12.3.7, picking up the bug fix outlined in bitcoindevkit/bdk_wallet#488.

## Changelog
```md
## Fixed
- Small bug in the Psbt::finalize implementation (rust-miniscript#958)
```

PR #1015, opened 2026-05-27
Bump miniscript to 12.3.7 patch

thunderbiscuit:
This is the sister PR to #1014. Issue bitcoindevkit/bdk_wallet#488.

### Changelog

```md
## Fixed
- Small bug in the Psbt::finalize implementation (rust-miniscript#958)
```

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/Cargo.lock
+++ b/bdk-ffi/Cargo.lock
@@ -799,9 +799,9 @@ checksum = "68354c5c6bd36d73ff3feceb05efa59b6acb7626617f4962be322a825e61f79a"
 
 [[package]]
 name = "miniscript"
-version = "12.3.6"
+version = "12.3.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c14116d8342edd3626b0f8e84df16f4e6a76dc04799ba747493403236a1b8ac5"
+checksum = "b8343cc1ef1408bd9bdbf69f7aef47017dfab7e6349ec26fddf62e0e9fb5a4cf"
 dependencies = [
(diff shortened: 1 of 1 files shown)


reez (commented):
can we do `cargo update -p miniscript --precise 12.3.7` so it updates the lockfile with only the miniscript changes?

thunderbiscuit:
Yep that's what I did on the first commit. The second commit is just a good'ol cargo update (I figured we need to run one of those once in a while). Would you prefer I leave that commit out of the PR? No hard feelings if you prefer that.

reez:
> Yep that's what I did on the first commit. The second commit is just a good'ol cargo update (I figured we need to run one of those once in a while). Would you prefer I leave that commit out of the PR? No hard feelings if you prefer that.

I would prefer leaving it out of the pr *unless* we did need that broader dependency update anyway? let me know if we do need that though and I can review that part a bit more closely too, I was kinda just reviewing this as a targeted miniscript patch

thunderbiscuit:
@reez sounds good, I removed the commit. We can think about and discuss a more formalized approach for bumping the lockfile on a regular schedule, or with releases, or something else entirely!

reez:
> @reez sounds good, I removed the commit. We can think about and discuss a more formalized approach for bumping the lockfile on a regular schedule, or with releases, or something else entirely!

sounds like a plan 👍

reez (approved):
ACK c45d5b0c626ffd2697f066e8bcb467af35f04bde

LGTM!

PR #1022, opened 2026-06-04
Add test of sync callbacks

thunderbiscuit:
This PR adds 2 tests showcasing the use of the sync and full_scan inspector callbacks.

### Documentation

- [ ] Other: 

### Changelog

No changelog.

#### All Submissions:

### DIFF
--- a/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/SyncCallbacksTest.kt
+++ b/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/SyncCallbacksTest.kt
@@ -1,0 +1,112 @@
+package org.bitcoindevkit
+
+import android.util.Log
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class SyncCallbacksTest {
+    private val conn1: Persister = Persister.newInMemory()
+    private val conn2: Persister = Persister.newInMemory()
+
+    @Test
+    fun syncWithCallbackInspector() {
+        class SyncCallback : SyncScriptInspector {
+            var totalSynced = 0
+            var lastTotal: ULong = 0u
+
+            override fun inspect(script: Script, total: ULong) {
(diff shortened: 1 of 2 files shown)


thunderbiscuit:
Just putting this here to remember: our current implementation of the callback for sync is not perfect, because if you just want to know how many scripts are being synced, you need to do a bit of workaround to do it instead of just pulling the data out of the type. This is not an issue at the ffi layer but the Rust layer. Issue here: bitcoindevkit/bdk#2220.

Currently logging the number of scripts being synced for Tatooine with this (works but is clearly not great).

```kotlin
class SyncCallback : SyncScriptInspector {
    private val logger = KotlinLogging.logger {}
    var totalSynced = 0

    // On the first run of the callback, log the number of scripts that will be inspected.
    override fun inspect(script: Script, total: ULong) {
        if (totalSynced == 0) logger.info { "Syncing $total scripts in this sync" }
        totalSynced++
    }
}
```

Note that this triggers only once the sync is started, and so the number of scripts is unknown until the sync starts.

bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/SyncCallbacksTest.kt
```diff
@@ -0,0 +1,85 @@
...
+            network = Network.REGTEST,
+            persister = conn1
+        )
+        wallet.revealAddressesTo(KeychainKind.EXTERNAL, 16u)
+        val electrumClient = ElectrumClient(ELECTRUM_REGTEST_URL)
+        val syncRequest: SyncRequest = wallet.startSyncWithRevealedSpks()
+            .inspectSpks(SyncCallback())
+            .build()
+
+        electrumClient.sync(
```
reez, line 40: Can we retain the callback instance and assert what it observed after sync/fullScan? Right now these tests pass even if the callback is never called.

thunderbiscuit: Yeah it occured to me after I pushed that this is not actually testing anything yet so I'll add at least an assertion or two on those.

reez (approved):
ACK 8d0cef1fda957196e5b3292b2462b86e70a8bb2c

LGTM!

PR #1024, opened 2026-06-09
Clean up Swift tests

thunderbiscuit:
This PR does 2 things:

1. If you have the `xcbeautify` tool installed, running `just test` will print out more readable output when you run the Swift tests. If you don't have it, it will work as before.
2. Remove the old tests we have not been using, and renaming the ones we do use to their more standard names (removing the `Offline` prefix).

### Changelog

No changelog.

#### All Submissions:


reez (commented):
Could we also update the Swift CI workflow to use this renamed/alltests path? Since the remaining test classes no longer include Offline the existing workflow command swift test --filter Offline passes but runs 0 tests

thunderbiscuit:
Oh shoot yep.

reez (approved):
Still waiting for CI to finish ⏳

but 

ACK cabfda3d60a989ccea92b99a27c7e4d75014ec8f

Issue #1020, opened 2026-06-03
Implement the `Display` trait on a ton of our types


thunderbiscuit:
We don't really have a reason not to export the Display trait already defined on so many of our types. In cases where it's defined on the Rust side, we should consider simply passing it along!

```rust
#[uniffi::export(Display)]
```

chaitika:
hi, can i take this up ?

PR #1031, opened 2026-06-17
feat: expose `Wallet::keychains`

reez:
### Description

Exposes `Wallet::keychains` so users can inspect each wallet keychain and its public descriptor

### Documentation

https://docs.rs/bdk_wallet/3.0.0/src/bdk_wallet/wallet/mod.rs.html#592-594

- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/types.rs
+++ b/bdk-ffi/src/types.rs
@@ -209,6 +209,22 @@ impl From<BdkAddressInfo> for AddressInfo {
     }
 }
 
+/// A wallet keychain and its public descriptor.
+#[uniffi::export(Display)]
+#[derive(uniffi::Record)]
+pub struct WalletKeychain {
+    /// Type of keychain.
+    pub keychain: KeychainKind,
+    /// Public descriptor for the keychain.
+    pub public_descriptor: Arc<Descriptor>,
+}
+
+impl Display for WalletKeychain {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}: {}", self.keychain, self.public_descriptor)
+    }
+}
+
 /// Balance, differentiated into various categories.
 #[derive(uniffi::Record)]
 pub struct Balance {
--- a/bdk-ffi/src/wallet.rs
+++ b/bdk-ffi/src/wallet.rs
@@ -8,10 +8,11 @@ use crate::store::{PersistenceType, Persister};
 use crate::types::{
     AddressInfo, Balance, BlockId, CanonicalTx, ChangeSet, EvictedTx, FullScanRequestBuilder,
     KeychainAndIndex, KeychainKind, LocalOutput, Policy, SentAndReceivedValues, SignOptions,
-    SyncRequestBuilder, UnconfirmedTx, Update, WalletEvent,
+    SyncRequestBuilder, UnconfirmedTx, Update, WalletEvent, WalletKeychain,
 };
 
 use bdk_wallet::bitcoin::Network;
+use bdk_wallet::keys::KeyMap;
 #[allow(deprecated)]
 use bdk_wallet::signer::SignOptions as BdkSignOptions;
 use bdk_wallet::{PersistedWallet, Wallet as BdkWallet};
@@ -422,6 +423,21 @@ impl Wallet {
     /// Get the Bitcoin network the wallet is using.
     pub fn network(&self) -> Network {
         self.get_wallet().network()
+    }
+
+    /// Iterator over all keychains in this wallet
+    pub fn keychains(&self) -> Vec<WalletKeychain> {
+        let wallet = self.get_wallet();
+        wallet
+            .keychains()
+            .map(|(keychain, descriptor)| WalletKeychain {
+                keychain,
+                public_descriptor: Arc::new(Descriptor {
+                    extended_descriptor: descriptor.clone(),
+                    key_map: KeyMap::default(),
+                }),
+            })
+            .collect()
     }
 
     /// Return the balance, separated into available, trusted-pending, untrusted-pending and
--- a/bdk-ffi/src/tests/wallet.rs
+++ b/bdk-ffi/src/tests/wallet.rs
@@ -47,6 +47,39 @@ fn test_create_wallet() {
 }
 
 #[test]
+fn test_keychains() {
+    let wallet = build_wallet();
+
+    let keychains = wallet.keychains();
+
+    assert_eq!(keychains.len(), 2);
+
+    let external = keychains
+        .iter()
+        .find(|keychain| keychain.keychain == KeychainKind::External)
+        .unwrap();
+    let internal = keychains
+        .iter()
+        .find(|keychain| keychain.keychain == KeychainKind::Internal)
+        .unwrap();
+    let external_public_descriptor = external.public_descriptor.to_string();
+    let internal_public_descriptor = internal.public_descriptor.to_string();
+
+    assert_eq!(
+        external_public_descriptor,
+        wallet.public_descriptor(KeychainKind::External)
+    );
+    assert_eq!(
+        internal_public_descriptor,
+        wallet.public_descriptor(KeychainKind::Internal)
+    );
+    assert!(external.public_descriptor.has_wildcard());
+    assert!(internal.public_descriptor.has_wildcard());
+    assert!(!external_public_descriptor.contains("tprv"));
+    assert!(!internal_public_descriptor.contains("tprv"));
+}
+
+#[test]
 fn test_reveal_next_address() {
     let wallet = build_wallet();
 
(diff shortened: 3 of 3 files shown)


bdk-ffi/src/types.rs
```diff
@@ -209,6 +209,15 @@ impl From<BdkAddressInfo> for AddressInfo {
     }
 }
 
+/// A wallet keychain and its public descriptor.
+#[derive(uniffi::Record)]
+pub struct WalletKeychain {
+    /// Type of keychain.
+    pub keychain: KeychainKind,
+    /// Public descriptor for the keychain.
+    pub public_descriptor: String,
```
thunderbiscuit: Suggestion: we could use the full `Descriptor` type here (expanded upon in my main comment).

bdk-ffi/src/wallet.rs
```diff
@@ -424,6 +424,18 @@ impl Wallet {
...
     }
 
+    /// Iterator over all keychains in this wallet
+    pub fn keychains(&self) -> Vec<WalletKeychain> {
+        let wallet = self.get_wallet();
+        wallet
+            .keychains()
+            .map(|(keychain, descriptor)| WalletKeychain {
+                keychain,
+                public_descriptor: descriptor.to_string(),
```
thunderbiscuit: These lines would become

```rust
.map(|(keychain, descriptor)| WalletKeychain {
    keychain,
    public_descriptor: Arc::new(Descriptor {
        extended_descriptor: descriptor.clone(),
        key_map: KeyMap::default(),
    }),
})
```

thunderbiscuit (commented):
This is great! I'm surprised we didn't have this one in before.

My one suggestion would be to consider returning the full `Descriptor` type instead of a string. This would allow users to then call methods like `Descriptor::descriptor_id` and `Descriptor::desc_type` and whatnot.

The trick is to populate the key_map type with an empty keymap, which is what happens on the constructors of `Descriptor` when the provided string is a public descriptor. I convinced myself of this by hacking on the `Descriptor::new` method and adding things like

```rust
eprintln!("The keymap at this point is: {:?}", &key_map);
if key_map.is_empty() {
    eprintln!("####### The key_map is empty! #######")
}
```

Trying this locally and fixing the tests, I realized that our current method `Wallet::public_descriptor` does return a string, but rather than matching this behaviour I would suggest we modify the `Wallet::public_descriptor` to return the full type on our next major release (it'd be a breaking change).

Tests had to be modified:

```rust
    assert_eq!(
        external.public_descriptor.to_string(),
        wallet.public_descriptor(KeychainKind::External).to_string()
    );
    assert_eq!(
        internal.public_descriptor.to_string(),
        wallet.public_descriptor(KeychainKind::Internal).to_string()
    );
    assert!(!external.public_descriptor.to_string().contains("tprv"));
    assert!(!internal.public_descriptor.to_string().contains("tprv"));
```

I would also suggest we derive/write the `Display` trait for this new struct.

```rust
/// A wallet keychain and its public descriptor.
#[derive(uniffi::Record)]
pub struct WalletKeychain {
    /// Type of keychain.
    pub keychain: KeychainKind,
    /// Public descriptor for the keychain.
    pub public_descriptor: Arc<Descriptor>,
}
```

reez:
> This is great! I'm surprised we didn't have this one in before.
> 
> 
> 
> My one suggestion would be to consider returning the full `Descriptor` type instead of a string. This would allow users to then call methods like `Descriptor::descriptor_id` and `Descriptor::desc_type` and whatnot.
> 
> 
> 
> The trick is to populate the key_map type with an empty keymap, which is what happens on the constructors of `Descriptor` when the provided string is a public descriptor. I convinced myself of this by hacking on the `Descriptor::new` method and adding things like
> 
> 
> 
> ```rust
> 
> eprintln!("The keymap at this point is: {:?}", &key_map);
> 
> if key_map.is_empty() {
> 
>     eprintln!("####### The key_map is empty! #######")
> 
> }
> 
> ```
> 
> 
> 
> Trying this locally and fixing the tests, I realized that our current method `Wallet::public_descriptor` does return a string, but rather than matching this behaviour I would suggest we modify the `Wallet::public_descriptor` to return the full type on our next major release (it'd be a breaking change).
> 
> 
> 
> Tests had to be modified:
> 
> 
> 
> ```rust
> 
>     assert_eq!(
> 
>         external.public_descriptor.to_string(),
> 
>         wallet.public_descriptor(KeychainKind::External).to_string()
> 
>     );
> 
>     assert_eq!(
> 
>         internal.public_descriptor.to_string(),
> 
>         wallet.public_descriptor(KeychainKind::Internal).to_string()
> 
>     );
> 
>     assert!(!external.public_descriptor.to_string().contains("tprv"));
> 
>     assert!(!internal.public_descriptor.to_string().contains("tprv"));
> 
> ```
> 
> 
> 
> I would also suggest we derive/write the `Display` trait for this new struct.
> 
> 
> 
> ```rust
> 
> /// A wallet keychain and its public descriptor.
> 
> #[derive(uniffi::Record)]
> 
> pub struct WalletKeychain {
> 
>     /// Type of keychain.
> 
>     pub keychain: KeychainKind,
> 
>     /// Public descriptor for the keychain.
> 
>     pub public_descriptor: Arc<Descriptor>,
> 
> }
> 
> ```

Yes! Great call. Let me update Pr for them

bdk-ffi/src/types.rs
reez: totally, I can't believe I still do this old String thing sometimes, good call out, should be fixed now

bdk-ffi/src/wallet.rs
reez: yes sir, updated 👍

reez:
All great comments @thunderbiscuit , v good review, I think I fixed them all so this should be good to go (other than rebase, I've got a bunch of prs to rebase though and not sure which order I'll merge them in so I'll wait to rebase this atm)

bdk-ffi/src/types.rs
```diff
@@ -209,6 +209,21 @@ impl From<BdkAddressInfo> for AddressInfo {
     }
 }
 
+/// A wallet keychain and its public descriptor.
+#[derive(uniffi::Record)]
+pub struct WalletKeychain {
```
thunderbiscuit, line 215: The Display trait is defined on line 221 but not actually exported to the bindings. We need to add `#[uniffi::export(Display)]`.

In this case for Kotlin the `WalletKeychain` type is a `data class` and so gets a `toString()` implementation for free so we get a nice display, but it's not exactly the string you defined in the `Display` trait implementation on line 221, and I don't think all languages produce this sort of automated toString for simple data types like this one.

```kotlin
// Without exporting Display
[WalletKeychain(keychain=EXTERNAL, publicDescriptor=wpkh([9122d9e0/84'/1'/0']tpubDCYVtmaSaDzTxcgvoP5AHZNbZKZzrvoNH9KARep88vESc6MxRqAp4LmePc2eeGX6XUxBcdhAmkthWTDqygPz2wLAyHWisD299Lkdrj5egY6/0/*)#zpaanzgu), WalletKeychain(keychain=INTERNAL, publicDescriptor=wpkh([9122d9e0/84'/1'/0']tpubDCYVtmaSaDzTxcgvoP5AHZNbZKZzrvoNH9KARep88vESc6MxRqAp4LmePc2eeGX6XUxBcdhAmkthWTDqygPz2wLAyHWisD299Lkdrj5egY6/1/*)#n4cuwhcy)]

// After exporting Display
[External: wpkh([9122d9e0/84'/1'/0']tpubDCYVtmaSaDzTxcgvoP5AHZNbZKZzrvoNH9KARep88vESc6MxRqAp4LmePc2eeGX6XUxBcdhAmkthWTDqygPz2wLAyHWisD299Lkdrj5egY6/0/*)#zpaanzgu, Internal: wpkh([9122d9e0/84'/1'/0']tpubDCYVtmaSaDzTxcgvoP5AHZNbZKZzrvoNH9KARep88vESc6MxRqAp4LmePc2eeGX6XUxBcdhAmkthWTDqygPz2wLAyHWisD299Lkdrj5egY6/1/*)#n4cuwhcy]
```

thunderbiscuit (changes requested):
Just one last little thing.

bdk-ffi/src/types.rs
reez: really nice catch, making a mental note to make sure this doesnt get missed in future prs, fixing

reez: updated code

thunderbiscuit (approved):
ACK 72fe618139aad82a29e452469ffea3b8448b8ffc.

I don't know if there is value in it for Swift users (and maybe Python?) but I know the debug trait populates another field (I forget the name of it) for those (and doesn't do anything for Kotlin).

Anyway something to look at as part of #1020 maybe.

Just needs a rebase.

thunderbiscuit:
Ready for rebase + merge.

PR #1032, opened 2026-06-17
feat: expose `Wallet::get_psbt_input`

reez:
### Description

Exposes Wallet::get_psbt_input so users can derive PSBT input metadata for a wallet owned LocalOutput

### Notes to the reviewers

Tests can be minimized/removed before merge if needed.

### Documentation

https://docs.rs/bdk_wallet/3.0.0/bdk_wallet/struct.Wallet.html#method.get_psbt_input
  
- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/error.rs
+++ b/bdk-ffi/src/error.rs
@@ -1087,20 +1087,11 @@ impl GetPsbtInputError {
             BdkCreateTxError::UnknownUtxo => GetPsbtInputError::UnknownUtxo {
                 outpoint: format!("{}:{}", outpoint.txid, outpoint.vout),
             },
-            other => CreateTxError::from(other).into(),
-        }
-    }
-}
-
-impl From<CreateTxError> for GetPsbtInputError {
-    fn from(error: CreateTxError) -> Self {
-        match error {
-            CreateTxError::UnknownUtxo { outpoint } => GetPsbtInputError::UnknownUtxo { outpoint },
-            CreateTxError::MiniscriptPsbt { error_message } => {
-                GetPsbtInputError::MiniscriptPsbt { error_message }
-            }
+            BdkCreateTxError::MiniscriptPsbt(error) => GetPsbtInputError::MiniscriptPsbt {
+                error_message: error.to_string(),
+            },
             other => GetPsbtInputError::CreateTx {
-                error_message: other.to_string(),
+                error_message: CreateTxError::from(other).to_string(),
             },
         }
     }
--- a/bdk-ffi/src/error.rs
+++ b/bdk-ffi/src/error.rs
@@ -724,6 +724,21 @@ pub enum SighashParseError {
 }
 
 #[derive(Debug, thiserror::Error, uniffi::Error)]
+pub enum GetPsbtInputError {
+    #[error("invalid sighash type: {error_message}")]
+    InvalidSighash { error_message: String },
+
+    #[error("reference to an unknown utxo: {outpoint}")]
+    UnknownUtxo { outpoint: String },
+
+    #[error("miniscript psbt error: {error_message}")]
+    MiniscriptPsbt { error_message: String },
+
+    #[error("create tx error: {error_message}")]
+    CreateTx { error_message: String },
+}
+
+#[derive(Debug, thiserror::Error, uniffi::Error)]
 pub enum PsbtFinalizeError {
     #[error("an input at index {index} is invalid: {reason}")]
     InputError { reason: String, index: u32 },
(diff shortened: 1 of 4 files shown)


reez:
rebased, open for review

bdk-ffi/src/types.rs
```diff
@@ -107,6 +107,24 @@ impl From<BdkChainPosition<BdkConfirmationBlockTime>> for ChainPosition {
...
+        match chain_position {
+            ChainPosition::Confirmed {
+                confirmation_block_time,
+                transitively,
+            } => BdkChainPosition::Confirmed {
+                anchor: confirmation_block_time.into(),
+                transitively: transitively.map(|txid| txid.0),
+            },
+            ChainPosition::Unconfirmed { timestamp } => BdkChainPosition::Unconfirmed {
+                first_seen: timestamp,
```
ItoroD, line 121: I see we leave out first_seen in our ChainPosition type definition but now we have to use same timestamp when passing back. 

Maybe not related to this PR; Asking in general... Do we really not need first seen?

bdk-ffi/src/wallet.rs
```diff
@@ -214,6 +217,26 @@ impl Wallet {
...
     }
 
+    /// Get the corresponding PSBT Input for a `LocalOutput`.
+    #[uniffi::method(default(sighash_type = None, only_witness_utxo = false))]
+    pub fn get_psbt_input(
+        &self,
+        utxo: LocalOutput,
+        sighash_type: Option<String>,
+        only_witness_utxo: bool,
+    ) -> Result<Input, GetPsbtInputError> {
```
ItoroD, line 227: Nice! Took me a while to see why we are returning GetPsbtInputError here when we have CreateTxError. This is good as we need to handle for string parsing, the 2 possible errors from bdk-wallet (miniscript and unknown), and all others.

bdk-ffi/src/error.rs
```diff
@@ -1066,6 +1081,41 @@ impl From<BdkCreateTxError> for CreateTxError {
...
+            },
+            other => CreateTxError::from(other).into(),
+        }
+    }
+}
+
+impl From<CreateTxError> for GetPsbtInputError {
+    fn from(error: CreateTxError) -> Self {
+        match error {
+            CreateTxError::UnknownUtxo { outpoint } => GetPsbtInputError::UnknownUtxo { outpoint },
```
ItoroD: Will there ever be a match on CreateTxError::UnknownUtxo here since we always call from_bdk_create_tx_error which handles the UnknownUtxo?

bdk-ffi/src/wallet.rs
reez: Thanks, yup exactly

bdk-ffi/src/error.rs
reez: Good call, cuz in the current path no, so I simplified this so get_psbt_input maps the upstream errors directly now d9a61d5dc532932c91111c9761636fc8cc51a4a4

bdk-ffi/src/types.rs
reez: Good question, cuz we had already exposed first_seen on TxGraphChangeSet, but ChainPosition was still using the older single timestamp field which maps to upstream last_seen. I opened a followup PR to expose both first_seen and last_seen there too: https://github.com/bitcoindevkit/bdk-ffi/pull/1039

ItoroD:
Thanks for making all the updates! No more comments from me, looks all good.

Do you want to merge #1039 and update first_seen last seen here or vice versa. Either way is fine with me. I could approve this now

reez:
> Thanks for making all the updates! No more comments from me, looks all good.
> 
> 
> 
> Do you want to merge #1039 and update first_seen last seen here or vice versa. Either way is fine with me. I could approve this now

I think we gotta wait on 1039 because it's a breaking change.

And thanks for the good review comments.

bdk-ffi/src/wallet.rs
```diff
@@ -214,6 +217,26 @@ impl Wallet {
...
+        let outpoint = utxo.outpoint.clone();
+        let sighash_type = sighash_type
+            .map(|sighash_type| parse_sighash_type(&sighash_type))
+            .transpose()?;
+        let psbt_input = self
+            .get_wallet()
+            .get_psbt_input(utxo.into(), sighash_type, only_witness_utxo)
```
j-kon, line 234: Could we resolve the `LocalOutput` from `utxo.outpoint` using `wallet.get_utxo()` before calling upstream? Since `LocalOutput` is a constructible UniFFI record, callers can supply an unknown outpoint with a wallet-owned script. I reproduced that this returns `Ok(Input)` with derivation metadata but no witness or non-witness UTXO instead of `UnknownUtxo`. Accepting only an `OutPoint`, or resolving the complete output from wallet state, would keep the PSBT metadata consistent.

PR #1033, opened 2026-06-17
feat: expose TxBuilder coin selection

reez:
### Description

Exposes `TxBuilder::coin_selection` so people can choose the wallet coin selection algo when building transactions.

### Notes to the reviewers

Looking for feedback on if we think this is something we want to do or not

### Documentation

https://docs.rs/bdk_wallet/3.0.0/bdk_wallet/struct.TxBuilder.html#method.coin_selection
  
- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:


reez:
reviewable+rebased

thunderbiscuit (approved):
ACK 9f9233af967a865560bc62d1233ee011e1fc05aa.

It took me a bit of time to dig in and understand the requirement for the new method but I suspected an issue with the generics of the coin selection algo, which was indeed the case.

Thanks for adding this! 

My only little thing I would like to have somewhere in the API docs is a statement of which coin selection is used if none is defined. We could maybe add it to the new enum?

I'll open an issue on the Rust side to bring that up as well.

reez:
rebased

PR #1034, opened 2026-06-18
fix: generate correct BIP39 entropy size in Mnemonic::new()

shivv23:
## Description

Mnemonic::new() was always generating 32 bytes of entropy regardless of the requested word count, with an explicit TODO expressing uncertainty about the approach.

### Changes
- Generate entropy of the correct size based on word count (16 bytes for 12-words, 24 for 18-words, 32 for 24-words)
- Replace the redundant generate_with_entropy → 	o_string() → parse_in() round-trip with a direct rom_entropy() call
- Remove unused imports (GeneratableKey, GeneratedKey, BareCtx)

### Testing
- Added 	est_mnemonic_generation_word_count to verify generated mnemonics have the correct word count

### DIFF
--- a/bdk-ffi/src/keys.rs
+++ b/bdk-ffi/src/keys.rs
@@ -11,10 +11,9 @@ use bdk_wallet::keys::bip39::WordCount;
 use bdk_wallet::keys::bip39::{Language, Mnemonic as BdkMnemonic};
 use bdk_wallet::keys::{
     DerivableKey, DescriptorPublicKey as BdkDescriptorPublicKey,
-    DescriptorSecretKey as BdkDescriptorSecretKey, ExtendedKey, GeneratableKey, GeneratedKey,
+    DescriptorSecretKey as BdkDescriptorSecretKey, ExtendedKey,
 };
 use bdk_wallet::miniscript::descriptor::{DescriptorXKey, Wildcard};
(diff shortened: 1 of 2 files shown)


thunderbiscuit (commented):
Thank you for looking into this. After digging into what the inner call to `Mnemonic::from_entropy` does, you can find that the bdk_wallet library already grabs the required slice of the entropy for the number of words requested. Code here: https://docs.rs/bdk_wallet/3.0.0/src/bdk_wallet/keys/bip39.rs.html#135-150

I think we will simply remove the old outdated TODO, and this PR can be closed.

If you don't mind me asking @shivv23, how did you find out about us and this repo? Have you been working on BDK/bitcoin codebases for a while, or did you arrive at us a different way, or maybe use a crawler for TODOs or first issues on GitHub? Curious how you arrived at us. Thanks!

shivv23:
@thunderbiscuit Hi there! I accept your verdict on the PR. I had been exploring Rust and wanted to work on Bitcoin related projects. BDK came up as an approachable codebase to me to work on wallet payments. I started reading through the bdk-ffi source to understand how UniFFI bindings work and noticed the TODO comment that seemed like a concrete starting point to dive in. Still learning the codebase but found the project interesting. Thanks for the review and explanation about from_entropy. I would love to contribute again something useful. Thanks again!

Issue #1037, opened 2026-06-23
Clarify blocking behavior of ElectrumClient construction


reez:
ElectrumClient::new in the generated bindings performs blocking network I/O because bdk-ffi constructs the upstream electrum_client::Client immediately. 

This issue is to talk about whether bdk-ffi should document that behavior more explicitly, or expose a clearer config/connect API so downstream mobile bindings can avoid UI-thread/main-isolate misuse without diverging per language, etc.

reez:
cc https://github.com/bitcoindevkit/bdk-dart/issues/91

reez:
https://github.com/bitcoindevkit/rust-electrum-client/issues/166 Existing Issue

https://github.com/bitcoindevkit/rust-electrum-client/pull/214 PR opened

PR #1038, opened 2026-06-23
feat(electrum): add config based client constructor

reez:
### Description

Exploration of #1037 

Add an `ElectrumClientConfig` record and explicit connect constructor so bindings can pass around Electrum connection settings before constructing the blocking client

### Documentation

- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/electrum.rs
+++ b/bdk-ffi/src/electrum.rs
@@ -21,6 +21,21 @@ use std::convert::TryFrom;
 use std::sync::Arc;
 use std::time::Duration;
 
+/// Configuration for an electrum client
+#[derive(Clone, uniffi::Record)]
+pub struct ElectrumClientConfig {
+    /// Electrum server URL
+    pub url: String,
+    /// Proxy socks5 configuration, default None
+    pub socks5: Option<String>,
+    /// timeout in seconds, default None (depends on TcpStream default)
+    pub timeout: Option<u8>,
+    /// number of retry if any error, default 1
+    pub retry: Option<u8>,
+    /// when ssl, validate the domain, default true
+    pub validate_domain: bool,
+}
+
 /// Wrapper around an electrum_client::ElectrumApi which includes an internal in-memory transaction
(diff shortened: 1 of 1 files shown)


thunderbiscuit:
I looked into this a bit and agree a fix would be great! I think maybe the issue is at the rust-electrum-client level however, where the constructor makes that network call instead of doing a sort of lazy connection on the first sync, or maybe adding a new method to check the server connection or something. I'll investigate this a bit more on my side too!

bitcoindevkit/rust-electrum-client#212

reez:
> I looked into this a bit and agree a fix would be great! I think maybe the issue is at the rust-electrum-client level however, where the constructor makes that network call instead of doing a sort of lazy connection on the first sync, or maybe adding a new method to check the server connection or something. I'll investigate this a bit more on my side too!
> 
> [bitcoindevkit/rust-electrum-client#212](https://github.com/bitcoindevkit/rust-electrum-client/issues/212)

totally, my first step was figuring out what we could do here on ffi for the upcoming 3.1 (I feel like what I have stood out to me as best/clearest api), and I was still thinking thru what I thought the best idea on upstream electrum-client was so thanks for opening that issue to get it started 👍

j-kon (commented):
Thanks for working on this. I like the direction of introducing `ElectrumClientConfig` because it gives bindings a cleaner way to pass connection settings around before constructing the blocking client.

I also like that the existing constructor now delegates to `ElectrumClient::connect(...)`, which keeps the old API path while avoiding duplicated builder logic.

One question: should we add a small test for the new config-based constructor? Since this PR adds a new public binding-facing constructor, a focused test using the same config values as the existing constructor would help prove both paths stay equivalent.

Also, for downstream bindings, it may be useful to clarify whether `validate_domain` is intentionally required instead of optional with a default. The other config fields are optional, but `validate_domain` must always be passed, so I just want to make sure that is the intended API shape.

PR #1030, opened 2026-06-16
feat: bump bdk_wallet to 3.1.0

reez:
### Description

Dependency bump. The other commits are more experimental to see the shape of exposing new APIs.

### Documentation

- [ ] [`bdk_wallet`](https://docs.rs/bdk_wallet/latest/bdk_wallet/)
  
- [ ] [`bitcoin`](https://docs.rs/bitcoin/latest/bitcoin/index.html)
  
- [ ] [`uniffi`](https://github.com/mozilla/uniffi-rs) 
- [ ] Other: 

#### All Submissions:

* [ ] I've signed all my commits
* [ ] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [ ] I ran `cargo fmt` and `cargo clippy` before committing
* [ ] I've added exactly one `changelog:*` label
* [ ] I've linked the relevant upstream docs or specs above

#### New Features:

* [ ] I've added tests for the new feature
* [ ] I've added docs for the new feature

#### Bugfixes:

* [ ] This pull request breaks the existing API
* [ ] I've added tests to reproduce the issue which are now passing
* [ ] I'm linking the issue being fixed by this PR

PR #1042, opened 2026-07-01
feat: expose two path descriptor wallet loading

reez:
https://github.com/bitcoindevkit/bdk-ffi/pull/1030 needs to go in first

### Description

Adds `Wallet::load_from_two_path_descriptor`, allowing bindings consumers to load a persisted wallet from a BIP 389 two-path descriptor while checking both external and internal keychains.

This mirrors the upstream `bdk_wallet` load flow using `Wallet::load().two_path_descriptor(...).load_wallet(...)`.

### Documentation

https://docs.rs/bdk_wallet/latest/bdk_wallet/struct.LoadParams.html#method.two_path_descriptor
  
- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/wallet.rs
+++ b/bdk-ffi/src/wallet.rs
@@ -168,6 +168,36 @@ impl Wallet {
         })
     }
 
+    /// Build a two-path descriptor `Wallet` by loading from persistence.
+    ///
+    /// Checks that the provided two-path descriptor matches exactly what is loaded
+    /// for both the external and internal keychains.
+    ///
+    /// Note that descriptor secret keys are not persisted to the db. This method
+    /// extracts keys from the provided descriptor while loading.
+    #[uniffi::constructor(default(lookahead = 25))]
+    pub fn load_from_two_path_descriptor(
+        two_path_descriptor: Arc<Descriptor>,
+        persister: Arc<Persister>,
+        lookahead: u32,
+    ) -> Result<Wallet, LoadWithPersistError> {
+        let descriptor = two_path_descriptor.to_string_with_secret();
+        let mut persist_lock = persister.inner.lock().unwrap();
+        let deref = persist_lock.deref_mut();
+
+        let wallet: PersistedWallet<PersistenceType> = BdkWallet::load()
+            .two_path_descriptor(descriptor)
+            .lookahead(lookahead)
+            .extract_keys()
+            .load_wallet(deref)
+            .map_err(LoadWithPersistError::from)?
+            .ok_or(LoadWithPersistError::CouldNotLoad)?;
+
+        Ok(Wallet {
+            inner_mutex: Mutex::new(wallet),
+        })
+    }
+
     /// Build a single-descriptor Wallet by loading from persistence.
     ///
     /// Note that the descriptor secret keys are not persisted to the db.
--- a/bdk-ffi/src/wallet.rs
+++ b/bdk-ffi/src/wallet.rs
@@ -181,7 +181,7 @@ impl Wallet {
         persister: Arc<Persister>,
         lookahead: u32,
     ) -> Result<Wallet, LoadWithPersistError> {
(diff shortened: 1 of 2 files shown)


bdk-ffi/src/wallet.rs
```diff
@@ -168,6 +168,36 @@ impl Wallet {
         })
     }
 
+    /// Build a two-path descriptor `Wallet` by loading from persistence.
```
thunderbiscuit, line 176: Merged yesterday was an update to the API docs for this. See https://github.com/bitcoindevkit/bdk_wallet/commit/77a727d16c13967bc5a78688768f83c556856c10

They now read:

```rust
/// The provided descriptor may only contain extended public keys (`xpub`) with exactly 2 paths,
/// or an error will occur at load time.
```

In our case you can't use that sentence directly because it's just pure impossible to create the `Descriptor` with private keys so you won't "error at load time", but I would use the first part:

```rust
/// The provided descriptor may only contain extended public keys (`xpub`) with exactly 2 paths
```

Note that this goes away with miniscript 13 so bdk_wallet 4.0. Issue here: bitcoindevkit/bdk_wallet#511.

bdk-ffi/src/wallet.rs
```diff
@@ -168,6 +168,36 @@ impl Wallet {
...
+    ///
+    /// Checks that the provided two-path descriptor matches exactly what is loaded
+    /// for both the external and internal keychains.
+    ///
+    /// Note that descriptor secret keys are not persisted to the db. This method
```
thunderbiscuit: This cannot be done so should be removed from API docs.

bdk-ffi/src/wallet.rs
```diff
@@ -168,6 +168,36 @@ impl Wallet {
...
+        lookahead: u32,
+    ) -> Result<Wallet, LoadWithPersistError> {
+        let descriptor = two_path_descriptor.to_string_with_secret();
+        let mut persist_lock = persister.inner.lock().unwrap();
+        let deref = persist_lock.deref_mut();
+
+        let wallet: PersistedWallet<PersistenceType> = BdkWallet::load()
+            .two_path_descriptor(descriptor)
+            .lookahead(lookahead)
+            .extract_keys()
```
thunderbiscuit: And this one too.

bdk-ffi/src/wallet.rs
```diff
@@ -168,6 +168,36 @@ impl Wallet {
...
+    ///
+    /// Note that descriptor secret keys are not persisted to the db. This method
+    /// extracts keys from the provided descriptor while loading.
+    #[uniffi::constructor(default(lookahead = 25))]
+    pub fn load_from_two_path_descriptor(
+        two_path_descriptor: Arc<Descriptor>,
+        persister: Arc<Persister>,
+        lookahead: u32,
+    ) -> Result<Wallet, LoadWithPersistError> {
+        let descriptor = two_path_descriptor.to_string_with_secret();
```
thunderbiscuit: We can remove this line.

thunderbiscuit (changes requested):
Just a few things that need fixing IMO, otherwise the feature is good.

thunderbiscuit:
Oh shoot sorry @reez I reviewed this commit by commit and was only looking at your commit 3 when I made the comments! My bad.

thunderbiscuit (commented):
We can keep the API docs you have here if you like, or you can take the ones that were merged on the Rust side if you prefer. 

The only valid comment from the review above is that line 199 can be removed, as there are no keys to extract.

reez:
undrafted, ready for review

updated pr to address previous comments. 

I will squash everything into one commit when I merge after ack

but I made separate commits for the things I addressed so it would be clearer what was addressed and how

thunderbiscuit (approved):
ACK 3141d04279fad07f1359a8cbf8a694dbb43a07ea. Glad to see this one go in!

PR #1043, opened 2026-07-01
feat: expose wallet signing with provided signers

reez:
https://github.com/bitcoindevkit/bdk-ffi/pull/1030 first

### Description

Adds wrapper for `bdk_wallet::Wallet::sign_with_signers`, allowing bindings consumers to sign a PSBT with explicit `SignersContainer` values instead of only the wallet-owned signers.

This also adds a `SignersContainer` wrapper that can be constructed from descriptor key material and passed into the new wallet signing method.

### Documentation

https://docs.rs/bdk_wallet/3.1.0/bdk_wallet/struct.Wallet.html#method.sign_with_signers
  
- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/signer.rs
+++ b/bdk-ffi/src/signer.rs
@@ -40,7 +40,7 @@ impl SignersContainer {
         Self { inner }
     }
 
-    /// Returns the number of signers in the container.
+    /// Returns the number of signer entries registered in the container.
     pub fn len(&self) -> u64 {
         self.inner.signers().len() as u64
     }
--- a/bdk-ffi/src/signer.rs
+++ b/bdk-ffi/src/signer.rs
@@ -1,0 +1,58 @@
+use crate::descriptor::Descriptor;
+
+use bdk_wallet::bitcoin::key::Secp256k1;
+use bdk_wallet::signer::SignersContainer as BdkSignersContainer;
+
+use std::sync::Arc;
+
+/// Container for multiple signers.
+#[derive(Debug, uniffi::Object)]
+pub struct SignersContainer {
+    pub(crate) inner: BdkSignersContainer,
+}
+
+#[uniffi::export]
+impl SignersContainer {
+    /// Build a new signer container from a descriptor's key map.
+    ///
+    /// Also looks at the same descriptor to determine the `SignerContext` to attach to the signers.
+    #[uniffi::constructor]
+    pub fn from_descriptor(descriptor: Arc<Descriptor>) -> Self {
+        Self::from_descriptor_with_context(Arc::clone(&descriptor), descriptor)
+    }
+
+    /// Build a new signer container from a signer descriptor's key map.
+    ///
+    /// Also looks at the corresponding descriptor to determine the `SignerContext` to attach to
+    /// the signers.
+    #[uniffi::constructor]
+    pub fn from_descriptor_with_context(
+        signer_descriptor: Arc<Descriptor>,
+        context_descriptor: Arc<Descriptor>,
+    ) -> Self {
+        let secp = Secp256k1::new();
+        let inner = BdkSignersContainer::build(
+            signer_descriptor.key_map.clone(),
+            &context_descriptor.extended_descriptor,
+            &secp,
+        );
+
+        Self { inner }
+    }
+
+    /// Returns the number of signers in the container.
+    pub fn len(&self) -> u64 {
+        self.inner.signers().len() as u64
+    }
+
+    /// Returns true when the container has no signers.
+    pub fn is_empty(&self) -> bool {
+        self.inner.signers().is_empty()
+    }
+}
+
+impl From<BdkSignersContainer> for SignersContainer {
+    fn from(inner: BdkSignersContainer) -> Self {
+        Self { inner }
+    }
+}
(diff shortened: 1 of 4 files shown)


bdk-ffi/src/signer.rs
```diff
@@ -0,0 +1,52 @@
...
+        Self { inner }
+    }
+
+    /// Returns the number of signer entries registered in the container.
+    pub fn len(&self) -> u64 {
+        self.inner.signers().len() as u64
+    }
+
+    /// Returns true when the container has no signers.
+    pub fn is_empty(&self) -> bool {
```
thunderbiscuit, line 49: Is it possible for us to hit a false here? The way our constructor works, it appears you'll always have one, unless maybe the public descriptors sort of silently fail and don't populate/produce a valid signer?

The Rust type has a `SignersContainer::new` constructor which creates one without any signers in it, but here we're not offering that (and I'm not sure why we would either even if we could). https://docs.rs/bdk_wallet/latest/bdk_wallet/signer/struct.SignersContainer.html#method.new

thunderbiscuit (commented):
I have one small question regarding the usefulness of the `is_empty` method, but otherwise this is ready to go.

Because it's one of the core new features/workflow for the library and hints at where we're going moving forward (creating wallets using only public keys and whatnot (see the upcoming [Design page on this topic](https://thunderbiscuit.github.io/book-of-bdk/design/signing/)), I'd like to have it expanded upon in tests if possible.

I'd be happy to add tests if you'd like me to push on this branch @reez. That will get me playing with the API a bit more too. But they could also come in follow-up PRs, I'm not picky.

reez:
> I'd be happy to add tests if you'd like me to push on this branch @reez. That will get me playing with the API a bit more too. But they could also come in follow-up PRs, I'm not picky.

Still have this in draft because I need to update it to make sure it's easy to review after breaking it out from the larger pr, so just keep that in mind but if you do want to check it out and have feedback feel free to post it and I can make the updates myself as I get this pr in cleanest shape

bdk-ffi/src/signer.rs
reez: > Is it possible for us to hit a false here?

Im looking at it like secret descriptors make `is_empty()` return `false`, public descriptors make it return `true`

> The way our constructor works, it appears you'll always have one,

Only when the descriptor contains secret key material, public only descriptors have an empty `key_map` so that's how I see the constructor can produce a container with no signers

> unless maybe the public descriptors sort of silently fail and don't populate/produce a valid signer?

I don't think they silently fail, a public descriptor is valid but it doesnt have secret key material in its `key_map` so there's just nothing for `SignersContainer::build(...)` to turn into a signer

> The Rust type has a SignersContainer::new constructor which creates one without any signers in it, but here we're not offering that (and I'm not sure why we would either even if we could).

Agreed

reez:
addressed questions (I think all of them?) from https://github.com/bitcoindevkit/bdk-ffi/pull/1030 in a few of the follow up commits here

I'll squash everything into one commit when I merge it though for sure

Just FYI

bdk-ffi/src/signer.rs
reez: Added explicit coverage in `test_signers_container_from_descriptor` now

reez:
I think this is ready for final review @thunderbiscuit 

I can squash commits when merge, but just leaving commits separate for review so its easier to see what changes where made and when, especially since I split this out of a mega pr that might have had some related comments on it

thunderbiscuit (approved):
ACK 1c8ac280968fc01d14ce33795c1e0abaac323004. I also agree that the commit history could be cleaned up a bit before merging.

While trying this out locally I found the flow between public and private descriptors is not as smooth as it could be, especially since this is expected to be the new workflow for users. Currently instead of having access to a direct conversion from a private descriptor to it's public equivalent, you roughly do this:

```koltin
val privateDescriptor = Descriptor.newBip86(secretKey, KeychainKind.EXTERNAL, NetworkKind.TEST)
val publicDescriptor = Descriptor(privateDescriptor.toString(), NetworkKind.TEST)
```

There are other ways to do it to, but they're even worse (create the private descriptor, get the xpub, find the fingerprint, and feed them to the `Descriptor.newBip84Public` API for example).

Food for thoughts for further PRs!

reez:
> ACK [1c8ac28](https://github.com/bitcoindevkit/bdk-ffi/commit/1c8ac280968fc01d14ce33795c1e0abaac323004). I also agree that the commit history could be cleaned up a bit before merging.
> 
> While trying this out locally I found the flow between public and private descriptors is not as smooth as it could be, especially since this is expected to be the new workflow for users. Currently instead of having access to a direct conversion from a private descriptor to it's public equivalent, you roughly do this:
> 
> ```
> val privateDescriptor = Descriptor.newBip86(secretKey, KeychainKind.EXTERNAL, NetworkKind.TEST)
> val publicDescriptor = Descriptor(privateDescriptor.toString(), NetworkKind.TEST)
> ```
> 
> There are other ways to do it to, but they're even worse (create the private descriptor, get the xpub, find the fingerprint, and feed them to the `Descriptor.newBip84Public` API for example).
> 
> Food for thoughts for further PRs!

adding a follow up here https://github.com/bitcoindevkit/bdk-ffi/pull/1054

PR #1044, opened 2026-07-01
feat: expose Wallet create and load params

reez:
https://github.com/bitcoindevkit/bdk-ffi/pull/1030 first

### Description

Exposes CreateParams and LoadParams so callers can configure genesis hash, lookahead, SPK cache, and loadtime checks when creating or loading wallets.

### Documentation

- [x] [`CreateParams`](https://docs.rs/bdk_wallet/3.1.0/bdk_wallet/struct.CreateParams.html)
- [x] [`LoadParams`](https://docs.rs/bdk_wallet/3.1.0/bdk_wallet/struct.LoadParams.html)
- [x] [`LoadParams::two_path_descriptor`](https://docs.rs/bdk_wallet/3.1.0/bdk_wallet/struct.LoadParams.html#method.two_path_descriptor)

- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/tests/wallet.rs
+++ b/bdk-ffi/src/tests/wallet.rs
@@ -413,3 +413,37 @@ fn test_load_from_two_path_descriptor() {
         1
     );
 }
+
+#[test]
+fn test_load_from_two_path_descriptor_with_params() {
+    let persister = Arc::new(Persister::new_in_memory().unwrap());
+    Wallet::create_from_two_path_descriptor(
+        two_path_descriptor(),
+        Network::Signet,
+        Arc::clone(&persister),
+        25,
+    )
+    .unwrap();
+
+    let params = LoadParams {
+        check_network: Some(Network::Bitcoin),
+        check_genesis_hash: None,
+        lookahead: 25,
+        use_spk_cache: false,
+    };
+    let error = match Wallet::load_from_two_path_descriptor_with_params(
+        two_path_descriptor(),
+        persister,
+        params,
+    ) {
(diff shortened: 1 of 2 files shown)


reez:
I kept it in 2 commits because 2nd commit built on #1042, but will squash into 1 commit when merge

thunderbiscuit (approved):
ACK af845a5a2da9f80ca9885eef646fd5f3269e942b.

We're now at 12 constructors on the wallet, which is a bit of a code smell for me (and the lookahead is now potentially passed twice, once as a straight argument and then again in the params argument).

Looking at this I wonder if adding the Params as a default of None would solve this (12 constructors to 6). The issues is maybe how downstream languages handle default arguments (I haven't dug into whether that'd be breaking for JS for example).

```rust
#[derive(Clone, Debug, uniffi::Record)]
pub struct CreateParams {
    #[uniffi(default = None)]
    pub genesis_hash: Option<Arc<BlockHash>>,
    #[uniffi(default = false)]
    pub use_spk_cache: bool,
}

#[uniffi::constructor(default(lookahead = 25, params = None))]
pub fn new(
    descriptor: Arc<Descriptor>,
    change_descriptor: Arc<Descriptor>,
    network: Network,
    persister: Arc<Persister>,
    lookahead: u32,
    params: Option<CreateParams>,
) -> Result<Self, CreateWithPersistError>
```

Overall this is a good addition and non-breaking, so I think we can just go ahead and ship 3.1 with it. But looking at 4.0 cleaning this up would be a good goal.

PR #1045, opened 2026-07-03
feat: expose Display trait on Amount and Psbt types

shivv23:
### Description

Add #[uniffi::export(Display)] and implement the Display trait on the Amount and Psbt types, exposing them to the FFI bindings (Kotlin, Swift, Python).

### Notes to the reviewers

This follows the same pattern used by other types in the codebase (FeeRate, Script, Address, Transaction, WalletEvent, etc.) that already expose Display to bindings. Part of issue #1020.

### Documentation

- [x] [bitcoin](https://docs.rs/bitcoin/latest/bitcoin/index.html) — [Amount](https://docs.rs/bitcoin/latest/bitcoin/amount/struct.Amount.html#impl-Display-for-Amount) and [Psbt](https://docs.rs/bitcoin/latest/bitcoin/psbt/struct.Psbt.html#impl-Display-for-Psbt) already implement Display upstream.

### Changelog

changelog: none

#### All Submissions:

- [x] I've signed all my commits
- [x] I followed the contribution guidelines
- [x] I ran cargo fmt and cargo clippy before committing
- [x] I've added exactly one changelog:* label
- [x] I've linked the relevant upstream docs or specs above

#### New Features:

- [ ] I've added tests for the new feature
- [ ] I've added docs for the new feature

### DIFF
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -217,6 +217,7 @@ impl_into_core_type!(FeeRate, BdkFeeRate);
 /// underflow occurs. Also note that since the internal representation of amounts is unsigned,
 /// subtracting below zero is considered an underflow and will cause a panic.
 #[derive(Debug, Clone, PartialEq, Eq, uniffi::Object)]
+#[uniffi::export(Display)]
 pub struct Amount(pub(crate) BdkAmount);
 
 #[uniffi::export]
@@ -248,6 +249,12 @@ impl Amount {
 
 impl_from_core_type!(BdkAmount, Amount);
 impl_into_core_type!(Amount, BdkAmount);
+
+impl Display for Amount {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
 
 /// A bitcoin script: https://en.bitcoin.it/wiki/Script
 #[derive(Clone, Debug, uniffi::Object)]
@@ -1389,6 +1396,7 @@ impl From<&BdkOutput> for Output {
 
 /// A Partially Signed Transaction.
 #[derive(uniffi::Object)]
+#[uniffi::export(Display)]
 pub struct Psbt(pub(crate) Mutex<BdkPsbt>);
 
 #[uniffi::export]
@@ -1529,6 +1537,12 @@ impl Psbt {
 impl From<BdkPsbt> for Psbt {
     fn from(psbt: BdkPsbt) -> Self {
         Psbt(Mutex::new(psbt))
+    }
+}
+
+impl Display for Psbt {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{}", self.0.lock().unwrap())
     }
 }
 


j-kon:
Nice small addition, and this seems consistent with the Display exports already used for types like FeeRate, Script, Address, and Transaction.

One small question: should this PR add a focused test for the new binding-facing Display behavior? Since this exposes Display through UniFFI, it would be useful to assert that Amount and Psbt stringify as expected, especially because the PR checklist marks tests as added but the diff only changes bitcoin.rs.

shivv23:
Hey @j-kon , I should have included tests with the initial commit. I've pushed test_amount_display and test_psbt_display to cover the new Display behavior for both types.

Regarding the changelog label — I have changelog: added in the PR body but it seems the label itself needs to be applied by someone with triage access. Could you add it?

Thanks for the review!

j-kon:
> Hey @j-kon , I should have included tests with the initial commit. I've pushed test_amount_display and test_psbt_display to cover the new Display behavior for both types.
> 
> Regarding the changelog label — I have changelog: added in the PR body but it seems the label itself needs to be applied by someone with triage access. Could you add it?
> 
> Thanks for the review!

Thanks for adding the tests, that addresses my review comment.

I don’t think I have triage access to apply labels on this repo, so I can’t add the `changelog: added` label myself. Maybe a maintainer can help apply it.

bdk-ffi/src/bitcoin.rs
```diff
@@ -217,8 +217,15 @@ impl_into_core_type!(FeeRate, BdkFeeRate);
 /// underflow occurs. Also note that since the internal representation of amounts is unsigned,
 /// subtracting below zero is considered an underflow and will cause a panic.
 #[derive(Debug, Clone, PartialEq, Eq, uniffi::Object)]
+#[uniffi::export(Display)]
 pub struct Amount(pub(crate) BdkAmount);
 
+impl Display for Amount {
```
ItoroD: Just a nit. Usually we place the display Impl at the bottom of the type implementation. i.e after all implementations of the particular type.

bdk-ffi/src/bitcoin.rs
```diff
@@ -1389,8 +1396,15 @@ impl From<&BdkOutput> for Output {
 
 /// A Partially Signed Transaction.
 #[derive(uniffi::Object)]
+#[uniffi::export(Display)]
 pub struct Psbt(pub(crate) Mutex<BdkPsbt>);
 
+impl Display for Psbt {
```
ItoroD: same here

ItoroD (changes requested):
Thanks for adding this. LGTM. Just a few comments commit message for the test should start with 

`test: add ....`

bdk-ffi/src/bitcoin.rs
shivv23: Done!

bdk-ffi/src/bitcoin.rs
shivv23: Done!

shivv23:
Hi @ItoroD! I have made all changes you mentioned to the PR. Kindly review it again.

bdk-ffi/src/tests/bitcoin.rs
```diff
@@ -583,6 +583,22 @@ fn test_psbt_output_tap_tree() {
     );
 }
 
+#[test]
+fn test_amount_display() {
+    let amount = crate::bitcoin::Amount::from_sat(100_000_000);
```
ItoroD: Here, maybe we can use the import at the top of the file like the others. So no need to use the fully qualified import here.

ItoroD (changes requested):
On second thought I am thinking we might not need the tests. As in the past we do not typically write tests explicitly for display. This will mean we will have lots of solely display tests for various types which I do not think its necessary. I will just leave this comment here to see what other reviews think. 

I am happy to approve one the import change is made.

bdk-ffi/src/tests/bitcoin.rs
```diff
@@ -583,6 +583,22 @@ fn test_psbt_output_tap_tree() {
     );
 }
 
+#[test]
```
thunderbiscuit: I agree with @ItoroD; we do not need to add tests for the `Display` trait.

thunderbiscuit (changes requested):
If you can:
1. Remove the test
2. Rebase on master

I think this one is ready to go.

bdk-ffi/src/tests/bitcoin.rs
shivv23: @ItoroD Done!

bdk-ffi/src/tests/bitcoin.rs
shivv23: @thunderbiscuit Agree with you, have made the required changes.

shivv23:
@ItoroD and @thunderbiscuit Hi! I have made the required changes requested. Thanks!

thunderbiscuit:
I apologize @shivv23 but I didn't notice your commits were not PGP signed. This is a requirement of the BDK Org, so you'll have to sign that commit before I can merge. Otherwise this looks good! (also keep an eye out because if I merge anything in between you'll also have to rebase).

shivv23:
Hi @thunderbiscuit! All feedback addressed (tests removed, Display placement moved to bottom of type impls, rebased on master). Commit is now GPG-signed and verified. Please take another look.

ItoroD (approved):
ACK a62a58bd8e6ea46c2bb4d2719e653587a0bc0d8d

thunderbiscuit (approved):
ACK a62a58bd8e6ea46c2bb4d2719e653587a0bc0d8d.

Issue #1048, opened 2026-07-13
Expose SecpCtx in Wallet


ItoroD:
Psbt::sign requires Secp256k1 as parameter.

```rust
pub fn sign<C, K>(
        &mut self,
        k: &K,
        secp: &Secp256k1<C>,
    )
```

ItoroD:
currently working on this

PR #1049, opened 2026-07-14
feat: Add type secpCtx for Wallet

ItoroD:
### Description

Addresses #1048 

### Notes to the reviewers
This exposes Secp256k1 for the purposes of using this in our psbt sign. Especially if we want the API to be as close to bdk-wallet as possible.

### Documentation

- [ ] Other: 

#### All Submissions:

#### New Features:

### DIFF
--- a/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/WalletTest.kt
+++ b/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/WalletTest.kt
@@ -6,6 +6,7 @@ import kotlin.test.assertTrue
 import kotlin.test.assertFalse
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import org.junit.runner.RunWith
+import kotlin.test.assertNotNull
 
 @RunWith(AndroidJUnit4::class)
 class WalletTest {
@@ -64,5 +65,18 @@ class WalletTest {
             actual = address2.address,
             "Addresses should be the same"
         )
+    }
+
+    @Test
+    fun getWalletSecpCtx() {
+        val wallet: Wallet = Wallet.createSingle(
+            descriptor = BIP84_DESCRIPTOR,
+            network = Network.TESTNET,
+            persister = conn
+        )
+        val secp = wallet.secpCtx()
+        val stringSpec = secp.toString()
+
+        assertNotNull(stringSpec, "SecpCtx should not be null")
     }
 }
--- a/bdk-ffi/src/types.rs
+++ b/bdk-ffi/src/types.rs
@@ -43,6 +43,7 @@ use bdk_esplora::esplora_client::api::MerkleProof as BdkMerkleProof;
 use bdk_esplora::esplora_client::api::OutputStatus as BdkOutputStatus;
 use bdk_esplora::esplora_client::api::Tx as BdkTx;
 use bdk_esplora::esplora_client::api::TxStatus as BdkTxStatus;
+use bdk_wallet::bitcoin::secp256k1::{All, Secp256k1};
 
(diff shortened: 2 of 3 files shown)


thunderbiscuit (commented):
Secp context is only really useful if you need to keep one around and not have to re-create it every time you need it, but in this case we can't actually pass ownership of it to the bindings side and instead do a full copy of it, defeating the purpose.

My take on this is that we should create a new secp context inside the Psbt::sign wrapper instead, and not require it as an argument. Note that this is also what we do with all our other APIs that require a secp context; we just create one on the fly.

Asking Claude to dig into this a bit here is what I get:

---

The secp parameter in Psbt::sign is a Rust-ergonomics affordance (letting callers amortize context creation across many calls in hot loops), not a semantic input: any Signing-capable context produces identical signatures, and there is nothing about the wallet's context that Psbt::sign needs — it doesn't even have to come from a wallet.

The rationale was that it was easier to keep one context around because they were costly to create. This is largely historical now.

Why it used to matter: in older versions of libsecp256k1, creating a context computed large precomputed multiplication tables (on the order of a megabyte of table generation work) used to accelerate the elliptic-curve operations. Doing that per signature would have dwarfed the cost of the signature itself, so the library was designed around "create one context at startup, pass it everywhere" — and every API downstream (rust-secp256k1, rust-bitcoin, BDK) inherited that shape.

What changed: around libsecp256k1 v0.5.0 (rust-secp256k1 ~0.22, 2022), the precomputed tables were moved into the library binary as static data. A new context is now a small heap allocation plus initialization — cheap enough that creating one per operation is fine for anything that isn't a tight loop signing thousands of times.

ItoroD:
> My take on this is that we should create a new secp context inside the Psbt::sign wrapper instead, and not require it as an argument. Note that this is also what we do with all our other APIs that require a secp context; we just create one on the fly.


Yes this is true and something I also considered. Secp does not need to be passed as a param and can be handled and used under the hood. The only reason I did this is trying to make the sign method as close to the rust API as possible. Because I assumed that there can be a custom secp256k1 instances user might want to pass. This is possible in rust. I can also see there are different context that Secp256k1 can have. i.e just signing or just verifying or all.

But I think in our context we sticking to a default which is `All`. Which is what wallets secpCtx always returns anyway. This is why not making it a param can work.

> Note that this is also what we do with all our other APIs that require a secp context; we just create one on the fly.

True. Only thing is that the way psbt sign has been presented seemed like, now a user has full control of how they want to sign even down to the sepCtx.

> Signing-capable context produces identical signatures, and there is nothing about the wallet's context that Psbt::sign needs — it doesn't even have to come from a wallet.

I agree as from my tests it spits out the same sepcCtx value every time. Makes me wonder why we have this method in Wallet. I should [ask](https://github.com/bitcoindevkit/bdk_wallet/pull/505/changes#r3588798320)

Since we will not pass this as param for psbt.sign, I assume we do not want to expose this on wallet too ?

j-kon:
I think that makes sense from the downstream binding API side too.

If `Psbt::sign` can create the secp context internally, then exposing `Wallet.secpCtx()` may add a public binding API that users do not really need. For Dart/Kotlin/Swift users, keeping `sign` simple and hiding the secp context internally feels cleaner than asking callers to manage a context that is not really a semantic input to signing.

So I’d lean toward not exposing this on `Wallet` unless there is another binding-facing use case for `SecpCtx` outside `Psbt::sign`.

ItoroD:
> So I’d lean toward not exposing this on `Wallet` unless there is another binding-facing use case for `SecpCtx` outside `Psbt::sign`.

I agree. I will close this once everything with psbt.sign PR has been approved.

PR #1050, opened 2026-07-14
test: optimize the podman-regtest-infinity-pro service in CI

thunderbiscuit:
This PR bumps the podman-regtest-infinity-pro image to the latest release (0.4.0), and optimizes some of the setup.

- No need to forward port 3003, that's the bitcoin explorer (only really usable in a browser)
- The container now fires much faster, and so doing the check on a quicker time loop will save time on each run
- We should really check for 100 blocks mined (which is what the container does) before continuing, because the tests use the faucet made available in the regtest environment, but that only becomes usable after block 100.

Sister PR on bdk-jvm is here: bitcoindevkit/bdk-jvm#68

### Changelog

No changelog

#### All Submissions:

### DIFF
--- a/.github/workflows/test-android.yaml
+++ b/.github/workflows/test-android.yaml
@@ -23,19 +23,18 @@ jobs:
 
     services:
       bitcoin-regtest:
-        image: ghcr.io/thunderbiscuit/podman-regtest-infinity-pro:0.3.2
+        image: ghcr.io/thunderbiscuit/podman-regtest-infinity-pro:0.4.0
         ports:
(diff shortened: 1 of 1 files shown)


ItoroD (approved):
ACK 02827a7fe472aa48c8bf52cd1aa54562c3eb659e

thunderbiscuit:
Running the `Test Android` workflow against this update here: https://github.com/bitcoindevkit/bdk-ffi/actions/runs/29423446937/job/87379600740.

PR #1056, opened 2026-08-04
feat: expose extract_tx_with_fee_rate_limit and extract_tx_unchecked_fee_rate on Psbt

### LINKED ISSUE #1053: Expose extract_tx_with_fee_rate_limit on Psbt — default cap rejects all normal Dogecoin transactions
### Problem

`Psbt.extractTx()` is unusable for Dogecoin: rust-bitcoin enforces its hardcoded `DEFAULT_MAX_FEE_RATE` of 25,000 sat/vB inside `extract_tx()`, and every regular DOGE transaction exceeds it. Normal Dogecoin fees are ~50,000–120,000 sat/vB (≈1 DOGE/kvB — worth cents in fiat); Dogecoin Core's own [`RECOMMENDED_MIN_TX_FEE` (https://github.com/dogecoin/dogecoin/blob/master/src/policy/policy.h#L23) is already `COIN / 100` per kvB = 1,000 sat/vB, i.e. the cap sits *below* the network's recommended minimum territory. So any PSBT-based flow on DOGE fails with `ExtractTxError::AbsurdFeeRate` on perfectly valid transactions (reproduced on mainnet with fee rates around 95k sat/vB).

### What rust-bitcoin already has

rust-bitcoin exposes public methods for exactly this case — [`extract_tx_with_fee_rate_limit(FeeRate)` and `extract_tx_unchecked_fee_rate()`](https://github.com/rust-bitcoin/rust-bitcoin/blob/bitcoin-0.32.0/bitcoin/src/psbt/mod.rs#L160) — but bdk-ffi wraps only plain `extract_tx()` ([`bdk-ffi/src/bitcoin.rs`](https://github.com/bitcoindevkit/bdk-ffi/blob/master/bdk-ffi/src/bitcoin.rs#L1444)), so the cap is not configurable from Swift/Kotlin/Python.

### Proposal

Add to the `Psbt` interface:

```rust
pub fn extract_tx_with_fee_rate_limit(
    &self,
    max_fee_rate: Arc<FeeRate>,
) -> Result<Arc<Transaction>, ExtractTxError>
```

`FeeRate` is already exposed, so it's a small wrapper following the existing `extract_tx()` pattern. Happy to open the PR with a test and CHANGELOG entry if you're open to it.

ArcaniZZZ:
### Description

Exposes two `Psbt` transaction-extraction variants that already exist in rust-bitcoin:

- `extract_tx_with_fee_rate_limit(max_fee_rate: FeeRate)` — same as `extract_tx`, but with a caller-provided maximum fee rate instead of the hardcoded 25,000 sat/vB cap
- `extract_tx_unchecked_fee_rate()` — no fee rate check at all

The default cap makes `extract_tx` unusable on networks with low coin value: e.g. on Dogecoin a typical fee rate is 50,000–120,000 sat/vB, so every regular transaction fails with `ExtractTxError::AbsurdFeeRate` despite being perfectly valid on the network.

Fixes #1053

### Notes to the reviewers

- Both wrappers follow the existing `extract_tx` pattern in `bdk-ffi/src/bitcoin.rs`; `FeeRate` is already exposed, and the `ExtractTxError` mapping already covers all variants, so the change is pure delegation.
- I didn't add tests since the wrappers delegate directly to rust-bitcoin, where this logic is already covered — happy to add binding-level tests if you'd prefer.

### Documentation

  - [`Psbt::extract_tx_with_fee_rate_limit`](https://docs.rs/bitcoin/0.32.8/bitcoin/struct.Psbt.html#method.extract_tx_with_fee_rate_limit)
  - [`Psbt::extract_tx_unchecked_fee_rate`](https://docs.rs/bitcoin/0.32.8/bitcoin/struct.Psbt.html#method.extract_tx_unchecked_fee_rate)

#### All Submissions:

* [ ] I ran `cargo fmt` and `cargo clippy` before committing (no local Rust toolchain — relying on CI)
* [ ] I've added exactly one `changelog:*` label (`changelog: added` — I don't have permission to add labels)

#### New Features:

### DIFF
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -1447,6 +1447,38 @@ impl Psbt {
         Ok(Arc::new(transaction))
     }
 
+    /// Extracts the `Transaction` from a `Psbt` by filling in the available signature information.
+    ///
+    /// #### Errors
+    ///
+    /// See `extract_tx`.
+    pub fn extract_tx_with_fee_rate_limit(
+        &self,
+        max_fee_rate: Arc<FeeRate>,
+    ) -> Result<Arc<Transaction>, ExtractTxError> {
+        let tx: BdkTransaction = self
+            .0
+            .lock()
+            .unwrap()
+            .clone()
+            .extract_tx_with_fee_rate_limit(max_fee_rate.0)?;
+        let transaction: Transaction = tx.into();
+        Ok(Arc::new(transaction))
+    }
+
+    /// Perform `extract_tx` without the fee rate check.
+    ///
+    /// This can result in a transaction with absurdly high fees. Use with caution.
+    pub fn extract_tx_unchecked_fee_rate(&self) -> Arc<Transaction> {
+        let tx: BdkTransaction = self
+            .0
+            .lock()
+            .unwrap()
+            .clone()
+            .extract_tx_unchecked_fee_rate();
+        Arc::new(tx.into())
+    }
+
     /// Calculates transaction fee.
     ///
     /// 'Fee' being the amount that will be paid for mining a transaction with the current inputs


bdk-ffi/src/bitcoin.rs
```diff
@@ -1447,6 +1447,50 @@ impl Psbt {
...
+    ///
+    /// This is useful for networks where fee rates routinely exceed the default limit (for example
+    /// forks with a much lower coin value), on which `extract_tx` would reject perfectly valid
+    /// transactions with `ExtractTxError::AbsurdFeeRate`.
+    ///
+    /// #### Errors
+    ///
+    /// `ExtractTxError` variants will contain either the `Psbt` itself or the `Transaction`
+    /// that was extracted. These can be extracted from the Errors in order to recover.
+    /// See the error documentation for info on the variants. In general, it covers large fees.
```
reez: can we use upstreams code comment https://docs.rs/bitcoin/0.32.8/src/bitcoin/psbt/mod.rs.html#155 as the code comment for this instead?

bdk-ffi/src/bitcoin.rs
```diff
@@ -1447,6 +1447,50 @@ impl Psbt {
...
+        Ok(Arc::new(transaction))
+    }
+
+    /// Extracts the `Transaction` from a `Psbt` by filling in the available signature information,
+    /// without checking the fee rate.
+    ///
+    /// ## Warning
+    ///
+    /// The transaction is not checked against the fee rate, so this can produce a transaction with
+    /// absurdly high fees. Use `extract_tx` or `extract_tx_with_fee_rate_limit` where possible.
```
reez: can we use these lines from upstream code comment https://docs.rs/bitcoin/0.32.8/src/bitcoin/psbt/mod.rs.html#169-171 as the code comments for this instead?

reez (commented):
Looking good overall. I left two small documentation nits about mirroring the upstream comments. @thunderbiscuit  could you take a look and let me know whether you agree? I’m fine with keeping the current wording if you think it’s better.

bdk-ffi/src/bitcoin.rs
ArcaniZZZ: Done in 431881a 👍

bdk-ffi/src/bitcoin.rs
ArcaniZZZ: Done in 431881a. One small adaptation: upstream says `extract_tx_fee_rate_limit`, but the bindings only expose `extract_tx` (which upstream aliases to it), so I referenced `extract_tx` to keep the doc pointing at a method that exists here. Happy to switch to the literal upstream wording if you prefer.

thunderbiscuit (changes requested):
This looks good and is a valuable API to add. Please remove the changelog entry and rebase on master and it's ready to merge!

thunderbiscuit:
One thing this PR made me realize is that our docs on `extract_tx` indeed match the Rust side but are misleading because our error unfortunately does not expose neither the problematic tx nor PSBT.

reez:
> One thing this PR made me realize is that our docs on `extract_tx` indeed match the Rust side but are misleading because our error unfortunately does not expose neither the problematic tx nor PSBT.

Yup thanks for opening that Issue

ArcaniZZZ:
Done: removed the changelog entry and rebased on master (squashed into a single commit while at it).

thunderbiscuit (approved):
ACK a5f1a45ce1000dbf1ece3c01e8d267eb97085cc8.

PR #1067, opened 2026-08-10
fix(ci): fail unsuccessful external binding dispatches

reez:
### Description

Updates external binding dispatches to fail on unsuccessful HTTP responses which prevents downstream CI failures from silently passing. But someone let me know if that's intended for a reason I'm not remembering?

### Documentation

- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/.github/workflows/trigger-external-bindings.yml
+++ b/.github/workflows/trigger-external-bindings.yml
@@ -42,6 +42,7 @@ jobs:
 
       # Trigger trigger-bdk-python-test workflow in bdk-python repository
       - name: Trigger tests in bdk-python repository
+        if: ${{ !cancelled() }}
         env:
           BDK_PYTHON_ACCESS_TOKEN: ${{ secrets.BDK_PYTHON_ACCESS_TOKEN }}
         run: |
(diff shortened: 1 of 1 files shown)


thunderbiscuit (commented):
Good find. Worth merging. But I have 2 things I'd add while we're on this workflow:

1. The Python job will never run if the jvm run fails. We can fix that.
2. (from Claude): The body dump never prints on failure. Actions runs run: steps under bash -e, so a nonzero curl aborts the step immediately — the echo "Response body:" / jq block below it is skipped exactly when you'd want to read the error. You get curl's own curl: (22) The requested URL returned error: 401 from --show-error, but not GitHub's "message": "Bad credentials". Piping the body inline instead of to a file, or --output /dev/stderr, would keep it.

reez:
> Good find. Worth merging. But I have 2 things I'd add while we're on this workflow:
> 
> 1. The Python job will never run if the jvm run fails. We can fix that.
> 2. (from Claude): The body dump never prints on failure. Actions runs run: steps under bash -e, so a nonzero curl aborts the step immediately — the echo "Response body:" / jq block below it is skipped exactly when you'd want to read the error. You get curl's own curl: (22) The requested URL returned error: 401 from --show-error, but not GitHub's "message": "Bad credentials". Piping the body inline instead of to a file, or --output /dev/stderr, would keep it.

Pushed two separate commits addressing both points, good catches. Holler if I didn't do it quite how you might have thought tho!

thunderbiscuit (approved):
ACK 45f65149a5d37fd7cb4ba15ef143ad40b7678975.

PR #1069, opened 2026-08-10
(4.0) fix: return structured insufficient funds errors

reez:
### Description

We already expose a structured `CreateTxError::InsufficientFunds` containing `needed` and `available` amounts. 

Currently are converting the errors into the generic string based CoinSelection variant though.

So this PR just maps the upstream amounts to the existing `InsufficientFunds` and adds regression coverage. 

### Documentation

https://docs.rs/bdk_wallet/3.1.0/bdk_wallet/coin_selection/struct.InsufficientFunds.html

https://docs.rs/bdk_wallet/3.1.0/src/bdk_wallet/wallet/coin_selection.rs.html#123-132

https://docs.rs/bdk_wallet/3.1.0/bdk_wallet/error/enum.CreateTxError.html#variant.CoinSelection

- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/error.rs
+++ b/bdk-ffi/src/error.rs
@@ -1082,9 +1082,11 @@ impl From<BdkCreateTxError> for CreateTxError {
             BdkCreateTxError::OutputBelowDustLimit(index) => CreateTxError::OutputBelowDustLimit {
                 index: index as u64,
             },
-            BdkCreateTxError::CoinSelection(e) => CreateTxError::CoinSelection {
-                error_message: e.to_string(),
-            },
+            BdkCreateTxError::CoinSelection(e) => {
+                let needed = e.needed.to_sat();
+                let available = e.available.to_sat();
+                CreateTxError::InsufficientFunds { needed, available }
+            }
             BdkCreateTxError::NoRecipients => CreateTxError::NoRecipients,
             BdkCreateTxError::Psbt(e) => CreateTxError::Psbt {
                 error_message: e.to_string(),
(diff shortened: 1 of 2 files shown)


Johnosezele (approved):
ACK https://github.com/bitcoindevkit/bdk-ffi/pull/1069/commits/d80034a1a4ec0df7faa8662df49be4eb1872abf7

thunderbiscuit (commented):
I'm trying to dig into this one and I think we might have had an extra variant that cannot be hit?

If you look at the `CreateTxError` in Rust, you find a CreateTxError::CoinSelection variant: https://docs.rs/bdk_wallet/3.1.0/bdk_wallet/error/enum.CreateTxError.html#variant.CoinSelection. This variant contains the [`InsufficientFunds` struct](https://docs.rs/bdk_wallet/3.1.0/bdk_wallet/coin_selection/struct.InsufficientFunds.html).

Here instead we have InsufficientFunds as an error variant, and the CreateTx::CoinSelection variant can never really be returned anywhere as far as I see.

I think we can't return full structs inside our errors (wait can we with the new 0.31? need to check), and in my mind the `CreateTxError::InsufficientFunds` we are currently using is the most intuitive way to return this to the user (rather than putting the fields on a `CreateTxError::CoinSelection` variant), but just want to point it out, and suggest we remove the `CreateTxError::CoinSelection` variant as part of this PR.

thunderbiscuit:
Ok following up on this:

1. Yes we can return fields with full complex structs in our errors! Not needed here IMO but it opens up the space of what we can do with the errors, and adds to my neverending list of things to take a look at.
2. According to Claude this is unfortunately a breaking change, because although not a compilation breaking change, behaviouraly if you previously expected to get this "insufficient funds" data through the `CreateTxError::CoinSelection` and your were matching on it, and now that variant never returns anymore and instead this info comes in from the `CreateTxError::InsufficientFunds` variant, you wouldn't know about it and not get the info you previously were able to get.
3. When we do add it as a 4.0 breaking change, I would remove the `CreateTxError::CoinSelection` variant at the same time.

reez:
Agree. Adding to 4.0 tracking, and we can get this just how we want it for 4.0

PR #1070, opened 2026-08-10
fix: reject trailing bytes when decoding transactions

reez:
### Description

Previously we used Cursor + consensus_decode which decodes one transaction but could leave trailing bytes. 

We now use upstream deserialize which requires the entire array to be consumed (matching Transaction::new’s expectation that its input contains one complete transaction)

### Notes to the reviewers

One thing to note is that consensus_decode had a rust-bitcoin 4 MB read limit, while deserialize can process the full byte array passed to it. (Very large inputs could therefore take more work to process than before but this does not change Bitcoin consensus behavior or affect valid onchain transactions from what I'm thinking but let me know otherwise?!)

### Documentation

- [x] Other: 
https://docs.rs/bitcoin/0.32.8/bitcoin/consensus/fn.deserialize.html

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -14,10 +14,8 @@ use bdk_wallet::bitcoin::blockdata::block::Block as BdkBlock;
 use bdk_wallet::bitcoin::blockdata::block::Header as BdkHeader;
 use bdk_wallet::bitcoin::consensus::encode::deserialize;
 use bdk_wallet::bitcoin::consensus::encode::serialize;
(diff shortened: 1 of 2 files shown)


thunderbiscuit (approved):
ACK b1caf98bd2ad8cdb70f23f1bc34895fa2c87cefb.

thunderbiscuit:
This is a good one and I'd merge next after a rebase.

reez:
> This is a good one and I'd merge next after a rebase.

rebased. will let ci pass then I'll merge

Issue #1036, opened 2026-06-22
Breaking Changes For Next Major Version (4.0)


thunderbiscuit:
This is a meta issue to help us track and remember the breaking changes we'd like to add to the next major release. We sometimes find small improvements that could be added to the codebase but are breaking changes, and we don't have a good place to keep those in mind.

- [ ] #1035 
- [ ] #1039 
- [ ] Possible wallet constructor changes [#1044 review](https://github.com/bitcoindevkit/bdk-ffi/pull/1044#pullrequestreview-4820591473)
- [ ] #1072 
- [ ] #1069

thunderbiscuit:
#1035

thunderbiscuit:
#1039

reez:
Possible constructor changes https://github.com/bitcoindevkit/bdk-ffi/pull/1044#pullrequestreview-4820591473

j-kon:
Tracking the descriptor constructor breaking changes in #1072.

reez:
Tracking #1069

j-kon:
Tracking #1084 for the 4.0 `CbfNode::run()` breaking change.

PR #1072, opened 2026-08-11
fix: return descriptor template key type errors

### LINKED ISSUE #1071: BIP descriptor constructors panic for valid single descriptor keys
**Describe the bug**

`DescriptorSecretKey::from_string()` accepts a valid single private key encoded as WIF, but the BIP template constructors treat `BdkDescriptorSecretKey::Single(_)` as unreachable. Passing that accepted key to `Descriptor::new_bip44()`, `new_bip49()`, `new_bip84()`, or `new_bip86()` therefore panics inside Rust instead of returning a recoverable error to the UniFFI caller.

The corresponding public constructors have the same issue for `BdkDescriptorPublicKey::Single(_)`. Multipath variants are also handled with `unreachable!()`.

Current example: [`new_bip84()`](https://github.com/bitcoindevkit/bdk-ffi/blob/1fdfdf63f7dad278c1110dd5caadf5cf483684d2/bdk-ffi/src/descriptor.rs#L185-L210).

**To Reproduce**

```rust
let key = DescriptorSecretKey::from_string(
    "L2wTu6hQrnDMiFNWA5na6jB12ErGQqtXwqpSL7aWquJaZG8Ai3ch".to_string(),
)
.unwrap();

let _descriptor = Descriptor::new_bip84(
    &key,
    KeychainKind::External,
    NetworkKind::Main,
);
```

`from_string()` succeeds, then `new_bip84()` reaches `BdkDescriptorSecretKey::Single(_) => unreachable!()`. The same sequence is available from generated Kotlin and Swift bindings.

**Expected behavior**

Unsupported key variants should produce a typed error such as an invalid-key-type/descriptor error instead of panicking across the FFI boundary. The other BIP template constructors and their public-key variants should behave consistently.

**Build environment**

- BDK FFI commit: `1fdfdf63f7dad278c1110dd5caadf5cf483684d2` (`master`)
- OS/target: not platform-specific; reachable through UniFFI bindings

**Additional context**

This is related to, but not covered by, [PR #1027](https://github.com/bitcoindevkit/bdk-ffi/pull/1027), which fixed the analogous single-key panic only in `DescriptorSecretKey::secret_bytes()`. [PR #506](https://github.com/bitcoindevkit/bdk-ffi/pull/506) added `InvalidKeyType` handling to descriptor key derive/extend methods, but did not change the BIP descriptor constructors.

j-kon:
### Description

Closes #1071.

The BIP44, BIP49, BIP84, and BIP86 descriptor constructors accepted descriptor key objects whose underlying single or multipath variants were handled with `unreachable!()`. This allowed generated binding callers to trigger a Rust panic after successfully parsing a WIF or other valid descriptor key.

This change makes every BIP template constructor return `DescriptorError::InvalidKeyType` for unsupported variants. The secret-key constructors now return `Result`, and their template build errors are propagated instead of unwrapped.

### Breaking change

The four secret BIP template constructors are now throwable in generated bindings.

### Validation

- `cargo fmt --all -- --config format_code_in_doc_comments=true --check`
- `cargo clippy --all-targets --features "uniffi/bindgen-tests"`
- `CLASSPATH=./tests/jna/jna-5.14.0.jar cargo test --features uniffi/bindgen-tests`

### DIFF
--- a/bdk-ffi/src/descriptor.rs
+++ b/bdk-ffi/src/descriptor.rs
@@ -54,25 +54,22 @@ impl Descriptor {
         secret_key: &DescriptorSecretKey,
         keychain_kind: KeychainKind,
         network_kind: NetworkKind,
-    ) -> Self {
+    ) -> Result<Self, DescriptorError> {
         let derivable_key = &secret_key.0;
 
         match derivable_key {
-            BdkDescriptorSecretKey::Single(_) => {
-                unreachable!()
+            BdkDescriptorSecretKey::Single(_) | BdkDescriptorSecretKey::MultiXPrv(_) => {
+                Err(DescriptorError::InvalidKeyType)
             }
             BdkDescriptorSecretKey::XPrv(descriptor_x_key) => {
                 let derivable_key = descriptor_x_key.xkey;
                 let (extended_descriptor, key_map, _) = Bip44(derivable_key, keychain_kind)
                     .build(network_kind)
-                    .unwrap();
-                Self {
+                    .map_err(DescriptorError::from)?;
+                Ok(Self {
                     extended_descriptor,
                     key_map,
-                }
-            }
-            BdkDescriptorSecretKey::MultiXPrv(_) => {
-                unreachable!()
+                })
             }
         }
     }
@@ -93,8 +90,8 @@ impl Descriptor {
         let derivable_key = &public_key.0;
 
         match derivable_key {
-            BdkDescriptorPublicKey::Single(_) => {
-                unreachable!()
+            BdkDescriptorPublicKey::Single(_) | BdkDescriptorPublicKey::MultiXPub(_) => {
+                Err(DescriptorError::InvalidKeyType)
             }
             BdkDescriptorPublicKey::XPub(descriptor_x_key) => {
                 let derivable_key = descriptor_x_key.xkey;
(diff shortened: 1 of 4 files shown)


reez:
Looking good thanks for taking this on. Please add the `changelog: breaking` label. Since the four secret constructors now throw/return a result I dont think this can merge into a near term non-breaking release, so could we also track or link it from the existing 4.0 changes issue https://github.com/bitcoindevkit/bdk-ffi/issues/1036 so it’s queued up for the next breaking release? Otherwise the implementation and tests looking good to me.

j-kon:
Thanks for the review. I linked #1072 from the 4.0 tracking issue #1036. I also attempted to add `changelog: breaking`, but GitHub rejected it because I do not have label permissions on this repository, so a maintainer will need to apply that label.

Johnosezele (changes requested):
Looking at the scope of things, wouldn't it be beneficial if we extract the shared `require_xprv`, `require_xpub` and `from_template` helpers so the validation lives in one place, I think there's already an existing debt that would keep compounding if  someone later decides `MultiXPrv` should map to a different error than `Single`, or adds a fourth variant, they'd need to update eight sites... mssing one is how `unreachable!()` survived in the first place.

j-kon:
@Johnosezele Addressed in b47a869: key-type validation now lives in shared `require_xprv` and `require_xpub` helpers, and template construction/error mapping lives in generic `from_template`. The eight exported constructors remain explicit for their UniFFI APIs. I also rebased onto current `master` and reran format, bindgen-feature clippy, and the full test suite.

PR #1073, opened 2026-08-12
Fix recommended gpg command docs

thunderbiscuit:
The command we recommend in SECURITY.md doesn't work as intended. See bitcoindevkit/bdk#2257.


reez (approved):
ACK 73383cdda29506ae8714e496eb4c5de23177df6f

PR #1076, opened 2026-08-12
fix: prevent maximum fee rate overflow

reez:
### Description

Fixes overflow when converting and displaying extremely large FeeRate values (including PSBT fee overflow errors), and adds regression coverage for both paths.

### Documentation

- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/error.rs
+++ b/bdk-ffi/src/error.rs
@@ -1332,6 +1332,7 @@ impl From<BdkExtractTxError> for ExtractTxError {
     fn from(error: BdkExtractTxError) -> Self {
         match error {
             BdkExtractTxError::AbsurdFeeRate { fee_rate, .. } => {
+                let fee_rate = crate::bitcoin::FeeRate::from(fee_rate);
                 let sat_per_vbyte = fee_rate.to_sat_per_vb_ceil();
                 ExtractTxError::AbsurdFeeRate {
                     fee_rate: sat_per_vbyte,
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -163,7 +163,7 @@ impl FeeRate {
 
     /// Converts to sat/vB rounding up.
     pub fn to_sat_per_vb_ceil(&self) -> u64 {
-        self.0.to_sat_per_vb_ceil()
+        self.0.to_sat_per_kwu().div_ceil(250)
     }
 
     /// Converts to sat/vB rounding down.
@@ -203,7 +203,7 @@ impl FeeRate {
 
 impl Display for FeeRate {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        write!(f, "{:#}", self.0)
+        write!(f, "{}.00 sat/vbyte", self.to_sat_per_vb_ceil())
     }
 }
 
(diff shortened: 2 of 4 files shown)


ItoroD (approved):
ACK 00a1163a41209c649007342bb3121aa6f2445c0a

bdk-ffi/src/error.rs
```diff
@@ -1332,7 +1332,7 @@ impl From<BdkExtractTxError> for ExtractTxError {
     fn from(error: BdkExtractTxError) -> Self {
         match error {
             BdkExtractTxError::AbsurdFeeRate { fee_rate, .. } => {
-                let sat_per_vbyte = fee_rate.to_sat_per_vb_ceil();
+                let sat_per_vbyte = fee_rate.to_sat_per_kwu().div_ceil(250);
                 ExtractTxError::AbsurdFeeRate {
                     fee_rate: sat_per_vbyte,
                 }
```
Johnosezele, line 1339: small nit here: I think this conversion could be routed through the existing `FeeRate` wrapper

```suggestion
BdkExtractTxError::AbsurdFeeRate { fee_rate, .. } => ExtractTxError::AbsurdFeeRate {
                fee_rate: crate::bitcoin::FeeRate::from(fee_rate).to_sat_per_vb_ceil(),
            },
```

Johnosezele (changes requested):
cACK, one small nit on my end

bdk-ffi/src/error.rs
reez: Agreed, updated. I tweaked the suggestion slightly. Thanks!

Johnosezele (approved):
ACK https://github.com/bitcoindevkit/bdk-ffi/pull/1076/commits/b1200609a72da28e132eb70295a8583ad54f08db

thunderbiscuit (approved):
ACK b1200609a72da28e132eb70295a8583ad54f08db. This is a good fix to a bug. Ready to merge.

## More thoughts

The issue has been flagged in rust-bitcoin however, and the fix has been merged there, so I opened #1122 to remind us to remove this custom logic here when we bump to the 0.33.0 version of rust-bitcoin.

The change on the `Display` trait reminded me of something I had brushed against in #859. We display the fee rate with 2 decimal points, but they can never be anything else than 0s! This is done I think to mirror what we're used to seeing (fee rates are often expresses as floating numnber, i.e. `2.96`, but in our case it's kind of implying a lot more precision than our type can give. Check my table below for examples. Anyway it's not lying because the method name implies this is a ceiling, and we also offer the floor version, but it's odd!

Note that this has been discussed in rust-bitcoin, and their decision for the latest Feerate in 0.33 was actually to not have any Display implemented on the type, and recommend applications write their own.

FeeRate doc comment:
> FeeRate explicitly does not have any format/display trait implementations, as it doesn't have a standard unit for measure. Users are expected to format it on their own by extracting values in desired units with from_sat_per* functions.

All of this to say that when we do upgrade, we'll have to write our own implemetation of the trait, or leave it to users to do it for themselves.

```md
┌─────────┬─────────────┬────────────────┐
│ sat/kwu │ Real sat/vB │   Displayed    │
├─────────┼─────────────┼────────────────┤
│ 250     │ 1.0         │ 1.00 sat/vbyte │
├─────────┼─────────────┼────────────────┤
│ 253     │ 1.012       │ 2.00 sat/vbyte │
├─────────┼─────────────┼────────────────┤
│ 375     │ 1.5         │ 2.00 sat/vbyte │
├─────────┼─────────────┼────────────────┤
│ 25      │ 0.1         │ 1.00 sat/vbyte │
└─────────┴─────────────┴────────────────┘
```

PR #1077, opened 2026-08-12
fix(kyoto): honor recovery birthdays on non mainnet networks

reez:
### Description

Honors explicit Kyoto recovery birthdays on every network (previously we had non mainnet birthdays fall back to genesis)

We had a non mainnet guard that predated `RecoveryPoint::Other` so I thought this might just be an oversight to fix now, but let me know if I'm wrong!

### Documentation

- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/kyoto.rs
+++ b/bdk-ffi/src/kyoto.rs
@@ -207,31 +207,33 @@ impl CbfBuilder {
                 checkpoint,
             } => {
                 let network = wallet.network();
-                // Any other network has taproot and segwit baked in since the genesis block.
-                if !matches!(network, Network::Bitcoin) {
-                    bdk_kyoto::ScanType::Recovery {
+                match checkpoint {
+                    RecoveryPoint::GenesisBlock => bdk_kyoto::ScanType::Recovery {
                         used_script_index,
                         checkpoint: HashCheckpoint::from_genesis(network),
-                    }
-                } else {
-                    match checkpoint {
-                        RecoveryPoint::GenesisBlock => bdk_kyoto::ScanType::Recovery {
-                            used_script_index,
-                            checkpoint: HashCheckpoint::from_genesis(wallet.network()),
-                        },
-                        RecoveryPoint::SegwitActivation => bdk_kyoto::ScanType::Recovery {
+                    },
+                    RecoveryPoint::SegwitActivation if matches!(network, Network::Bitcoin) => {
+                        bdk_kyoto::ScanType::Recovery {
                             used_script_index,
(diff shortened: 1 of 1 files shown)


bdk-ffi/src/kyoto.rs
```diff
@@ -207,31 +207,33 @@ impl CbfBuilder {
...
+                    }
+                    RecoveryPoint::SegwitActivation | RecoveryPoint::TaprootActivation => {
+                        bdk_kyoto::ScanType::Recovery {
                             used_script_index,
-                            checkpoint: HashCheckpoint::new(birthday.height, birthday.hash.0),
-                        },
+                            checkpoint: HashCheckpoint::from_genesis(network),
+                        }
                     }
+                    RecoveryPoint::Other { birthday } => bdk_kyoto::ScanType::Recovery {
```
j-kon, line 233: Could we add a regression test using `RecoveryPoint::Other` on a non-mainnet wallet and assert that the resulting checkpoint preserves the supplied birthday? This exact branch was previously bypassed for every non-mainnet network, and the current test suite would not catch that behavior returning.

reez: Yeah I considered this when I was writing the code because I agree it'd be strong regression coverage, and actually did it at first before I changed to whats currently in the PR. I opted to change it to what I have now because previously the cleanest unit test approach required extracting the mapping into a private helper solely to create a test seam, so I wasn’t loving that because I couldn’t find any real similar patterns of that here since conversions are generally handled through From/TryFrom. And then testing `CbfBuilder::build` directly would instead require running a Kyoto node with a loopback peer because the configured checkpoint is private upstream, etc. So what I preferred was what I ended up with but happy to see how you feel still and what others think, totally open to it if we feel like the tradeoff makes more sense.

rustaceanrob:
Looks good to me

bdk-ffi/src/kyoto.rs
j-kon: Thanks, that makes sense given `CbfBuilder::build` does not expose the configured checkpoint. I still think a small pure mapping helper would be justified here because it isolates this network-sensitive conversion and would let us regression-test the exact branch, but I don’t consider that blocking for this focused fix.

j-kon (approved):

rustaceanrob (approved):
Seems fine to go in without a dedicated test. The match is straightforward and this only effects test networks.

thunderbiscuit (approved):
ACK 9155ecdc6cf05efa23124af2068fa213e402907c. Please rebase and merge.

PR #1080, opened 2026-08-12
Fix persistence error propagation and unexpected callback crashes

reez:
### Description

Fix persistence error propagation across the Rust and foreign-binding boundaries.

Previously:

- `Wallet::persist` wrapped an existing `PersistenceError` producing messages such as `persistence error: persistence error: ...`.
- Unexpected errors thrown by custom Swift/Kotlin(?) persisters could panic the Rust callback bridge and crash the host application instead of returning a catchable error
- Create/load errors used the backend-specific prefix `sqlite persistence error:` even for custom persisters while their associated payloads contained another formatted prefix.

This PR keeps raw error causes in associated payloads and applies one backend-neutral `persistence error:` prefix when displaying them. No API types or signatures change but bindings messages + associated error payloads are corrected.

The third commit could stand as a separate PR because it adjusts create/load message semantics, but I included it here because it completes the same persistence error stuff.

### Documentation

- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/error.rs
+++ b/bdk-ffi/src/error.rs
@@ -239,7 +239,7 @@ pub enum CreateTxError {
 #[derive(Debug, thiserror::Error, uniffi::Error)]
 #[uniffi::export(Debug, Display)]
 pub enum CreateWithPersistError {
-    #[error("sqlite persistence error: {error_message}")]
+    #[error("persistence error: {error_message}")]
     Persist { error_message: String },
 
     #[error("the wallet has already been created")]
@@ -464,7 +464,7 @@ pub enum HashParseError {
 #[derive(Debug, thiserror::Error, uniffi::Error)]
 #[uniffi::export(Debug, Display)]
 pub enum LoadWithPersistError {
-    #[error("sqlite persistence error: {error_message}")]
+    #[error("persistence error: {error_message}")]
     Persist { error_message: String },
 
     #[error("the loaded changeset cannot construct wallet: {error_message}")]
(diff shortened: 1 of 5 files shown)


j-kon (approved):

Johnosezele (approved):

bdk-ffi/src/error.rs
thunderbiscuit: This is now unused, since all persistence errors now go through `PersistenceError`, since PersistenceType::initialize/persist convert rusqlite::Error internally via From<BdkSqliteError> for PersistenceError.

I think we should clean it up in a separate commit but as part of this PR as well.

thunderbiscuit: Same as above. This is now unused, since all persistence errors now go through `PersistenceError`.

I think we should clean it up in a separate commit but as part of this PR as well.

thunderbiscuit: And removing this one, when I compile locally, gives you a warning for the unsued `use bdk_wallet::chain` import.

bdk-ffi/src/error.rs
```diff
@@ -1397,6 +1397,14 @@ impl From<BdkSqliteError> for PersistenceError {
     }
 }
 
+impl From<uniffi::UnexpectedUniFFICallbackError> for PersistenceError {
```
thunderbiscuit, line 1368: I now understand that implementing this is useful for all traits implemented in the foreign languages. This library has 3:

- `pub trait Persistence`
- `pub trait FullScanScriptInspector`
- `pub trait SyncScriptInspector`

And this PR takes care of the first one. Thanks! Opening a new issue for this here: #1087.

thunderbiscuit (changes requested):
Good catch. This one sent me down paths I had not gone before, particularly with respect to the `uniffi:UnexpectedUniFFICallbackError` API. Anyway I have a few things I'd propose to add to this PR while we're here and dealing with those persistence errors.

thunderbiscuit:
To be fair I'm asking for changes in the review above but this could be merged as is too. Removing the dead errors could be just a follow-up. I just don't want to forget about it, and figured it's just as easy to add it here, but feel free to merge if you prefer.

reez:
> To be fair I'm asking for changes in the review above but this could be merged as is too. Removing the dead errors could be just a follow-up. I just don't want to forget about it, and figured it's just as easy to add it here, but feel free to merge if you prefer.

Happy to put em in here, added in 90b13bd7a8b112b8ee2a463809b54e27a9b1bcab if that looks like what you're thinking.

thunderbiscuit (approved):
ACK 90b13bd7a8b112b8ee2a463809b54e27a9b1bcab. This one needs a rebase and is good to go.

PR #1083, opened 2026-08-12
fix: validate taproot control block conversion

reez:
### Description

Reject invalid Taproot output key parity and mismatched leaf versions during PSBT input conversion (which prevents silent control block corruption).

### Documentation

- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/bitcoin.rs
+++ b/bdk-ffi/src/bitcoin.rs
@@ -1029,7 +1029,30 @@ impl TryFrom<Input> for BdkInput {
                 })?;
 
                 let output_key_parity = k.output_key_parity;
+                if !matches!(output_key_parity, 0 | 1) {
+                    return Err(AddForeignUtxoError::InputConversionError {
+                        error_message: format!(
+                            "invalid output key parity: {}",
+                            output_key_parity
+                        ),
+                    });
+                }
+
                 let leaf_version_u8 = k.leaf_version;
+                if v.leaf_version != leaf_version_u8 {
+                    return Err(AddForeignUtxoError::InputConversionError {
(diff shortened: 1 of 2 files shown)


thunderbiscuit (commented):
This looks like it solves the problem. Trying to review it alongside with Claude I have 1 piece of feedback I would keep, and it's the fact that the leaf-version equality check is reject PSBTs that rust-bitcoin itself accepts?

rust-bitcoin apparently doesn't check that the two leaf versions match. This allows creating a PSBT that cannot be finalized because it would create an invalid transaction. The check/guard here doesn't really hurt, but also is adding a layer of check not done elsewhere.

The other 2 parts of the PR are good and worth adding for sure.

reez:
> This looks like it solves the problem. Trying to review it alongside with Claude I have 1 piece of feedback I would keep, and it's the fact that the leaf-version equality check is reject PSBTs that rust-bitcoin itself accepts?
> 
> rust-bitcoin apparently doesn't check that the two leaf versions match. This allows creating a PSBT that cannot be finalized because it would create an invalid transaction. The check/guard here doesn't really hurt, but also is adding a layer of check not done elsewhere.
> 
> The other 2 parts of the PR are good and worth adding for sure.

Yeah, I think that rust-bitcoin accepts PSBTs where the control block and script value have different leaf versions.

I added the check because our current conversion otherwise uses the control block’s version for both fields. Removing only the check would silently change the script value’s version.

We could keep the check and reject mismatches, or convert each version separately and preserve both values (matching rust-bitcoin’s behavior?)

Any preference?

PR #1085, opened 2026-08-13
fix: enforce last call wins fee policy

reez:
### Description

Clear the competing fee setting whenever fee_rate or fee_absolute is called, enforcing documented last-call-wins behavior.

### Documentation

https://docs.rs/bdk_wallet/3.1.0/bdk_wallet/struct.TxBuilder.html#method.fee_absolute

- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/tx_builder.rs
+++ b/bdk-ffi/src/tx_builder.rs
@@ -275,6 +275,7 @@ impl TxBuilder {
     pub fn fee_rate(&self, fee_rate: &FeeRate) -> Arc<Self> {
         Arc::new(TxBuilder {
             fee_rate: Some(fee_rate.clone()),
+            fee_absolute: None,
             ..self.clone()
         })
     }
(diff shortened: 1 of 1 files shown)


ItoroD (commented):
Nice fix! Makes me wonder what was happening with when they were both set for TxBuilder since it was not being enforced before. 

Why do we have some test in the main txbuilder.rs file and in its test file. (This my be out of scope for this PR. Just wondering)

reez:
> Nice fix! Makes me wonder what was happening with when they were both set for TxBuilder since it was not being enforced before.
> 
> Why do we have some test in the main txbuilder.rs file and in its test file. (This my be out of scope for this PR. Just wondering)

Finish always applied fee rate before fee absolute so absolute always won

Yeah i think we could look at the tests thing in a separate pr but inline because they assert private fields at the moment

ItoroD (approved):
ACK b9a00111c855cd2646d48a781b28d5c076f19072

thunderbiscuit (approved):
ACK b9a00111c855cd2646d48a781b28d5c076f19072.

PR #1086, opened 2026-08-13
fix: reject invalid transaction builder current heights

reez:
### Description

Validates the current height values in both transaction builders, invalid heights now return `CreateTxError::LockTimeConversionError` from `finish()` instead of panicking

### Notes to the reviewers

Added a 2nd commit just trying to match upstream docs:

https://github.com/bitcoindevkit/bdk_wallet/blob/v3.1.0/src/wallet/tx_builder.rs#L642-L643

... doesnt have to be in this PR, but thought it made sense.

### Documentation

- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/tests/wallet.rs
+++ b/bdk-ffi/src/tests/wallet.rs
@@ -1,10 +1,10 @@
-use crate::bitcoin::{Amount, BlockHash, Network, NetworkKind};
+use crate::bitcoin::{Amount, BlockHash, FeeRate, Network, NetworkKind};
 use crate::descriptor::Descriptor;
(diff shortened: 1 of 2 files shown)


thunderbiscuit (approved):
ACK 21f61235d2be4438fb246c5baa9aed3fb179c111.

The fact that the upstream code panics there if you provide a wrong value is odd. We could open an issue but honestly with the TxBuilder not really taking those little polish PRs anymore I suspect it's not worth it.

Again this one sent me down the rabbit hole a bit. Thanks for all of these!

Issue #1089, opened 2026-08-13
Handle Kyoto node thread startup and exit failures


reez:
`CbfNode::run` currently builds a Tokio runtime with `unwrap()` and discards the result of `node.run().await`.

This can abort the process when runtime setup fails and reduces terminal node errors like `NoReachablePeers` to a generic `NodeStopped`.

j-kon:
I propose building the Tokio runtime before detaching the thread and using `std::thread::Builder::spawn`, allowing `CbfNode::run()` to return typed runtime/thread startup errors alongside #1092’s `NodeAlreadyStarted` error. Since `node.run().await` can fail later, should we expose that through an async `CbfNode::wait_for_exit()` method or a terminal-error receiver on `CbfClient`? I prefer `wait_for_exit()` because it keeps node lifecycle errors on `CbfNode` without changing the detached behavior.

PR #1092, opened 2026-08-13
Return an error when CbfNode is started twice

### LINKED ISSUE #1084: Make CbfNode::run return an error when called more than once in 4.0
### Context

Follow-up to #1062 and tracked for the next breaking release in #1036.

Commit `f5c5ebc` fixed the binding-facing panic from calling `CbfNode::run()` more than once by making subsequent calls no-ops. That was appropriate for the current non-breaking release, but it silently hides an invalid lifecycle call.

As discussed in #1062, the 4.0 API can make this state explicit and align `run()` with `CbfClient::shutdown()`, which already returns a `CbfError` when the node is no longer running.

### Proposed change

Change the exported method from:

```rust
pub fn run(self: Arc<Self>)
```

to:

```rust
pub fn run(self: Arc<Self>) -> Result<(), CbfError>
```

The first call should start the node and return `Ok(())`. A later or concurrent call after the inner node has been consumed should return a binding-visible `CbfError` rather than panic or silently succeed.

This does not make `CbfNode` restartable; `run()` remains one-shot.

The implementation could either reuse `CbfError::NodeStopped` for consistency with `shutdown()` or introduce a more specific variant such as `NodeAlreadyStarted`. Maintainer preference should determine the exact error semantics.

### Acceptance criteria

- `CbfNode::run()` returns `Result<(), CbfError>`.
- The first call returns `Ok(())` and starts the node.
- A second or racing call returns a typed error without panicking.
- Generated binding call sites and tests are updated for the throwing API.
- Regression coverage verifies the consumed-node path.
- The PR is marked as an API-breaking 4.0 change.

### References

- Original panic report and non-breaking fix: #1062
- 4.0 breaking-change tracker: #1036
- Fix commit: `f5c5ebc8299f4f6cd135d23b3ba75cc24ab12751`

j-kon:
### Description

Closes #1084.

Changes the exported `CbfNode::run()` method to return `Result<(), CbfError>` and adds the binding-visible `NodeAlreadyStarted` error. The first caller starts the node; later or racing callers receive the typed error instead of silently succeeding.

Regression coverage exercises consumed, sequential, and concurrent calls in Rust. The Android binding test also verifies the generated `CbfException.NodeAlreadyStarted` exception.

### Notes to the reviewers

This intentionally covers only duplicate lifecycle calls. Reporting Tokio runtime construction or Kyoto node-exit failures remains tracked separately in #1089.

Binding verification completed with Kotlin generation plus `:lib:compileDebugAndroidTestKotlin`, and Swift generation plus `swift test`.

### Documentation

- [x] Other: lifecycle semantics and acceptance criteria are defined in #1084

### Changelog

Breaking API change: `CbfNode.run()` now throws/returns an error. Please apply `changelog: breaking`.

#### All Submissions:

* [x] I have signed all my commits
* [x] I followed the contribution guidelines
* [ ] I have added exactly one `changelog:*` label
* [x] I have linked the relevant upstream docs or specs above

#### Bugfixes:

* [x] I have added tests to reproduce the issue which are now passing
* [x] I am linking the issue being fixed by this PR

### DIFF
--- a/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/CbfSyncTest.kt
+++ b/bdk-android/lib/src/androidTest/kotlin/org/bitcoindevkit/CbfSyncTest.kt
@@ -1,6 +1,7 @@
 package org.bitcoindevkit
 
 import kotlin.test.Test
+import kotlin.test.assertFailsWith
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.platform.app.InstrumentationRegistry
 import kotlinx.coroutines.launch
@@ -53,6 +54,9 @@ class CbfSyncTest {
                 .build(wallet)
 
             node.run()
+            assertFailsWith<CbfException.NodeAlreadyStarted> {
+                node.run()
+            }
 
             val warningJob = launch {
                 try {
--- a/bdk-ffi/src/error.rs
+++ b/bdk-ffi/src/error.rs
@@ -164,6 +164,9 @@ pub enum CannotConnectError {
 pub enum CbfError {
     #[error("the node is no longer running")]
     NodeStopped,
+
+    #[error("the node has already been started")]
+    NodeAlreadyStarted,
 }
 
 #[derive(Debug, thiserror::Error, uniffi::Error)]
--- a/bdk-ffi/src/kyoto.rs
+++ b/bdk-ffi/src/kyoto.rs
@@ -67,12 +67,10 @@ pub struct CbfNode {
 #[uniffi::export]
 impl CbfNode {
     /// Start the node on a detached OS thread and immediately return.
-    /// Subsequent calls have no effect.
-    pub fn run(self: Arc<Self>) {
+    /// Returns an error if the node has already been started.
+    pub fn run(self: Arc<Self>) -> Result<(), CbfError> {
         let mut lock = self.node.lock().unwrap();
-        let Some(node) = lock.take() else {
-            return;
-        };
+        let node = lock.take().ok_or(CbfError::NodeAlreadyStarted)?;
         std::thread::spawn(|| {
             tokio::runtime::Builder::new_multi_thread()
                 .enable_all()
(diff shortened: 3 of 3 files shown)


j-kon:
@reez, could you please review this API-breaking fix for #1084 when you have a chance?

henrypeters (changes requested):
## Review of PR #1092: Return an error when CbfNode is started twice

###  **Suggestions for Improvement**

#### - **Add Swift Binding Test**
Currently only Android/Kotlin tests the generated exception. Consider adding equivalent Swift test coverage to ensure consistency across all bindings.

#### - **Documentation: Recovery Strategies**
Add guidance in docstring about error handling patterns for users. Include examples of how to handle the error gracefully.

#### - **Consider Adding State Inspection Method**
To help consumers avoid race conditions, consider adding (as a follow-up):
```rust
pub fn is_running(&self) -> bool {
    self.node.lock().unwrap().is_none()
}
```

#### - **Changelog Label** 
The PR description mentions 'Please apply `changelog: breaking`' but notes the label wasn't added. **Ensure this is applied before merge** - this is required.

#### - **Comment in Concurrent Test**
The `concurrent_run_calls_allow_one_caller` test is excellent but could use inline comments explaining the barrier pattern for clarity.

j-kon:
@henrypeters Thanks for the review. I checked each suggestion against the current API:

- `CbfClient::is_running()` already reports the Kyoto requester runtime state. A `CbfNode::is_running()` implemented as `node.is_none()` would only report that the one-shot node was consumed; it would remain true after the detached node thread exits, so I have not added that misleading API.
- The generated Swift binding exposes `run() throws`, and the XCFramework build plus Swift test suite compile that API. The Android instrumentation test provides the binding-level runtime regression; this repository currently has no Swift CBF fixture, so adding one would expand the scope beyond this fix.
- The `run()` docs already state the one-shot lifecycle and typed `NodeAlreadyStarted` failure. The barrier test directly coordinates the two competing callers, so I kept it focused rather than adding comments that repeat the code.

The branch is now rebased onto current `master`. Rust format/clippy/tests, Android test compilation, and Swift XCFramework/tests all pass locally.

@thunderbiscuit, could you apply `changelog: breaking` and review when convenient?

Issue #1093, opened 2026-08-14
DescriptorSecretKey.as_public panics for multipath private keys


reez:
`DescriptorSecretKey::from_string` accepts multipath xprvs, but `as_public()` unwraps Miniscript’s conversion error

PR #1108, opened 2026-08-31
fix: Add DescriptorKeyParseError for as_public

ItoroD:
### Description

Addresses #1093 

### Notes to the reviewers

- Error is expected for multipath in [rust-miniscript](https://github.com/rust-bitcoin/rust-miniscript/blob/d2ec50453612c8888df0266f871c27011b1c8ecc/src/descriptor/key.rs#L629). Now I return error rather than panic.
- Also, it seems later version of mini-script now handles for multipath [here](https://github.com/rust-bitcoin/rust-miniscript/blob/d2ec50453612c8888df0266f871c27011b1c8ecc/src/descriptor/key.rs#L325) (master)

### Documentation

- [ ] Other: 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/keys.rs
+++ b/bdk-ffi/src/keys.rs
@@ -256,11 +256,32 @@ impl DescriptorSecretKey {
         }
     }
 
-    /// Return the descriptor public key corresponding to this secret.
-    pub fn as_public(&self) -> Arc<DescriptorPublicKey> {
+    /// Returns the public version of this key.
+    ///
+    /// If the key is an "XPrv", the hardened derivation steps will be applied
+    /// before converting it to a public key.
+    ///
+    /// It will return an error if the key is a "multi-xpriv", as we wouldn't
+    /// always be able to apply hardened derivation steps if there are multiple
+    /// paths.
+    pub fn as_public(&self) -> Result<Arc<DescriptorPublicKey>, DescriptorKeyError> {
         let secp = Secp256k1::new();
-        let descriptor_public_key = self.0.to_public(&secp).unwrap();
-        Arc::new(DescriptorPublicKey(descriptor_public_key))
+        let descriptor_public_key = self.0.to_public(&secp).map_err(DescriptorKeyError::from)?;
+        Ok(Arc::new(DescriptorPublicKey(descriptor_public_key)))
+    }
+
+    /// Get as many keys as derivation paths in this key.
+    ///
+    /// For raw keys and single-path extended keys it will return the key itself.
+    /// For multipath extended keys it will return a single-path extended key per derivation
+    /// path.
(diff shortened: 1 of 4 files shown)


bdk-ffi/src/keys.rs
```diff
@@ -256,11 +256,32 @@ impl DescriptorSecretKey {
...
+    /// before converting it to a public key.
+    ///
+    /// It will return an error if the key is a "multi-xpriv", as we wouldn't
+    /// always be able to apply hardened derivation steps if there are multiple
+    /// paths.
+    pub fn as_public(&self) -> Result<Arc<DescriptorPublicKey>, DescriptorKeyError> {
         let secp = Secp256k1::new();
-        let descriptor_public_key = self.0.to_public(&secp).unwrap();
-        Arc::new(DescriptorPublicKey(descriptor_public_key))
+        let descriptor_public_key = self.0.to_public(&secp).map_err(DescriptorKeyError::from)?;
```
j-kon, line 269: Could we map the `MultiXPrv` case to `InvalidKeyType`, or preferably a dedicated `MultiPath` variant? The key parsed successfully; it is the conversion that is unsupported. Returning `Parse` makes binding callers inspect the error message to distinguish this case.

ItoroD: The key parse error refers to what this function in particular is doing. So I won't say "the key parsed successfully" because it cannot parse the mutlixPriv to make it public. Hence, the parsing error. This is also consistent with rust-miniscript error as I shared in the PR description. So returning an invalidKeyType might be misleading.

I suspecting the parsing you are refering to is the one that happens before this function is used. i.e DescriptorSecretKey::from_string like in my test.

j-kon: Yeah, I see what you mean. I was referring to the earlier `DescriptorSecretKey::from_string()` parse, since at that point we already have a valid `MultiXPrv`.

But you're right that `to_public()` itself returns `DescriptorKeyParseError` upstream, so mapping that through the existing `DescriptorKeyError::Parse` conversion is consistent with the current FFI error mapping.

My concern was mainly about whether binding callers need a typed way to distinguish this particular conversion failure rather than inspecting the error message, not that the upstream error type itself was wrong.

I also noticed newer miniscript handles some `MultiXPrv -> MultiXPub` conversions directly and only errors for incompatible hardened derivation paths, so `InvalidKeyType` would probably be too broad here.

Given that, I'm fine with `Parse` for this PR if we're intentionally following the existing upstream/error mapping. Thanks for clarifying.

ItoroD: Alright, thanks. Maybe if we update mini-script in the future we can update as well.

PR #1114, opened 2026-09-06
fix!: return errors from wallet address methods

### LINKED ISSUE #1105: fix!: return errors from Wallet.peek_address
Change `Wallet.peek_address` to return errors for out of range derivation indexes and descriptors that cannot produce an address (instead of allowing upstream `expect()` calls to panic)

Arowolokehinde:
### Description

Fixes #1105

`Wallet::peek_address` reached two upstream `expect` calls in bdk_wallet that panic: derivation
indexes in the BIP32 hardened range, and descriptors with no address form, i.e.
bare descriptors such as `pk(...)` and top-level `multi(...)`.

Per the discussion on the issue, the same `must have address form` panic is
reachable from `reveal_next_address`, `next_unused_address`, `reveal_addresses_to`
and `list_unused_addresses`, so all five address methods now return
`Result<_, AddressError>`.

The guard runs before delegating to `bdk_wallet`, which panics rather than
returning an error, so there is nothing to `map_err`. It also means the three
mutating methods no longer stage a derivation index change before failing.

### Notes to the reviewers

`AddressError` is shared across all five, though only `peek_address` can return
`IndexOutOfBounds`. 

`list_unused_addresses` is a behaviour change rather than a panic fix: it
previously returned an empty list on a bare descriptor wallet, because nothing
can be revealed on one. It now returns an error, for consistency with the other
four.

Swift needs `try` at three call sites. Kotlin and Python are source-compatible.

### Changelog

`changelog: breaking`

### Documentation

- [x] [`bdk_wallet`](https://docs.rs/bdk_wallet/3.1.0/bdk_wallet/struct.Wallet.html#method.peek_address)

#### All Submissions:
* [x] I followed the contribution guidelines

#### Bugfixes:

### DIFF
--- a/bdk-ffi/src/error.rs
+++ b/bdk-ffi/src/error.rs
@@ -53,6 +53,16 @@ pub enum AddForeignUtxoError {
 
 #[derive(Debug, thiserror::Error, uniffi::Error)]
 #[uniffi::export(Debug, Display)]
+pub enum AddressError {
+    #[error("derivation index {index} is greater than the bip32 maximum index of 2147483647")]
+    IndexOutOfBounds { index: u32 },
+
+    #[error("bare descriptor address")]
+    BareDescriptorAddr,
+}
+
+#[derive(Debug, thiserror::Error, uniffi::Error)]
+#[uniffi::export(Debug, Display)]
 pub enum AddressParseError {
     #[error("base58 address encoding error")]
     Base58,
--- a/bdk-ffi/src/tests/error.rs
+++ b/bdk-ffi/src/tests/error.rs
@@ -1,7 +1,7 @@
 use crate::error::{
-    Bip32Error, Bip39Error, CannotConnectError, DescriptorError, DescriptorKeyError, ElectrumError,
-    EsploraError, ExtractTxError, PsbtError, PsbtParseError, RequestBuilderError, SignerError,
-    TransactionError, TxidParseError,
+    AddressError, Bip32Error, Bip39Error, CannotConnectError, DescriptorError, DescriptorKeyError,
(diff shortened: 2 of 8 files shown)


Ugarba202 (approved):
This is an excellent catch and a very necessary fix! 💯 Mobile developers definitely prefer handling a thrown error over a complete app-crashing native panic.
I pulled the branch locally and ran through some checks:
* **Test Suite Verification:** Ran the standard `cargo test` suite, and all 77 unit tests passed. 
* **New Test Coverage:** The new tests you added (like `test_peek_address_bare_descriptor_returns_error` and `test_peek_address_above_max_index_returns_error`) perfectly cover the new guard logic.
* **API Break Acknowledgment:** Adding the `AddressError` return type to all five address methods is absolutely the right call to maintain consistency, even if it forces a `try` block in Swift. It's much better to have API-breaking safety than unpredictable panics.
The guards you placed before delegating to `bdk_wallet` are clean and effective. Thanks for making the mobile bindings more robust!

j-kon:
ACK c9aa633.

I checked the bounds handling, bare-descriptor guard, and the behavior of the other four address-returning methods against the upstream bdk_wallet implementation.

The wildcard check in peek_address looks correct, especially preserving the existing behavior where non-wildcard descriptors ignore the supplied index. Guarding the address form before calling the mutating upstream methods also avoids staging derivation changes before returning the error.

The Rust regression coverage looks good across the max index, out-of-range index, non-wildcard descriptors, and bare descriptors.

Non-blocking: since the main goal here is turning native panics into binding-visible errors, it would be nice to have one Kotlin or Swift regression test that actually catches the generated AddressError, for example verifying the out-of-range index field. The existing Rust coverage already proves the wrapper behavior, so I wouldn't block this PR on it.

Otherwise this looks good to me.

PR #1118, opened 2026-09-12
feat: build bdk-ffi for wasm32-unknown-unknown

bennyhodl:
### Description

Make the crate compile for `wasm32-unknown-unknown`, so that a UniFFI binding generator such as `uniffi-bindgen-react-native` (`wasm2` flavor) can produce browser and Node bindings from the same API the native bindings use. Native targets are unchanged.

How the wasm package gets built and published is still open. That discussion lives in this gist: https://gist.github.com/bennyhodl/e290b7de4ee8c746ee28f19695ab7bd2. Whatever is decided there, the `cfg` splits in this PR are required for any wasm build, because `rusqlite`, `minreq`, `bdk_electrum` and `bdk_kyoto` have no wasm32 support and cannot be replaced from a downstream package.

What changes:

- `Cargo.toml`: `rusqlite`, the blocking Esplora client, `bdk_electrum` and `bdk_kyoto` move to a `cfg(not(target_arch = "wasm32"))` table. A wasm32 table adds the async Esplora client, `getrandom/js`, UniFFI's `wasm-unstable-single-threaded` feature, `uniffi-runtime-wasm`, and `wasm-bindgen`/`js-sys`/`futures-channel` for the sleeper. `resolver = "2"` stops features from unifying across target tables.
- `esplora.rs`: on wasm32 `EsploraClient` wraps `AsyncClient`. Every method keeps its name, arguments and result but is `async`. The constructor returns a `Result` because building the reqwest client can fail. Retry backoff sleeps through the host's global `setTimeout`, so it works in browsers, workers and Node. The request/update helpers are shared by both implementations.
- `store.rs`: `Persister::custom` is available on every target. The SQLite constructors and the pre-v1 migration helper are native only.
- `error.rs`: `EsploraError` gains a `Reqwest` variant. It exists on every target so the generated API is identical, but native never produces it. Electrum, SQLite, migration and Kyoto conversions are native only.
- `lib.rs`: the `electrum` and `kyoto` modules are not compiled on wasm32. An `extern crate uniffi_runtime_wasm` line keeps the runtime exports from being dropped by the linker.
- CI: a new `wasm` job runs `cargo check --locked` and `cargo clippy --locked` for `wasm32-unknown-unknown`.

### Notes to the reviewers

- `wasm-unstable-single-threaded` is not a threading model choice. It only removes the `Send + Sync` bounds UniFFI requires on native, where foreign threads call in. `wasm32-unknown-unknown` has no threads. The `Persistence` trait keeps its `Send + Sync` bounds and compiles as is.
- `secp256k1-sys` needs a C compiler with a wasm32 backend. Ubuntu's clang has one, so CI sets `CC_wasm32_unknown_unknown=clang`. Apple's clang does not, so on macOS point that variable at an LLVM clang (for example Homebrew's `llvm`).
- Feature isolation was verified with `cargo tree`: native has no `reqwest` or `wasm-bindgen`, and wasm32 has no `rusqlite` or `minreq`.
- The `proxy` argument on the wasm32 `EsploraClient` constructor is kept for API parity and does nothing on that target.
- No `CHANGELOG.md` edit in this PR, following the release flow where entries come from the `changelog:` label.

### Documentation

- [x] Other:
  - esplora-client `AsyncClient`: https://docs.rs/esplora-client/0.12.3/esplora_client/async/struct.AsyncClient.html
  - esplora-client `Sleeper` trait: https://docs.rs/esplora-client/0.12.3/esplora_client/async/trait.Sleeper.html
  - bdk_esplora `EsploraAsyncExt`: https://docs.rs/bdk_esplora/0.22.2/bdk_esplora/trait.EsploraAsyncExt.html
  - `uniffi-runtime-wasm`: https://docs.rs/uniffi-runtime-wasm/0.31.0-5/uniffi_runtime_wasm/
  - uniffi-bindgen-react-native `wasm2` reference: https://github.com/jhugman/uniffi-bindgen-react-native/tree/main/docs/src/reference/wasm2
  - Build and publish proposal: https://gist.github.com/bennyhodl/e290b7de4ee8c746ee28f19695ab7bd2

### Changelog

`changelog: added`

#### All Submissions:

* [ ] I've added exactly one `changelog:*` label (needs a maintainer: `changelog: added`)

#### New Features:

* [ ] I've added tests for the new feature. The CI job checks and lints the wasm32 target. A runtime test needs a JavaScript host and lives with the binding generator, not in this crate. The native test suite is unchanged.

### DIFF
--- a/.github/workflows/rust-ci.yaml
+++ b/.github/workflows/rust-ci.yaml
@@ -47,6 +47,39 @@ jobs:
       - name: "Test"
         run: CLASSPATH=./tests/jna/jna-5.14.0.jar cargo test --features uniffi/bindgen-tests
 
+  wasm:
+    name: "Check wasm32"
+    runs-on: ubuntu-24.04
+    defaults:
+      run:
+        working-directory: bdk-ffi
+    env:
+      # secp256k1-sys needs a C compiler with a wasm32 backend
+      CC_wasm32_unknown_unknown: clang
+      AR_wasm32_unknown_unknown: llvm-ar
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v5
+        with:
+          persist-credentials: false
+
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          cache: false
+          components: clippy
+
+      - name: "Install LLVM"
+        run: sudo apt-get install -y llvm
+
+      - name: "Add wasm32 target"
+        run: rustup target add wasm32-unknown-unknown
+
+      - name: "Check wasm32-unknown-unknown"
+        run: cargo check --locked --target wasm32-unknown-unknown
+
+      - name: "Check clippy on wasm32"
+        run: cargo clippy --locked --target wasm32-unknown-unknown
+
   check:
     name: "Run clippy and fmt"
     needs: test
--- a/bdk-ffi/Cargo.lock
+++ b/bdk-ffi/Cargo.lock
@@ -81,6 +81,23 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
 
 [[package]]
+name = "async-trait"
+version = "0.1.92"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "82f6aeea286b8eb4dd3431a1be1b59d290ace00f5bfd8e2a159bc2a05e2c1667"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 3.0.4",
+]
+
+[[package]]
+name = "atomic-waker"
+version = "1.1.2"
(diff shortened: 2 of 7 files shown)


Ugarba202 (approved):
This is a massive step forward for the JS/web ecosystem! 
I pulled the branch locally and ran through some checks:
* **Dependency Isolation:** Verified the `cfg` feature gating in `Cargo.toml`. The `resolver = "2"` addition does a great job at isolating the features.
* **Wasm Compilation:** Ran `cargo clippy --target wasm32-unknown-unknown` and it compiled flawlessly without any feature leaks.
* **Native Integrity:** Ran the standard `cargo test` suite on the Native targets just to be extra safe, and everything passed perfectly. No regressions for existing mobile/native users.
The async translation for the Esplora client and the custom `setTimeout` sleeper wrapper for the retry backoffs look really solid. Exposing the `Reqwest` error variant globally to keep the UniFFI API shape identical across platforms is also a smart tradeoff.
Everything looks great on my end. Thanks for putting this together!

PR #1119, opened 2026-09-14
feat: add esplora client timeout option

### LINKED ISSUE #1109: Add a configurable request timeout to EsploraClient
EsploraClient.new does not expose upstream timeout configuration, so a stalled Esplora server can block requests without a library controlled deadline.

Arowolokehinde:
### Description

Closes #1109

`EsploraClient::new` had no way to set a request timeout: proxy was the
only setting it applied to `esplora_client::Builder`, leaving the
builder's timeout at `None`. A server that accepts a connection then stops
responding could block the calling thread indefinitely.

This adds an optional `timeout` (seconds), mirroring `ElectrumClient::new`. It
defaults to `None`, so callers opt in.

### Notes to the reviewers

- Defaults to `None` as suggested, so nothing changes for existing callers.
  Added a docstring note for the hang risk when it's unset.
- `Option<u8>` seconds matches `ElectrumClient`.
- Verified against a local stub server that accepts a connection and never
  replies. With `timeout` unset the call never returned; with `timeout = 3` it
  raised `Minreq("the timeout of the request was reached")` after 3.01s —
  catchable rather than a hang.
- No test added - this needs a server that accepts and then stalls.
- I don't have permissions to add labels - could someone apply
  `changelog: added`?

### Documentation

- [x] Other: [`esplora_client::Builder::timeout`](https://docs.rs/esplora-client/0.12.3/esplora_client/struct.Builder.html#method.timeout)

#### All Submissions:

#### New Features:


j-kon (approved):
ACK 853a5ba42179097ce8ffd5730692f469116c27af

PR #1123, opened 2026-09-18
Bump actions to their latest versions

thunderbiscuit:
This is a bump of some of our GitHub actions. I ran into problems with the `android-actions/setup-android@v3` on the Devkit Wallet and that prompted me to take a look at our actions here.

- Commit 1 is bumping actions that don't need anything new.
- Commit 2 bumps zizmor. We are currently using it the "manual" way, but there is also a GitHub action for it. I just didn't want to mess with it too much here. [See the action here](https://github.com/zizmorcore/zizmor-action)
- Commit 3 bumps the actions-rust-lang/setup-rust-toolchain action. [See the bdk_wallet PR for this](https://github.com/bitcoindevkit/bdk_wallet/pull/547). This one needed a small fix to keep our current behavior. The v2 version will not fail the runs on compiler warnings, which is what we currently do, unless we bump the compiler version to 1.97 or newer. I prefer to keep that bump to a separate PR just to make sure everything works well for our dependencies and whatnot, so I instead opted to turn on the flag manually. We can revert the `rustflags: "-D warnings"` lines once we bump the Rust compiler version.

Note that because these changes don't impact any of the Swift/Android/Rust files, the CI workflows don't run automatically. Here are 2 runs, showing green CI:

- [Swift](https://github.com/bitcoindevkit/bdk-ffi/actions/runs/35358939261)
- [Android](https://github.com/bitcoindevkit/bdk-ffi/actions/runs/35358706512)


ItoroD (approved):
LGTM

ACK c5c3ffb70c9f7a18e3e8ccf5ce8dc92df8605da9

thunderbiscuit:
Related issue is #1125

PR #1126, opened 2026-09-21
Drop support for 32-bit Android architectures

### LINKED ISSUE #1100: Prevent numeric input truncation on 32 bit Android
Several exported u64 inputs are converted to usize with `as`, silently truncating values above u32::MAX in packaged 32 bit Android artifacts.

(bdk-android documents 32 bit architectures as unsupported but still packages armeabi-v7a, so this could be p simple)

thunderbiscuit:
This officially drops support for 32-bit architectures on Android. 

We've talked to users quite a bit about this and there doesn't seem to be much users left on those very old devices. Our documentation already states that we don't support those; this is just making this a reality (oversight on our part).

Fixes #1100.

cc @JoelBryceAnderson, let me know if you'd rather we keep supporting it for a little longer! But after speaking with Renée last year it appeared that you barely had any downloads for it and were fine with us cutting it.

### Documentation

- [ ] Other: 

#### All Submissions:

#### Bugfixes:


JoelBryceAnderson (approved):
Approved from our end. We have pretty tiny audience of 32 bit (under 1%) so it shouldn't move the needle for us.

